Issue #5, opened 2019-08-28
Put the change of parameters inline


pthatcherg:
A reminder to:

Rather than call transform_stream.setParameters(...), attach new parameters to the stream that go into the WritableStream.

PR #14, opened 2019-09-04
Add detailed design discussion for codec configuration

steveanton:
Addresses #5 

PTAL @pthatcherg

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -293,6 +293,63 @@ input.readable.pipeInto(demuxer.writable);
 muxer.readable.pipeInto(output.writable);
 ```
 
+## Detailed design discussion
+
+### Codec configuration
+
+Many codecs and encoder/decoder implementations are highly configurable. WebCodecs intends to support most of the configuration options available in codecs today to efficiently allow for advanced use cases.
+
+Configuration options are classified into two types:
+- **Parameters** are metadata required to construct a compliant bitstream. These are required when constructing the encoder/decoder and cannot be changed. For example, the VP9 profile.
+- **Settings** are configuration options that influence the behavior of the encoder but do not change the type of bitstream produced. For example, target bitrate.
+
+Settings are further classified into two types:
+- **Static settings** must be specified when constructing the encoder and cannot be changed without reinitializing the encoder.
+- **Dynamic settings** apply to the lifetime of the encoder and can be changed at any point. Dynamic settings must be lightweight and not require reinitializing the encoder (concretely, does not require a new key frame to be produced).
+
+WebCodecs will maintain a standard definition of parameters for each supported codec. Additionally, the specification will establish common encoder settings that apply across codecs and implementation. However, we expect many settings will be implementation-specific. These will be available behind a feature detection and configuration API (TODO: sketch this).
+
+#### Configuration examples
+
+Both encoder and decoder constructors take in the codec name and required parameters. Encoders additionally take in a dictionary of codec settings.
+
+```javascript
+const encoder = new VideoEncoder({
+  codec: 'VP9',
+  profile: '1',
+  settings: {
+    targetBitRate: 80_000,
+  },
+});
+```
+
+Codec settings can be changed on-the-fly by bundling the changed settings with the next media chunk. The changed settings will be applied before encoding chunk and apply to subsequent chunks.
+
+```javascript
+const encoder = new VideoEncoder(...);
+const writer = encoder.writable.getWriter();
+writer.write({
+  imageData: ...,
+  timestamp: ...,
+  changeSettings: {
+    targetBitRate: 50_000,
+  },
+});
+```
+
+```javascript
+const encoder = new VideoEncoder(...);
+const writer = encoder.writable.getWriter();
+writer.write({
+  imageData: ...,
+  timestamp: ...,
+  changeSettings: {
+    // |imageData| or the next compatible image will be encoded as a key frame.
+    requestKeyFrame: true,
+  },
+});
+```
+
 ## Alternative designs considered
 
 Media Source Extensions (MSE) is already used widely for low-latency streaming.  However, there are some problems:
--- a/webidl.txt
+++ b/webidl.txt
@@ -3,29 +3,63 @@
 // TODO(when writing spec):
 // - Specify that encoding and decoding must happen off the main thread.
 
-[Constructor(MediaStreamTrack track)]
-interface AudioTrackReader {
-  readonly attribute ReadableStream readable; // of DecodedAudioPacket
+// Common definitions used for both audio and video.
+
+[Constructor(unsigned long long value, unsigned long long scale)]
+interface MediaTime {
+  readonly attribute unsigned long long value; 
+  readonly attribute unsigned long long scale;
 }
 
-interface DecodedAudioPacket {
-  readonly attribute MediaTime timestamp;
-  // Sample count == duration.value
-  // Sample rate == duration.scale
-  readonly attribute MediaTime duration;
-  readonly attribute unsigned long channelCount
+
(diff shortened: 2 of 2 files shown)


steveanton:
I took a stab at updating the webidl.txt. To be honest, I wonder if trying to map everything to webidl might be more confusing than it's worth at this point.

webidl.txt
```diff
@@ -72,50 +46,104 @@ dictionary AudioDecoderParams {
...
+  // TODO: decode stats.
+}
 
-[Constructor(MediaStreamTrack track)]
-interface VideoTrackReader {
-  readonly attribute ReadableStream readable; // of DecodedVideoFrame
+dictionary AudioEncoderStaticSettings {
+  // probably opus-specific
+  bool fec = false;  // enabled or not
+  bool dtx = false;  // enabled or not
```
pthatcherg: For network adaption, I think these need to be dynamic.  And maybe CBR as well.

webidl.txt
```diff
@@ -72,50 +46,104 @@ dictionary AudioDecoderParams {
...
+  unsigned long? bitsPerSecond;  
+
+  // codec-specific
+  // null/unset means use the codec default
+  unsigned long? complexity;
 }
 
-interface DecodedVideoFrame {
+dictionary AudioEncoderSettings : AudioEncoderStaticSettings, AudioEncoderDynamicSettings {
+}
```
pthatcherg: Would having the AudioEncoderSettings contain the other two rather than inherit from them be better?

webidl.txt
```diff
@@ -72,50 +46,104 @@ dictionary AudioDecoderParams {
...
+
+dictionary AudioEncoderInit : AudioCodecParameters {
+  AudioEncoderSettings? settings;
+}
+
+dictionary AudioEncoderFrameSettings {
```
pthatcherg: Since we went with "AudioPacket", I guess this should be "PacketSettings".

webidl.txt
```diff
@@ -148,51 +176,74 @@ dictionary VideoEncodeLayer {
...
-  // Info provided as a result from the encoder
-  // Not needed as input to a decoder
-  readonly attribute VideoEncodeResult? encoded;
+dictionary VideoEncoderStaticSettings {
+  // Can be used to initialize the encoder faster
+  // than waiting for the first frame
+  unsigned long? expectedWidth;
+  unsigned long? expectedHeight;  
+
+  VideoEncodeContentMode contentMode;
```
pthatcherg: Should this be static for switching between screencast and camera?

webidl.txt
```diff
@@ -148,51 +176,74 @@ dictionary VideoEncodeLayer {
...
+
+  sequence<VideoEncodeLayer> layers;
```
pthatcherg: Shouldn't this be dynamic to allow group call clients to change which layers they send with depending on the number of people in the room?

webidl.txt
```diff
@@ -148,51 +176,74 @@ dictionary VideoEncodeLayer {
...
+}
+
+dictionary VideoEncoderDynamicSettings {
+  // unset/null means the encoder will pick
+  // target will be exceeded for key frames
+  unsigned long long? targetBitRate;
+}
+
+dictionary VideoEncoderSettings : VideoEncoderStaticCodecSettings, VideoEncoderDynamicCodecSettings {
 }
```
pthatcherg: Same with inheritance here

webidl.txt
```diff
@@ -148,51 +176,74 @@ dictionary VideoEncodeLayer {
...
+  VideoEncoderSettings settings;
+}
+
+dictionary VideoEncoderFrameSettings {
+  boolean? forceKeyFrame;
+}
+
+dictionary VideoEncoderInput {
+  VideoFrame frame;
+  VideoEncoderFrameSettings settings;
```
pthatcherg: frameSettings?

webidl.txt
```diff
@@ -72,50 +46,104 @@ dictionary AudioDecoderParams {
...
+  AudioEncoderSettings? settings;
+}
+
+dictionary AudioEncoderFrameSettings {
+}
+
+dictionary AudioEncoderInput {
+  MediaTime timestamp;
+  ArrayBuffer buffer;
+  AudioEncoderFrameSettings settings;
```
pthatcherg: frameSettings?

webidl.txt
steveanton: I wanted to keep the JSON from becoming too hierarchical. I don't think it's too difficult to keep the names from clashing.

webidl.txt
steveanton: Actually removed this since I'm not sure it's useful to have frame-affiliated settings like it is for video.

webidl.txt
steveanton: Perhaps a distinction between static and dynamic parameters is the requirement to reinitialize the codec. Dynamic settings should be something that does not require reinitialization (concretely, does not require a new keyframe). In this case the right approach is to reinitialize the codec.

pthatcherg: Ah, that's a good idea.

pthatcherg: Can you put a comment to that effect somewhere in the doc?

steveanton: Yes, it was added to the explainer in the last diff.

webidl.txt
```diff
@@ -148,51 +173,74 @@ dictionary VideoEncodeLayer {
...
+}
+
+dictionary VideoEncoderInput {
+  VideoFrame frame;
+  VideoEncoderFrameSettings frameSettings;
+  VideoEncoderDynamicSettings changeCodecSettings;
+}
+
+dictionary VideoEncoderOutput {
+  Uint8Array data;
```
alvestrand: I'm not sure it's good to switch from interfaces to dictionaries with (passive) data at this point. If we want to make efficient transformations that only look at the metadata (such as a layer-discarding filter), it may be better to leave the data as an accessor so that the case of passing frames unmodified can be handled without any data copying.

explainer.md
```diff
@@ -293,6 +293,65 @@ input.readable.pipeInto(demuxer.writable);
...
+
+```javascript
+const encoder = new VideoEncoder(...);
+const writer = encoder.writable.getWriter();
+writer.write({
+  imageData: ...,
+  timestamp: ...,
+  changeCodecSettings: {
+    targetBitRate: 50_000,
+  },
```
pthatcherg: Idea expressed in conversation: allow changing of codec settings even without a frame (WritableStream of either settings or video frame)

explainer.md
```diff
@@ -293,6 +293,65 @@ input.readable.pipeInto(demuxer.writable);
...
+Frame settings are also bundled with the next input image. These settings do not persist beyond encoding for the image on which they appear.
+
+```javascript
+const encoder = new VideoEncoder(...);
+const writer = encoder.writable.getWriter();
+writer.write({
+  imageData: ...,
+  timestamp: ...,
+  frameSettings: {
+    forceKeyFrame: true,
```
pthatcherg: Not all frames can be key frames, so should this be a "make the next available frame a key frame" dynamic codec setting?  Or should we expect the user of the API to always be smart enough to time it right?

explainer.md
steveanton: https://github.com/WICG/web-codecs/issues/20

webidl.txt
steveanton: This is already the case for the heavyweight resources (e.g., raw video frames) but not for bitstream contents. I think that discussion is outside the scope of this PR but would be valuable.

explainer.md
steveanton: I was already having difficulty identifying frame settings, so I'm just going to remove the category. We can add it back later if it turns out to have use.

PR #17, opened 2019-09-06
Add detailed design discussion for execution environments

steveanton:
PTAL @pthatcherg @padenot

Should we allow instantiation inside service workers? I could see a use case for transcoding files as they download.

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -300,6 +300,18 @@ input.readable.pipeInto(demuxer.writable);
 muxer.readable.pipeInto(output.writable);
 ```
 
+## Detailed design discussion
+
+### Execution environment
+
+Encoder and decoder objects can be instantiated on the main thread and on dedicated workers. Additionally, the ReadableStream and WritableStream components of each object can be transferred to other contexts (utilizing the [transferable stream][transferable-streams] infrastructure).
+
+Encode and decode operations can be very computationally expensive. As such, user agents must perform the operations asynchronously with the JavaScript which initiates the operation. The execution environment of the codec implementation is defined by the user agent. Some possibilities include: sequentially or in parallel on an internal thread pool, or on a hardware acceleration unit (e.g., a GPU).
+
+By building on the Streams infrastructure, WebCodecs allows the user agent to optimize the transfer of chunks between JavaScript and internal execution environments. The user agent should take great care to efficiently handle expensive resources (e.g., video frame contents, GPU resource handles).
+
+[transferable-streams]: https://github.com/whatwg/streams/blob/master/transferable-streams-explainer.md
+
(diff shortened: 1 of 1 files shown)


explainer.md
```diff
@@ -300,6 +300,18 @@ input.readable.pipeInto(demuxer.writable);
 muxer.readable.pipeInto(output.writable);
 ```
 
+## Detailed design discussion
+
+### Execution environment
+
+Encoder and decoder objects can be instantiated on the main thread and on dedicated workers (TODO: service workers too?). Additionally, the ReadableStream and WritableStream components of each object can be transferred to other contexts (utilizing the [transferrable stream][transferable-streams] infrastructure).
+
+Encode and decode operations can be very computationally expensive. As such, user agents must perform the operations asynchronously with the JavaScript which initiates the operation. The execution environment of the codec implementation is defined by the user agent. Some possibilities include: sequentially or in parallel on an internal thread pool, or on a hardware acceleration unit (e.g., a GPU).
```
padenot: It's unclear if this is a necessary feature. I could certainly see a mode where this is done synchronously in a specific context. This way, you can have authors decide on the number of threads, parallel, sequential, etc.

explainer.md
```diff
@@ -300,6 +300,18 @@ input.readable.pipeInto(demuxer.writable);
...
+
+By building on the Streams infrastructure, WebCodecs allows the user agent to optimize the transfer of chunks between JavaScript and internal execution environments.
```
padenot: Yes, it's critical that this design rational is well understood. Not only for chunks, but only for frames, texture handles, etc. (depending on the platform). Or maybe that's what you meant already? The terminology is still a bit in flux :-).

explainer.md
```diff
@@ -300,6 +300,18 @@ input.readable.pipeInto(demuxer.writable);
 muxer.readable.pipeInto(output.writable);
 ```
 
+## Detailed design discussion
+
+### Execution environment
+
+Encoder and decoder objects can be instantiated on the main thread and on dedicated workers (TODO: service workers too?). Additionally, the ReadableStream and WritableStream components of each object can be transferred to other contexts (utilizing the [transferrable stream][transferable-streams] infrastructure).
```
padenot: I don't know enough about service workers to answer, sorry.

explainer.md
steveanton: I'll file a follow-up issue to discuss synchronous execution.

explainer.md
steveanton: "Chunks" is a term taken from the Streams infrastructure and refers to the data that flows in the stream. I added some clarification text about handling expensive resources.

Issue #35, opened 2019-12-18
What should the shape of the API be?


pthatcherg:
A number of considerations all combine in somewhat complex ways, such as (re)initialization, buffering, failure recovery, and flushing.  Obviously we'd like a good API for all of these, but there are tradeoffs between different options.  This issue is for tracking the discussion of what we want the API to look like.  

Here are some options.  Note that it may be possible to have different options for encode and decode.  For example, we could do a combination of B for encode and D for decode.

## Option A: new Encoder/Decoder for each change

Every time you want to change something that requires (re)initialization, such as changing the codec or resolution, create a new Encoder/Decoder.   Also reinit every time a flush is desired.  

Pros:
+ Simple API
+ It's clear when (re)init fails, and recovering is straightforward
+ If we have buffered frames, it's clear which initialization applies to which frames.
+ Flushing is just closing the .writable

Cons:
- Dealing with downstream changes to the pipeline may be difficult (because you have a new .readable that you need to pipe somewhere).
- Dealing with upstream changes to the pipeline may be difficult (because you have a new .writable that you need to pipe into).
- It may be much more efficient to have only one encoder/decoder around at a time which is difficult to manage when the JS is creating new ones for every resolution change and/or flush.

## Option B: An Initialize() method for each change

If a change requires a reinitialization, call Initialize(), as many times as you want.  The .writable and .readable are stable.

Pros:
- It's clear when (re)init fails, and recovering is straightforward
- .readable is the same across multiple initializations, which makes the downstream consumption easier (nothing to re-pipe).
- .writable is the same across multiple initializations, which makes the upstream production easier (nothing to re-pipe).
- Reinitialization can be efficient because the implementation can keep only one encoder/decoder around at a time.

Cons:
- Flushing mush be a separate method (since you can only call .close() on the .writable once)
- If buffering is used, which initialization applies to which frames becomes unclear.  If we don't buffer on the .writable, we can avoid this, but that means that we must require the JS to respect .ready and we must deal with what happens when it does not.

## Option C: An Initialize() method that produces a new WritableStream

If a change requires a reinitialization, call Initialize(), as many times as you want.  The .readable is stable, but not the .writable (if there is one).

Pros:
- It's clear when (re)init fails, and recovering is straightforward
- .readable is the same across multiple initializations, which makes the downstream consumption easier (nothing to re-pipe).
- Reinitialization can be efficient because the implementation can keep only one encoder/decoder around at a time.
+ Flushing is just closing the WritableStream
[...]

PR #48, opened 2020-04-10
Update explainer (decouple from streams)

chcunningham:
Major changes to API shape. See rationale in
"Alternative designs considered".

Removes streamy aspects from interface.
Updates all examples.
Updates discussion of worker transfer.
Updates discussion of codec configuration.
Adds discussion of streams to "Alternative designs considered"
Adds "Future considerations" with mix of old/new ideas.

Addresses parts of #35. Will send a second CL with full IDL to completely address that.

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -95,7 +95,7 @@ const canvasContext = canvasElement.getContext('bitmaprenderer');
 
 function paintFrameToCanvas(videoFrame) {
   // Paint every video frame ASAP for lowest latency.
-  canvasContext.transferFromImageBitmap(videoFrame.image);
+  canvasContext.transferFromImageBitmap(videoFrame.transferToImageBitmap());
 }
 
 const videoDecoder = new VideoDecoder({
--- a/explainer.md
+++ b/explainer.md
@@ -1,3 +1,4 @@
+
 # WebCodecs Explainer
 
 ## Problem and Motivation
@@ -39,7 +40,7 @@ Provide web apps with efficient access to built-in (software and hardware) media
 
 ## Key use-cases
 
-- Extremely low latency live streaming (<3s delay)
+- Extremely low latency live streaming (< 3s delay)
 - Cloud gaming
 - Live stream uploading
 - Non-realtime encoding/decoding/transcoding, such as for local file editing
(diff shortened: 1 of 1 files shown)


chcunningham:
@sandersdan

chcunningham:
@padenot

explainer.md
```diff
@@ -48,218 +49,298 @@ Provide web apps with efficient access to built-in (software and hardware) media
...
 
-// First, the tracks are converted to ReadableStreams of framed bitstream chunks.
+// On the receive side, encoded media is likely received
+// from an out-of-order p2p transport and then put into a buffer.
+// The output of that buffer is the source of encoded audio and video here.
+function streamEncodedChunks(decodeAudioCallback, decodeVideoCallback) { ... }
+
+// App rendering audio in AudioWorklet
+// TODO(chcunningham): source a demo.
+function renderInAudioWorklet(audioPacket) { ... }
```
chcunningham, line 277: @padenot roughly, I think this is something like the impl of [playSample() here](https://developer.mozilla.org/en-US/docs/Web/API/Web_Audio_API/Advanced_techniques), but connected to a worklet node for a worklet that just does bypass rendering. LMK if I'm on the right track.

explainer.md
```diff
@@ -48,218 +49,298 @@ Provide web apps with efficient access to built-in (software and hardware) media
...
   - spatial and temporal scalability
 - Decoded and encoding images
-- Reencoding multiple input media streams in order to merge many encoded media streams into one encoded media stream.
+- Re-encoding multiple input media streams in order to merge many encoded media streams into one encoded media stream.
 
 
 ## Proposed solutions
 
-We build on top of [WHATWG Streams](https://streams.spec.whatwg.org/) (in particular [TransformStreams](https://streams.spec.whatwg.org/#ts-class)) and integrate with [MediaStreamTracks](https://www.w3.org/TR/mediacapture-streams/#dom-mediastreamtrack).
+The WebCodecs interface is modeled on well known platform and software codec APIs. This affords the full power of the underlying platform codecs and is intuitive for authors with experience in developing media applications.
```
sandersdan, line 57: Nit: I don't think we plan to offer the 'full' power, as that would require exposing every platform detail.

explainer.md
```diff
@@ -48,218 +49,298 @@ Provide web apps with efficient access to built-in (software and hardware) media
...
+
+Core interfaces include
+
+-   **EncodedAudioChunks** and **EncodedVideoChunks** contain codec-specific encoded media bytes.
 
-**EncodedAudioChunk**s and **EncodedVideoChunk**s provide access to codec-specific encoded media bytes so they may be transported, saved, etc.
+-   **AudioPacket** contains decoded audio data. It will provide an [AudioBuffer](https://webaudio.github.io/web-audio-api/#audiobuffer) for rendering via [AudioWorklet](https://webaudio.github.io/web-audio-api/#audioworklet).
```
sandersdan, line 63: AudioWorklet doesn't seem to have any interface of its own. Is there a specific kind of processing node that we're trying to integrate with?

explainer.md
```diff
@@ -48,218 +49,298 @@ Provide web apps with efficient access to built-in (software and hardware) media
...
 
-**AudioPacket** may expose an [AudioBuffer](https://webaudio.github.io/web-audio-api/#audiobuffer) for rendering via [AudioWorklet](https://webaudio.github.io/web-audio-api/#audioworklet), while **VideoFrame** may expose an ImageBitmap for efficient rendering to canvas. Alternatively, AudioPackets and VideoFrames may be passed as opaque handles to other APIs (such as to/from MediaStreamTracks).
+-   **VideoFrame** contains decoded video data. It will provide an [ImageBitmap](https://html.spec.whatwg.org/multipage/imagebitmap-and-animations.html#imagebitmap) for manipulating in WebGL, including rendering to Canvas.
```
sandersdan, line 65: I think this is accurate for the MVP, but we get enough questions that it's worth adding that we hope to support YUV ArrayBuffers eventually.

explainer.md
```diff
@@ -48,218 +49,298 @@ Provide web apps with efficient access to built-in (software and hardware) media
...
-const canvas = document.getElementById("canvas");
+// The document contains a canvas for displaying VideoFrames.
+const canvasElement = document.getElementById("canvas");
+const canvasContext = canvasElement.getContext('bitmaprenderer');
 
-// Wrap the rendering sink in a writable stream so it can be connected to the graph.
-const renderingStream = new WritableStream(new CanvasRendererSink(canvas));
+function paintFrameToCanvas(videoFrame) {
+  // Paint every video frame ASAP for lowest latency.
+  canvasContext.transferFromImageBitmap(videoFrame.image);
```
sandersdan, line 98: Nit: this will probably be takeAsImageBitmap(), returning a Promise.

explainer.md
```diff
@@ -48,218 +49,298 @@ Provide web apps with efficient access to built-in (software and hardware) media
...
+videoDecoder.configure({codec: 'vp8'}).then((supported) => {
+  // More elaborate fallback recommended.
+  if (!supported)
+    return;
 
-### Example of decode and rendering via TrackWriters (also for low-latency live streaming or cloud gaming)
+  // The app fetches VP8 chunks, feeding one at time to the
+  // provided callback for decoding.
+  streamEncodedChunks(function(newChunk) {
+    videoDecoder.decode(newChunk);
```
sandersdan, line 114: I don't know if this will work, we don't guarantee that each decode will complete without calling flush() or reset().

The simplest backpressure is none; just feed all the input. I'd recommend that with a comment that a real app must handle backpressure too.

The expected implementation is a feed() function that pushes chunks as long as there is input available and the backpressure signal is low. (Maybe we need a backpressure changed event/callback?)

explainer.md
```diff
@@ -48,218 +49,298 @@ Provide web apps with efficient access to built-in (software and hardware) media
...
-encodedVideo.pipeThrough(videoDecoder).pipeTo(renderingStream);
+const videoDecoder = new VideoDecoder({
+  output: paintFrameToCanvas,
+  error: onDecoderError
+});
 
-```
+videoDecoder.configure({codec: 'vp8'}).then((supported) => {
+  // More elaborate fallback recommended.
+  if (!supported)
```
sandersdan, line 108: Unsupported is presumably Promise rejection.

explainer.md
```diff
@@ -48,218 +49,298 @@ Provide web apps with efficient access to built-in (software and hardware) media
...
-  settings: {
-    bitsPerSecond: 1_000_000,
-  },
-});
+  // Build decoders.
+  const audioDecoder = new AudioDecoder({ output: reEncodeAudio, error: onCodecError });
+  const videoDecoder = new VideoDecoder({ output: reEncodeVideo, error: onCodecError });
+
+  // Configure and reset if not supported. More sophisticated fallback recommended.
+  if (!await audioDecoder.configure({ codec: 'aac', ... }))
```
sandersdan, line 231: I assume await throws on rejection?

explainer.md
chcunningham: Agree. Not sure what I was thinking. I've re-worded to:

The WebCodecs interface is modeled on well known platform and software codec APIs. This API shape is time-tested and intuitive for authors with experience in developing media applications.

explainer.md
chcunningham: Sure, I've changed examples to work that way now. 

@mounirlamouri any opinion on whats more idiomatic? EME spec rejects when things aren't supported, while MC resolves with supported=false.

explainer.md
chcunningham: > I don't know if this will work, we don't guarantee that each decode will complete without calling flush() or reset().

AFAIK I'm not relying on that guarantee? 

> The simplest backpressure is none; just feed all the input. I'd recommend that with a comment that a real app must handle backpressure too.

This is the intended meaning of the code I've written. I'm not sure what meaning you've taken, but LMK if I can name/comment to improve clarity. I've added a comment that real app must also handle backpressure (I guess in the low latency case that would mean Reset() and adapting down to a lower resolution).

explainer.md
chcunningham: Yes. Now in try catch

explainer.md
chcunningham: See more recent comment on l.63

chcunningham:
> Perhaps we should reduce the examples here and instead link to an examples directory in the repository?

Happy to. I'll wait for some dust to settle on the example feedback.

explainer.md
sandersdan: videoFrame.transferToImageBitmap(), rather than guaranteeing that an ImageBitmap is always available synchronously. We may prefer getImageBitmap() to imply a copy, I'm not sure about that aspect at this point.

explainer.md
sandersdan: Sorry, I must have misread.

explainer.md
chcunningham: Ah, of course. Done (transferToImageBitmap() for now)

chcunningham:
@padenot lmk if you'd like more time to review. If no one objects I'll split out the examples and land EOD tomorrow.

explainer.md
mounirlamouri: Both EME and MC are correct. What matters is whether failure is expected. In the case of MC, we would expected unsupported configurations all the time but mis-configurations are being rejected. I would assume `configure()` to behave similarly: if the parameters passed in are incorrect, it should reject but if the actual device doesn't support the given configuration, it should probably just say not supported instead of rejecting.

sandersdan: If the parameters cannot be supported then the codec is now in a bad state and must be discarded or recovered. Seems like it should be rejected to me.

mounirlamouri: sounds good 👍

chcunningham:
I'll go ahead and land as is and separate the examples in follow up. Thanks for reviewing.

PR #111, opened 2020-12-09
Add explicit arguments to codec algorithms.

chcunningham:
NOTE: I'm not asking to merge this. I think its hard to read. I'm working 
on an alternative impl (less parameterization, more duplication). We 
should compare them and decide which is prefered. I'll reply here when
option 2 is ready. 

Previously the algorithms assume that any mentioned interface attribute
is implicitly belonging to the codec interface shared by the method that
invoked the algorithm. Unfortunately, that meant none of the bikeshed
autolinking picked up on usage of the interface attributes within the
algorithm.

Now, all interface attributes are explicitly mentioned when invoking the
algorithm, so the linking picks up on this.

I do think it turns out pretty verbose and hard to read :/.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/web-codecs/pull/111.html" title="Last updated on Dec 9, 2020, 12:21 AM UTC (39cc41e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/web-codecs/111/69833fc...39cc41e.html" title="Last updated on Dec 9, 2020, 12:21 AM UTC (39cc41e)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -185,7 +185,8 @@ Methods {#audiodecoder-methods}
     When invoked, run these steps:
     1. If |config| is not a <a>valid AudioDecoderConfig</a>, throw a
         {{TypeError}}.
-    2. Run the <a>Configure Decoder</a> algorithm with |config|.
+    2. Run the <a>Configure Decoder</a> algorithm with |config|,
+        {{AudioDecoder/state}}, and {{AudioDecoder/[[codec implementation]]}}.
   </dd>
 
   <dt><dfn method for=AudioDecoder>decode(chunk)</dfn></dt>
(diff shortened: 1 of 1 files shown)


yoavweiss:
@chcunningham - can you join the WICG to appease the IPR bots?

chcunningham:
Re: IPR - I think I'm already joined? When I go to https://www.w3.org/community/wicg/join it says "You are already a participant in this group."

In an earlier PR the bots were happy on one commit, then unhappy on the next: https://github.com/WICG/web-codecs/pull/109/commits

Nothing changed on my end that I'm aware of. I filed https://github.com/w3c/ash-nazg/issues/188

chcunningham:
Closing in favor of #112

PR #113, opened 2020-12-11
Emit VideoDecoderConfig with VideoEncoder output

### LINKED ISSUE #93: Clean up crop/display size in VideoEncoderConfig
Right now it has only a vague width and height. We should instead have all the crop size members (the pixels in the VideoFrame that actually need encoding). And we should add display size (optional) info for codecs that want to include that in the bitstream.

### LINKED ISSUE #108: Encoders should emit a decoder config describing the encoded stream
We need Encoders to give all the info needed to decode the stream the produce.

For example, if you're VideoEncoder is producing h264 that is _not_ in annex-b format, then we need a means to emit the the avcC details for someone to later be able to decode this stream. 

The plan is for each Encoder to emit a DecoderConfig as part of its output callback. The DecoderConfig should be emitted with the first output frame or any time the config changes. Here's the updated IDL:

callback VideoFrameOutputCallback = undefined(VideoFrame output, **VideoDecoderConfig config**);

callback AudioFrameOutputCallback = undefined(AudioFrame output, **AudioDecoderConfig config**);

chcunningham:
Supplying the VideoDecoderConfig tells VideoEncoder users how to
eventually decode their encoded chunks. Fixes #108.

The EncodedVideoChunkOutputCallback will emit this config alongside the
EncodedVideoChunk anytime the config changes. State is added to detect
changes and fill out the config.

This also updates the sizing info in VideoEncoderConfig to be more
consistent with VideoDecoderConfig (helpful when generating the decoder
config). Fixes #93.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/web-codecs/pull/113.html" title="Last updated on Feb 10, 2021, 5:52 PM UTC (6a107eb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/web-codecs/113/72457ae...6a107eb.html" title="Last updated on Feb 10, 2021, 5:52 PM UTC (6a107eb)">Diff</a>


chcunningham:
@padenot approved the API shape changes in our weekly editors call. Bug filed for follow up on the issue of racing [[active encoder config]]

PR #117, opened 2020-12-21
Add AvcEncoderConfig to specify optional annex-b format.

### LINKED ISSUE #72: Video h264 decoder supports both Annex B and AVCC format
Both are very popular and used in different scenarios. Let application to choose the format during decoder configuration.

### LINKED ISSUE #97: Codec specific configuration parameters
Tracking spec work to allow for codec specific tuning of encoder/decoder configuration. First example of this will be AVC vs Annex-B for H264. 

Probably something like:

```
configure({
  contentType: 'avc....';
  avcOptions: {
     format: "annex-b"
  }
});
```

chcunningham:
Add AvcEncoderConfig to specify annex-b format.

Previously the spec doesn't say which format is used. This clarifies
that the default format is AVC and that the AVCConfigurationBox
(contents of MP4 avcC) is emitted by the encoder output callback.

Furhter, it the ability to instead use annex-b format.

Fixes #72. Fixes #97.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/web-codecs/pull/117.html" title="Last updated on Feb 10, 2021, 8:45 PM UTC (f625d42)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/web-codecs/117/05ff613...f625d42.html" title="Last updated on Feb 10, 2021, 8:45 PM UTC (f625d42)">Diff</a>


index.src.html
```diff
@@ -1201,6 +1213,8 @@
   required unsigned long cropHeight;
   unsigned long displayWidth;
   unsigned long displayHeight;
+
+  AvcEncoderConfig avc;
```
jianjunz, line 1217: Perhaps it could be named as `codecSpecificConfig`, and the type could be `(AvcEncoderConfig or CodecnameEncoderConfig)` in the future.

jianjunz (commented):
Hi,

How about adding a similar item to `VideoDecoderConfig`, so the decoder could be configured to decode Annex B bitstreams.

chcunningham (commented):
> How about adding a similar item to VideoDecoderConfig, so the decoder could be configured to decode Annex B bitstreams.

We do plan to add support support annex-b in VideoDecoder, but via a different mechanism. The idea is that the VideoDecoderConfig.description, when present, will contain the AVCC info and we will assume avc format. If absent, we will assume the AVC stream is in annex-b format.

The assymetric mechanism is a bit weird, but I think its the lesser evil. Think of VideoDecoderConfig.description as essentially the 'extradata' from FFmpeg's AVCodecContext. This is a common notion for all codecs, but the details are codec specific. Like AVC, other codecs including opus, aac, and vorbis will also use this mechanism to provide various metadata (e.g. the opus header, or the vorbis codebook). 

Using description in this way should be familiar and avoids needing to create a decoder-config codec specific structure for each codec. 

For encoding, codec specific configuration seems unavoidable. The avc vs annex-b is just the tip of the iceberg for possible codec-specific encoder configuration parameters.

chcunningham:
@padenot approved these api shape changes in our weekly editors call.

jianjunz:
Thanks for your explanation.

PR #149, opened 2021-03-06
Establish codec registry and AVC registration.

### LINKED ISSUE #126: Spec better codec strings for wav ulaw / alaw
wav files are described with mime types like:
'audio/wav; codecs=1' 

I believe codecs=1 signals 'Microsoft PCM Format', as described in this RFC: https://www.ietf.org/rfc/rfc2361.txt

Other notable digits include 6:ulaw and 7:alaw. Chrome recognizes 1 via canPlayType() and kin. Firefox recognizes more. Haven't checked the other UAs. 

Such codecs strings aren't  very readable, particularly if used in WebCodecs without the accompanying 'audio/wav' prefix. Strawman  solution: revive the idea to have a registry of codecs strings. This will serve to document the nice familiar strings for other codecs and give us an option to set a new precedent for the wav strings (e.g. 'pcm', 'ulaw', 'alaw').

### LINKED ISSUE #143: Make registry for codec strings and decoder config.description (extradata)
Centralized place to see supported codec strings is obviously helpful. It also allows us to fix #126. Furthermore, we should describe what a valid description (extradata) would be for each codec (e.g. AVC -> avcc info) as well as whether the extradata is strictly required (e.g. AVC w/out extradata -> expect annex-b format).

chcunningham:
This moves codec-specific defintions / descriptions into a registry.
I've also added some text to the main document to clarify that support
for any given codec is not required.

Fixes #143. Fixes #126.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/web-codecs/pull/149.html" title="Last updated on Mar 17, 2021, 11:39 PM UTC (49baae7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/web-codecs/149/57e2560...49baae7.html" title="Last updated on Mar 17, 2021, 11:39 PM UTC (49baae7)">Diff</a>

### DIFF
--- a/avc_codec_registration.src.html
+++ b/avc_codec_registration.src.html
@@ -11,10 +11,10 @@ Editor: Paul Adenot, w3cid 62410, Mozilla https://www.mozilla.org/
 Editor: Bernard Aboba, w3cid 65611, Microsoft Corporation https://www.microsoft.com/
 Boilerplate: omit conformance
 
-Abstract: This registration is entered into the [[WebCodecs-Codec-Registry]]. It
-    describes, for AVC (H.264), the (1) fully qualified codec strings, (2) the
-    {{WebCodecs/VideoDecoderConfig.description}} bytes, and (3) the
-    codec-sepcific extensions to the {{WebCodecs/VideoEncoderConfig}}.
+Abstract: This registration is entered into the [[webcodecs-codec-registry]].
+    It describes, for AVC (H.264), the (1) fully qualified codec strings, (2)
+    the {{VideoDecoderConfig.description}} bytes, and (3) the
+    codec-specific extensions to the {{VideoEncoderConfig}}.
 
     The registration is not intended to include any information on whether a
     codec format is encumbered by intellectual property claims. Implementers and
@@ -31,20 +31,25 @@ Markup Shorthands:css no, markdown yes, dfn yes
 </pre>
 
 <pre class='anchors'>
-spec: web-codecs; for: WebCodecs; urlPrefix: https://wicg.github.io/web-codecs/#
+spec: webcodecs; urlPrefix: https://wicg.github.io/web-codecs/#
     type: attribute
         text: VideoDecoderConfig.description; url: dom-videodecoderconfig-description
     type: interface
         text: EncodedVideoChunk; url: encodedvideochunk
         text: VideoEncoder; url: videoencoder
     type: dictionary
-        text: VideoEncoderConfig; url: video-encoder-config
+        text: VideoEncoderConfig; url: dictdef-videoencoderconfig
         text: VideoDecoderConfig; url: dictdef-videodecoderconfig
 </pre>
 
 <pre class='biblio'>
 {
-  "WebCodecs-Codec-Registry": {
+  "webcodecs": {
+      "href": "https://wicg.github.io/web-codecs/",
+      "title": "WebCodecs",
+      "publisher": "WICG"
+  },
+  "webcodecs-codec-registry": {
       "href": "https://wicg.github.io/web-codecs/codec_registry.html",
       "title": "WebCodecs Codec Registry",
       "publisher": "WICG"
@@ -72,17 +77,17 @@ NOTE: The common "avc3." prefix is intentionally not included. This prefix does
 VideoDecoderConfig description {#videodecoderconfig-description}
 ================================================================
 
-A {{WebCodecs/VideoDecoderConfig.description}} may or may not be required by the
+A {{VideoDecoderConfig.description}} may or may not be required by the
 AVC codec depending on the bitstream format of the AVC content to be decoded.
 
-If the {{WebCodecs/VideoDecoderConfig.description}} includes an
+If the {{VideoDecoderConfig.description}} includes an
 `AVCDecoderConfigurationRecord`, as defined by [[iso14496-15]], section 5.3.3.1,
 it will be assumed that the bitstream is in "avc" format.
 
 NOTE: This format is commonly used in .MP4 files, where the player generally has
     random access to the media data.
 
-If the {{WebCodecs/VideoDecoderConfig.description}} is not provided, it will be
+If the {{VideoDecoderConfig.description}} is not provided, it will be
 assumed that the bitstream is in “annexb” format.
 
 NOTE: "annexb" format is described in greater detail by [[ITU-T-REC-H.264]],
@@ -95,7 +100,7 @@ VideoEncoderConfig extensions {#videoencoderconfig-extensions}
 
 <pre class='idl'>
 <xmp>
-/* partial */ dictionary VideoEncoderConfig {
+partial dictionary VideoEncoderConfig {
   AvcEncoderConfig avc;
 };
 </xmp>
@@ -121,7 +126,7 @@ dictionary AvcEncoderConfig {
 <dl>
   <dt><dfn dict-member for=AvcEncoderConfig>format</dfn></dt>
   <dd>
-    Configures the format of output {{WebCodecs/EncodedVideoChunk}}s. See
+    Configures the format of output {{EncodedVideoChunk}}s. See
     {{AvcBitstreamFormat}}.
   </dd>
 </dl>
@@ -157,9 +162,9 @@ SPS and PPS are described in greater detail in sections G.3.41 and G.3.55 of
   <dt><dfn enum-value for=AvcBitstreamFormat>avc</dfn></dt>
   <dd>
     SPS and PPS data are not included in the bitstream and are instead emitted
-    via the {{WebCodecs/VideoEncoder}} [[output callback]] as the
-    {{WebCodecs/VideoDecoderConfig.description}} of the
-    {{WebCodecs/VideoDecoderConfig}}.
+    via the {{VideoEncoder}} [[output callback]] as the
+    {{VideoDecoderConfig.description}} of the
+    {{VideoDecoderConfig}}.
     <var ignore=''>output_config</var>.
 
     NOTE: This format is described in greater detail by [[iso14496-15]],
--- a/avc_codec_registration.src.html
+++ b/avc_codec_registration.src.html
@@ -2,7 +2,7 @@
 Title: AVC (H.264) WebCodecs Registration
 Repository: wicg/web-codecs
 Status: CG-DRAFT
-Shortname: web-codecs-avc-codec-registration
+Shortname: webcodecs-avc-codec-registration
 Level: none
 Group: wicg
 ED: none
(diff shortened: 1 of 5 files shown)


chcunningham:
@tidoust FYI - this is what I'd like to land before the FPWD transition.

chcunningham:
@aboba - looks like the pr-preview link at the bottom of the description only supports a preview of the main spec ([known issue](https://github.com/tobie/pr-preview/issues/18)). Here are some links to the generated HTML to assist in review. 

https://storage.googleapis.com/videostack-external/webcodecs_pr_149/index.html
https://storage.googleapis.com/videostack-external/webcodecs_pr_149/codec_registry.html
https://storage.googleapis.com/videostack-external/webcodecs_pr_149/avc_codec_registration.html

For now I've just done the AVC registration. Other registrations will be follow up PRs (not urgent, not blocking anything).

chcunningham:
@padenot FYI

avc_codec_registration.src.html
```diff
@@ -0,0 +1,168 @@
...
+    Annex B. This format is commonly used in live-streaming applications, where
+    including the SPS and PPS data periodically allows users to easily start
+    from the middle of the stream.
+
+VideoEncoderConfig extensions {#videoencoderconfig-extensions}
+==============================================================
+
+<pre class='idl'>
+<xmp>
+/* partial */ dictionary VideoEncoderConfig {
```
chcunningham, line 98: @tabatkins, the partial above is in a comment because I couldn't figure out how make bikeshed aware that this extends a definition from another file in this repository (index.src.html). My understanding from reading the documentation is that all idl definitions are automatically exported such that this should just work?

codec_registry.src.html
```diff
@@ -0,0 +1,172 @@
...
+    type: attribute
+        text: AudioDecoderConfig.description; url: dom-audiodecoderconfig-description
+        text: VideoDecoderConfig.description; url: dom-videodecoderconfig-description
+        text: VideoEncoderConfig; url: video-encoder-config
+        text: AudioEncoderConfig; url: audio-encoder-config
+</pre>
+
+<pre class='biblio'>
+{
+  "WebCodecs": {
```
tidoust, line 47: The other document has "WEB-CODECS". In general, it is good practice to use the same name as the spec's shortname, and good practice to use the repo name as spec shortname, so "web-codecs". If you would like to change all names to "webcodecs", now would be a good time as we transition the repo and spec to the W3C org.

codec_registry.src.html
```diff
@@ -0,0 +1,172 @@
...
+    This registry is non-normative.
+
+Markup Shorthands:css no, markdown yes, dfn yes
+!Participate: <a href="https://github.com/wicg/web-codecs">Git Repository.</a>
+!Participate: <a href="https://github.com/wicg/web-codecs/issues/new">File an issue.</a>
+!Version History: <a href="https://github.com/wicg/web-codecs/commits">https://github.com/wicg/web-codecs/commits</a>
+</pre>
+
+<pre class='anchors'>
+spec: web-codecs; for: WebCodecs; urlPrefix: https://wicg.github.io/web-codecs/#
```
tidoust, line 37: Same comment on the `for` as in the other document.

tidoust (commented):
See some comments inline.

I understand that the current final goal (pending evolutions of the W3C process to account for registries) is to publish the registry and the AVC registration as non-normative Working Group Notes. Is that correct?

chcunningham (commented):
> I understand that the current final goal (pending evolutions of the W3C process to account for registries) is to publish the registry and the AVC registration as non-normative Working Group Notes. Is that correct?

Yes, that is correct. And we will seek to maintain these as non-normative as the W3C process evolves.

codec_registry.src.html
tidoust: I was more thinking of string match regardless of case rather than exact match. I do not think that there is a written recommendation about the caps style for references (for instance, I don't see any in the [Manual of Style](https://w3c.github.io/manual-of-style/#References)). A majority of specs use upper case for biblio refs. I personally prefer that too, but up to you!

To be clear, there are no written guidelines that I'm aware of to recommend that biblio refs should use the shortname of the spec, which should also match the repo name. Given that naming is notoriously hard, it just seems a good idea to be consistent.

Noted for the change of name. Let's drop the `-` during repo transition!

tidoust:
> > I understand that the current final goal (pending evolutions of the W3C process to account for registries) is to publish the registry and the AVC registration as non-normative Working Group Notes. Is that correct?
> 
> Yes, that is correct. And we will seek to maintain the non-normative status of the registry and registrations as the W3C process evolves.

From a process perspective, this means that the specs need to be published separately. That is, we're going to publish three First Public Working Drafts:
1. The WebCodecs spec in itself, intended to become a Recommendation, under the shortname `webcodecs`
2. The registry spec, intended to become a WG Note, under the shortname `webcodecs-codec-registry`
3. The AVC (H.264) WebCodecs registration, intended to become a WG Note, under the shortname `webcodecs-avc-codec-registration`.

Practically speaking, tools such as PR preview and spec-prod are more or less designed for repos that contain only one spec. Splitting the specs to dedicated repositories as done for MSE (see [list of repos](https://github.com/w3c/media-source)) would simplify integration with these tools... That said, it also makes editing slightly more cumbersome, and it's certainly possible to workaround tools constraints.

Would you be open to creating dedicated repos for the registry and AVC WebCodecs registration specs? If so, two choices:
1. I can create the registry repos under w3c right away, initialize them with the appropriate content taken from this PR, and corresponding files can then be dropped from this PR before it gets merged. Good thing about this is that it means we'll keep this repo clean.
2. You'd prefer that PR to be merged before the actual repo transition. In that case, we'll just separate documents afterwards, but note that would mean we'll need to preserve the `codec_registry.html` and `avc_codec_registration.html` files in this repo, replacing them with empty HTML files with a meta redirect to their new locations (as also done in MSE).

I would do 1. and I'm happy to prepare that.

aboba:
Question: What is the approval process for registry entries?  The document mentions posting an Issue to the github. Is the expectation that the specification editors will do the review/approval on their own? Or that they will recommend a disposition to the WG who will then make the decision?  Also, what happens if for some reason the Github repo is no longer active?  IANA registries sometimes allow a successor mailing list or WG to be designated.

chcunningham:
> Question: What is the approval process for registry entries? The document mentions posting an Issue to the github. Is the expectation that the specification editors will do the review/approval on their own? Or that they will recommend a disposition to the WG who will then make the decision? Also, what happens if for some reason the Github repo is no longer active? IANA registries sometimes allow a successor mailing list or WG to be designated.

@tidoust @cwilso - pls weigh in on ^

tidoust:
> Question: What is the approval process for registry entries? The document mentions posting an Issue to the github. Is the expectation that the specification editors will do the review/approval on their own? Or that they will recommend a disposition to the WG who will then make the decision? Also, what happens if for some reason the Github repo is no longer active? IANA registries sometimes allow a successor mailing list or WG to be designated.

First, we don't yet have a clear and crisp process in place for managing registries at W3C. That is the purpose of [ongoing discussions within the Process CG](https://github.com/w3c/w3process/issues?q=is%3Aissue+is%3Aopen+registries). See also the registries section of the presentation on [Process 2021 during last TPAC](https://www.w3.org/2020/10/TPAC/talk/Process2021#talk). Hopefully, a future revision of the W3C Process will create a clear mechanism for registries at W3C.

The direction that this PR is taking is that the registry will be published as a WG Note. The Media WG would thus be in charge of updating it (and not only the editors). For maintenance purpose, W3C tends to keep WGs around forever. If we ever close the group, the W3C team would be in fine responsible for the maintenance of the registry (granted, in the W3C process, this is more aimed at [maintenance of Recommendations](https://www.w3.org/2020/Process-20200915/#rec-amended) than at the maintenance of Notes...).

Another group may be created to take over the maintenance of the registry if needed. One perfect illustration is the Media WG, which took over maintenance of MSE/EME and of their companion format registries from the HTML Media Extensions WG.

I do not know whether these possibilities need to be specified in the registry spec itself. The text in the requirements section of the WebCodecs registry seems to be based on the one of the [MSE Byte Stream Format Registry](https://www.w3.org/TR/mse-byte-stream-format-registry/)).

W3C guarantees that archives of mailing-lists will be available forever. If you believe that this guarantee is crucial, it may be preferable to request that an email be sent to some W3C archived public mailing-list (as done for MSE/EME registries) rather than to request that an issue gets posted to a GitHub repository. That said, my understanding here is that the contents of the issue (codec string, common name, link to a public spec) will be integrated into the registry directly, so is there really a need for a strong guarantee with regards to availability of the issue over time? More precisely, is there a need for a stronger guarantee than the one that currently exists for all group discussions that take place in GitHub issue trackers?

codec_registry.src.html
chcunningham: Thanks. Updated all to use UPPER case.

chcunningham: Thanks. Updated all to use UPPER case.

chcunningham:
> From a process perspective, this means that the specs need to be published separately. That is, we're going to publish three First Public Working Drafts:
> 
> The WebCodecs spec in itself, intended to become a Recommendation, under the shortname webcodecs
> The registry spec, intended to become a WG Note, under the shortname webcodecs-codec-registry
> The AVC (H.264) WebCodecs registration, intended to become a WG Note, under the shortname webcodecs-avc-codec-registration.

LGTM

> Practically speaking, tools such as PR preview and spec-prod are more or less designed for repos that contain only one spec.
> ...
> Would you be open to creating dedicated repos for the registry and AVC WebCodecs registration specs?

I don't love it. WebCodecs is going to have a large number of "registrations" (see TODO's in the registry table), and having a git repo for each one sounds cumbersome. 

It looks like spec-prod [intends to remedy this soon](https://github.com/w3c/spec-prod/issues/31). Reading the replies there, I conclude that this is widespread enough that the tooling should be updated. Meanwhile, am I correct that the current travis config happily supports multiple files (I think it publishes the entire out/ folder)?

Multi-spec PR preview is more of a nice-to-have since the workaround is trivial. The analogous [feature request](https://github.com/tobie/pr-preview/issues/18) has less momentum. [The maintainer's comments](https://github.com/tobie/pr-preview/issues/19#issuecomment-414052151) suggest its just a matter of staffing it. 

> The text in the requirements section of the WebCodecs registry seems to be based on the one of the MSE Byte Stream Format Registry.

That's correct. LMK if you find this in any way lacking. 

> More precisely, is there a need for a stronger guarantee than the one that currently exists for all group discussions that take place in GitHub issue trackers?

I think GitHub should be fine.

chcunningham:
@aboba sounds like the text around process for registry updates is reasonable. Any other concerns that block approval?

aboba (approved):

chcunningham:
Thanks everyone!

PR #152, opened 2021-03-17
Add Image Decoding, associated interfaces and algorithms

### LINKED ISSUE #50: Support for images
Is there a path for using this proposal to encode/decode images?

I'm suspecting it's possible by implementing the image container and somehow getting the binary keyframe out, but how to plumb this is somewhat unclear.

Given that a lot of developers have been asking for functionality to encode/decode images without using an HTMLImageElement as an intermediary, it would be nice to have this use case covered.

chcunningham:
Fixes #50.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/pull/152.html" title="Last updated on May 3, 2021, 9:02 PM UTC (8b7c907)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/152/8b69591...8b7c907.html" title="Last updated on May 3, 2021, 9:02 PM UTC (8b7c907)">Diff</a>


chcunningham:
@dalecurtis, mind taking a first pass? Interface wise, this is everything we discussed. But behind those friendly interfaces hides quite a bit of state. 

@aboba, FYI - will request review formally once Dale has had a go. 
@padenot FYI

index.src.html
```diff
@@ -2394,14 +2420,692 @@
...
+    4. Wait for any of the following conditions to be true (whichever happens
+        first):
+        1. {{ImageDecoder/[[encoded data]]}} contains enough bytes to
+            completely decode |encodedFrame|.
+        2. {{ImageDecoder/complete}} is `true`.
+        3. {{ImageDecoder/[[fatal error]]}} is `true`.
+    5. If {{ImageDecoder/[[encoded data]]}} does not contain enough bytes to
+        completely decode |encodedFrame|:
+        1. Remove |promise| from {{ImageDecoder/[[pending decode promises]]}}.
+        2. Reject |promise| with {{InvalidStateError}} {{DOMException}}.
```
dalecurtis, line 2799: Not sure about InvalidStateError, suggestions welcome.

index.src.html
```diff
@@ -2394,14 +2420,692 @@
...
+        `-1`.
+    3. Let |encodedFrame| be the encoded frame identified by |frameIndex| and
+        {{ImageDecoder/[[internal selected track index]]}}.
+    4. Wait for any of the following conditions to be true (whichever happens
+        first):
+        1. {{ImageDecoder/[[encoded data]]}} contains enough bytes to
+            completely decode |encodedFrame|.
+        2. {{ImageDecoder/complete}} is `true`.
+        3. {{ImageDecoder/[[fatal error]]}} is `true`.
+    5. If {{ImageDecoder/[[encoded data]]}} does not contain enough bytes to
```
dalecurtis, line 2796: Probably meant to say if |complete| before this section? Maybe the text needs a FALLTHROUGH macro :p

index.src.html
```diff
@@ -2394,14 +2420,692 @@
...
+        `-1`.
+    3. Let |encodedFrame| be the encoded frame identified by |frameIndex| and
+        {{ImageDecoder/[[internal selected track index]]}}.
+    4. Let |lastFrameGeneration| be `null`.
+    5. If {{ImageDecoder/[[progressive frame generations]]}} contains a map
+        entry with the key |frameIndex|, assign the value of the map entry to
+        |lastFrameGeneration|.
+    6. Wait for any of the following conditions to be true (whichever happens
+        first):
+        1. {{ImageDecoder/[[encoded data]]}} contains enough bytes to decode
```
dalecurtis, line 2833: The current implementation just waits until _any_ new bytes have been received and retries to see if a new image can be decoded. Not sure if that's easier to phrase here.

index.src.html
```diff
@@ -2394,14 +2420,692 @@
...
+            |encodedFrame| to produce an output who's [=Progressive Image
+            Frame Generation=] does not equal |lastFrameGeneration|.
+        2. {{ImageDecoder/complete}} is `true`.
+        3. {{ImageDecoder/[[fatal error]]}} is `true`.
+    7. If {{ImageDecoder/[[encoded data]]}} does not contain enough bytes to
```
dalecurtis, line 2838: I'm not sure this is right, if we reach complete and and there are no more successive frames decoded and there's no error then by definition that frame is the 'complete frame'.

index.src.html
```diff
@@ -2394,14 +2420,692 @@
...
+        progressively decode |encodedFrame| to produce an output who's
+        [=Progressive Image Frame Generation=] does not equal
+        |lastFrameGeneration|:
+        1. Remove |promise| from {{ImageDecoder/[[pending decode promises]]}}.
+        2. Reject |promise| with {{InvalidStateError}} {{DOMException}}.
+        3. Abort these steps.
+    8. Attempt to use {{ImageDecoder/[[codec implementation]]}} to decode
```
dalecurtis, line 2845: Can you just say something like run steps A..B in the algorithm above instead of duplicating here?

index.src.html
```diff
@@ -2394,14 +2420,692 @@
...
+    15. Remove |promise| from {{ImageDecoder/[[pending decode promises]]}}.
+    16. Resolve |promise| with |decodeResult|.
+
+: <dfn for=ImageDecoder>Check Type Support</dfn> (with |type|)
+:: 1. If the user agent can provide a codec to support decoding |type|, return
+        `true`.
+    2. Otherwise, return `false`.
+
+: <dfn for=ImageDecoder>Handle Error</dfn> (with |exception|)
+:: 1. Assign `true` to {{ImageDecoder/[[fatal error]]}}.
```
dalecurtis, line 2877: Does this need checking above in the decode and decodeMetadata algorithms?

index.src.html
```diff
@@ -2394,14 +2420,692 @@
...
+ImageTrackList Interface {#imagetracklist-interface}
+----------------------------------------------------
+<pre class='idl'>
+<xmp>
+interface ImageTrackList : EventTarget {
+  getter ImageTrack (unsigned long index);
+
+  readonly attribute unsigned long length;
+  readonly attribute long selectedIndex;
+
```
dalecurtis, line 2983: `readonly attribute ImageTrack? selected;` or possibly `current` to avoid weirdness like

`decoder.tracks.selected.selected = true;`

index.src.html
```diff
@@ -2394,14 +2420,692 @@
...
+:: Fired at the {{ImageTrackList}} when the list of {{ImageTrack}}s is
+    established or when attributes of the listed {{ImageTrack}}s are updated.
+
+ImageTrack Interface {#imagetrack-interface}
+--------------------------------------------
+<pre class='idl'>
+<xmp>
+interface ImageTrack {
+  readonly attribute boolean animated;
+  readonly attribute long frameCount;
```
dalecurtis, line 3024: unsigned long?

index.src.html
```diff
@@ -2394,14 +2420,692 @@
...
+  readonly attribute unrestricted float repetitionCount;
```
dalecurtis, line 3025: As noted above, maybe we only want to represent discrete counts and leave everything else as zero.

index.src.html
```diff
@@ -2394,14 +2420,692 @@
...
+:: The {{ImageTrack/repetitionCount}} getter steps are to return the value of
+    {{ImageTrack/[[repetition count]]}}.
+
+: <dfn attribute for=ImageTrack>selected</dfn>
+:: The {{ImageTrack/selected}} getter steps are to return the value of
+    {{ImageTrack/[[selected]]}}.
+
+    The {{ImageTrack/selected}} setter steps are:
+    1. Let |newValue| be [=the given value=].
+    2. If |newValue| equals {{ImageTrack/[[selected]]}}, abort these steps.
```
dalecurtis, line 3076: I think this needs to an abort step if the new value is 'false'. We don't want there to let there be no selected track.

Or is false just supposed to rerun the track selection algorithm?

index.src.html
```diff
@@ -2394,14 +2420,692 @@
...
+<pre class='idl'>
+<xmp>
+[Exposed=(Window,DedicatedWorker)]
+interface ImageDecoder {
+  constructor(ImageDecoderInit init);
+
+  readonly attribute boolean complete;
+  readonly attribute ImageTrackList tracks;
+
+  Promise<undefined> decodeMetadata();
```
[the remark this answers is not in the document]
aboba: Undefined is the new usage.

index.src.html
```diff
@@ -2394,14 +2420,692 @@
...
+    8. Assign |selectedTrackIndex| to {{ImageDecoder/[[internal selected track
+        index]]}}.
+    9. Assign `true` to {{ImageDecoder/[[tracks established]]}}.
+    10. Queue a task on the [=control thread=] event loop to perform the
+        following steps:
+        1. Assign |newTrackList| to the {{ImageDecoder/tracks}}
+            {{ImageTrackList/[[track list]]}} internal slot .
+        2. Assign |selectedTrackIndex| to {{ImageDecoder/tracks}}
+            {{ImageTrackList/[[selected index]]}}.
+        3. Fire a simple event named {{ImageTrackList/change}} at the
```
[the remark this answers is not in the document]
chcunningham: I think making ImageTrack have its own event is a good update. Maybe 'addframes' or 'framecountchanged'.

[Looking closer at VideoTrackList](https://html.spec.whatwg.org/multipage/media.html#event-media-change), it seems 'change' should only be fired when we enable/disable, so I'll continue to use it for that purpose.

```
change: One or more tracks in the track list have been enabled or disabled.
```

Also, it seems that 'changed' is not really the right event to throw when we add tracks. [\<video\>.tracks would use 'addtrack'](https://html.spec.whatwg.org/multipage/media.html#loading-the-media-resource:trackevent). So [this step](https://pr-preview.s3.amazonaws.com/WICG/web-codecs/pull/152.html#ref-for-eventdef-imagetracklist-change) in the Establish Tracks algo should be ammended to use addtrack.

index.src.html
```diff
@@ -2394,14 +2420,692 @@
...
+                  {{ImageDecoderInit/data}} as a {{ReadableStream}}, the
+                  {{ImageTrack/frameCount}} may change as additional bytes are
+                  appended to {{ImageDecoder/[[encoded data]]}}. See the
+                  [=Update Tracks=] algorithm.
+
+            5. If |image track| is found to describe a repetition count,
+                assign that count to {{ImageTrack/[[repetition count]]}}
+                internal slot. Otherwise, assign `0`.
+
+                NOTE: A value of `Infinity` indicates infinite repetitions.
```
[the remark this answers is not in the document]
chcunningham: Pulling in your comment here: https://bugs.chromium.org/p/chromium/issues/detail?id=1183456#c2

> To handle infinity we can either use zero like I do today or make the repetitionCount an optional field which isn't present when the value == not specified or infinity.

Overloading zero isn't great since zero is also a valid repetition count. IDL-wise, I'm only able to make dictionary members optional (ImageTrack is an interface). For interfaces the closest is probably nullable (`unsigned long? repetitionCount`). But I'm not in love with "null means zero, and zero means infinity". Do you like that better than float repetitionCount? I appreciate that it's weird to use float purely for the sake of including Infinity, but in practice I don't think this will ever cause anyone pain.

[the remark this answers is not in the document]
dalecurtis: Is zero valid? I think we should just require once to have a value of 1 and define anything other than that as zero. I think it's obvious that animated = true and repetitionCount = 0, means up to you.

index.src.html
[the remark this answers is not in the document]
dalecurtis: Yes I mentioned this in our last 1:1, but after thinking about it more I think just 'change' on the ImageTrack seems fine. I think it's overkill and potentially limiting to give each field a change listener... The html change event says:

`when an alteration to the element's value is committed by the user`

So I think that wording seems fine to continue using there.

'addtrack' on the ImageTrackList is definitely seems clearer though. I don't think having both 'addtrack' and 'change' on the ImageTrackList is necessary. Unlike audio track list where multiple tracks may be enabled, for images it'll always be one all track enable/disable actions are user controlled.

index.src.html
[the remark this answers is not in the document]
chcunningham: To me zero implies "play once through". If the name were playCount rather than repetitionCount, I could see your meaning.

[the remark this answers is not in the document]
dalecurtis: Let's just call playCount then? Maybe make it optional if it's infinity or not specified?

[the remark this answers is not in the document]
dalecurtis: I.e., let's go with simple meaning versus exporting complexity

index.src.html
chcunningham: Currently false may cause no tracks to be selected. I think this is possible with Audio|VideoTrackList, so just followed that model. It is an error to call decode() while in that state, checked on step 2 of decode()
```
If [[ImageTrackList]]'s [[selected index]] is '-1', return a Promise rejected with an InvalidStateError DOMException.
```

chcunningham (commented):
Thanks @dalecurtis, great feedback.

index.src.html
```diff
@@ -2394,14 +2420,692 @@
...
+            {{ImageDecoderInit/data}}.
+        5. In parallel, perform the [=Fetch Stream Data Loop=] on |d| with
+            |reader|.
+    12. Otherwise:
+        1. Assert that `init.data` is of type {{BufferSource}}.
+        2. Assign a copy of `init.data` to  {{ImageDecoder/[[encoded data]]}}.
+        3. Assign `true` to {{ImageDecoder/complete}}
+        4. Queue a control message to [=configure the image decoder=] with
+            |init|.
+        5. Queue a control message to [=decode track metadata=].
```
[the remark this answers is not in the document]
dalecurtis: That's what the Chrome decoders do, so that sgtm.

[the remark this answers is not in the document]

chcunningham:
@cconcolato, we're interested to get your take how [ImageDecoder](https://pr-preview.s3.amazonaws.com/WICG/web-codecs/pull/152.html#imagedecoder-interface) describes a list of of ImageTracks (ImageDecoder.tracks). We think it maps well onto video-based image formats like avif (given that "tracks" is a longstanding concept for video), but there's some debate about whether it is over engineered. That is, do formats like avif support an arbitrary number of tracks? If we instead expect all image formats to define at most 2 tracks (one animated, one still), then we could simplify the API to remove all mention of tracks and make selections simply by giving a value for [preferAnimation](https://pr-preview.s3.amazonaws.com/WICG/web-codecs/pull/152.html#dom-imagedecoderinit-preferanimation).

index.src.html
dalecurtis: Do we want EncodingError for fatal error?

index.src.html
dalecurtis: Sounds like yes you did mean to use the spec equivalent of FALLTHROUGH :)

chcunningham:
@aboba, I think Dale's review is sufficiently concluded for you to begin. Standing by for questions and feedback.

cconcolato:
> @cconcolato, we're interested to get your take how [ImageDecoder](https://pr-preview.s3.amazonaws.com/WICG/web-codecs/pull/152.html#imagedecoder-interface) describes a list of of ImageTracks (ImageDecoder.tracks). We think it maps well onto video-based image formats like avif (given that "tracks" is a longstanding concept for video), but there's some debate about whether it is over engineered. That is, do formats like avif support an arbitrary number of tracks? If we instead expect all image formats to define at most 2 tracks (one animated, one still), then we could simplify the API to remove all mention of tracks and make selections simply by giving a value for [preferAnimation](https://pr-preview.s3.amazonaws.com/WICG/web-codecs/pull/152.html#dom-imagedecoderinit-preferanimation).

@chcunningham Not sure I have enough information to answer. Let me try.
First, I'm a bit confused when you say "at most 2 tracks (one animated, one still)". In AVIF or HEIF-based formats, there are 2 distinct concepts: image items and image sequence tracks. One image item represents one image (ignoring scalable, layered image items). A image sequence track contains multiple images that are meant to be displayed in sequence, following some optional timing information (duration, and loop). Of course, if it helps the design of the API and the implementations, you could expose an item as a single-image track. 

Then, an ISOBMFF file can indeed contain many tracks. For example, an mp4 file could contain an AV1 video track, an audio track, a subtitle track. It could also contain an image sequence track (e.g. using only the key frames of the video track) meant to be viewed as a GIF-like animation before the video is clicked. It could also contain one image item to give a representative image of the video. Theoretically, you can construct files with as many tracks and items as you want. It could have N video tracks, K audio tracks, P image sequence tracks, Q subtitle tracks, R image items, etc... In practice, typical files will have simple configurations. 

Reading:
> [[animated]]
> Indicates whether this track contains an animated image with multiple frames.
> [[frame count]]
> The number of frames in this track.
> [[repetition count]]
> The number of times the animation is intended to repeat.

It seems to me that you want to expose only image items and image sequences (which is fine). I'm curious how an implementation is supposed to decide if it exposes a sequence of images as a video track or as an image track. Do you expect the track container to guide the implementation? For example, in ISOBMFF, video tracks and image sequence tracks are differentiated by the track handler `vide` vs. `pict`.

In practice, I don't think files containing images will contain more than one image sequence. As for image items, there are use cases where people think about storing multiple image items in the same file. This is for example because they are the result of a capture burst, or bracketed images, or multi-angle, multi-view images, or even to package multiple resolutions in the same file. But these use cases are rather rare IMHO. If you want to keep the API simple for now, the preferAnimation approach seems reasonable and matches the hypothetical reader API that MIAF defines:

> Inputs to a MIAF reader are:
> — a file with a FileTypeBox containing at least one brand specified in this document;
> — optionally one of the following:
>  — item_ID of the item to be output (psItemId),
>  — track_ID of the track to be output (psTrackId),
> — a selection between a static image (psImagePreferredFlag equal to 1) or track (psImagePreferredFlag equal to 0) to be output,
> — optionally constraints, such as the maximum width and height of an image item or track;
> — optionally one or more of the following roles of the image or track to be output:
> — master (default),
> — thumbnail,
> — auxiliary, which may be further classified by the type.

Maybe you want to consider adding preferThumbnail?

@joedrago may have additional feeback based on the libavif API and its integration in browsers
@dwsinger may have additional feedback based on ISOBMFF/HEIF/MIAF specs.

index.src.html
```diff
@@ -2394,14 +2420,812 @@
...
+    3. For each |decodePromise| in
+        {{ImageDecoder/[[pending decode promises]]}}:
+        1. Reject |decodePromise| with |exception|.
+        2. Remove |decodePromise| from
+            {{ImageDecoder/[[pending decode promises]]}}.
+
+: <dfn for=ImageDecoder>Close ImageDecoder</dfn> (with |exception|)
+:: 1. Run the [=ImageDecoder/Reset ImageDecoder=] algorithm with |exception|.
+    1. Assign `true` to {{ImageDecoder/[[closed]]}}.
+    2. Clear {{ImageDecoder/[[codec implementation]]}} and release associated
```
chcunningham, line 2958: I should add an additional step here to close the ImageTrackList as well (needs internal slots in ImageTrack to indicate they've been detached in case any one still holds a ref).

chcunningham:
Thank you @cconcolato!

> It seems to me that you want to expose only image items and image sequences (which is fine).

Yes. And I'm calling both a "track", where an image items simply has one frame. 

> I'm curious how an implementation is supposed to decide if it exposes a sequence of images as a video track or as an image track. Do you expect the track container to guide the implementation? For example, in ISOBMFF, video tracks and image sequence tracks are differentiated by the track handler vide vs. pict.

When you say "video track", I take it you mean an ImageTrack for which track.animated=true. If so, the my intent with animated is to provide an early signal that this track will have a frameCount > 1. Ideally we would do away with animated entirely and just have frameCount, but frameCount is not always known at the outset (particularly for gif), so this may cause folks to  prematurely consider a track with 1 frame as non-animated.

> This is for example because they are the result of a capture burst, or bracketed images, or multi-angle, multi-view images, or even to package multiple resolutions in the same file. But these use cases are rather rare IMHO. If you want to keep the API simple for now, the preferAnimation approach seems reasonable and matches the hypothetical reader API that MIAF defines:

An earlier draft did just have preferAnimation without the tracks mechanism. But then we had the frameCount and animated properties directly on ImageDecoder. This works, but its limiting if we later do want to add some description of alternative tracks. I noticed that html has long defined AudioTrackList and VideoTrackList interfaces, so this seemed like a pattern we might follow with ImageTrack(List).

> Maybe you want to consider adding preferThumbnail?

Could do. What happens when preferThumbnail and preferAnimated compete? Maybe prefer should be an enum w/ either type?

index.src.html
```diff
@@ -2394,14 +2420,692 @@
...
+: <dfn for=ImageDecoder>Establish Tracks</dfn>
+:: Run these steps:
+    1. Assert {{ImageDecoder/[[tracks established]]}} is `false`.
+    2. If {{ImageDecoder/[[encoded data]]}} does not contain enough data to
+        determine the number of tracks, abort these steps.
+    3. If the number of tracks is found to be '0', run the
+        [=ImageDecoder/Handle Error=] algorithm and abort these steps.
+    4. Let |selectedTrackIndex| be `-1`.
+    5. Let |newTrackList| be a new [=list=].
+    6. For each |image track| found in {{ImageDecoder/[[encoded data]]}}:
```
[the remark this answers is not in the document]
dalecurtis: Lets leave it simple as is for now. I think animated and frameCount preclude a lot of stuff anyways.

index.src.html
```diff
@@ -2394,14 +2420,812 @@
...
+    [=Running a control message=] to <dfn for=ImageDecoder>decode track
+    metadata</dfn> means running these steps:
+    1. Run the [=ImageDecoder/Establish Tracks=] algorithm.
+
+: <dfn method for=ImageDecoder>decode(options)</dfn>
+:: Enqueues a control message to decode the frame according to |options|.
+
+    When invoked, run these steps:
+    1. If {{ImageDecoder/[[closed]]}} is `true`, return a {{Promise}}
+        rejected with an {{InvalidStateError}} {{DOMException}}.
```
chcunningham, line 2634: TODO: consider throwing a NotSupportedError here if the reason for close was unsupported mime type. Generally, I think we should save the exception that was used when running the close algorithm and re-use in cases such as this.

cconcolato:
> When you say "video track", I take it you mean an ImageTrack for which track.animated=true.

No I really meant a VideoTrack. This spec defines an ImageTrack and that the HTML spec defines a VideoTrack. When both will be implemented, and the browser is presented with an MP4 video track or MP4 image sequence track, how will it decide if it uses a VideoTrack or an ImageTrack?

> Could do. What happens when preferThumbnail and preferAnimated compete? Maybe prefer should be an enum w/ either type?

Maybe, no strong opinion.

chcunningham:
> No I really meant a VideoTrack. This spec defines an ImageTrack and that the HTML spec defines a VideoTrack. When both will be implemented, and the browser is presented with an MP4 video track or MP4 image sequence track, how will it decide if it uses a VideoTrack or an ImageTrack?

Ah, I follow. Do you expect to see files like this in the wild? How often? Would it be reasonable to describe these with the image/* mimetype (vs video/*)?

cconcolato:
I can't predict the future, but I could envisage people creating dual-headed files (with an image sequence track and a video track, possibly sharing coded frames) to be used in both  and <video>. The same file could be served with either MIME type.

chcunningham:
> I can't predict the future, but I could envisage people creating dual-headed files (with an image sequence track and a video track, possibly sharing coded frames) to be used in both  and

Thanks. Probably you meant `<video>` and ``. I think its reasonable for us to draw the lines like so:
* ImageDecoder ignores the VideoTracks in these files. If you're using ImageDecoder, we assume you want image things. 
* You can still decode the video content for such files using WebCodecs, but this requires the app perform the typical demuxing and creation of {{EncodedVideoChunk}}s

chcunningham:
@aboba, just checking in - any feedback? @padenot I know your feedback is still being compiled.

index.src.html
```diff
@@ -2394,14 +2420,812 @@
...
+
+  readonly attribute boolean complete;
+  readonly attribute ImageTrackList tracks;
+
+  Promise<undefined> decodeMetadata();
+  Promise<ImageDecodeResult> decode(optional ImageDecodeOptions options);
+  undefined reset();
+  undefined close();
+
+  static Promise<boolean> isTypeSupported(DOMString type);
```
mathiasbynens, line 2466: Why is this a promise-based API? Can we instead return the boolean synchronously? https://github.com/dalecurtis/image-decoder-api/issues/6

mathiasbynens: I understand that _actual image decoding_ might depend on another process, but answering the question "do I know what to do with this image type at all (without necessarily doing the work)" seems orthogonal to that. Couldn't browsers just maintain a list of supported image types they support and synchronously check it whenever `isTypeSupported` is called?

If not, then have we considered changing the name of this API? IMHO it's surprising to give it the same name as an existing API without matching the signature of its return value.

chcunningham: > I understand that actual image decoding might depend on another process, but answering the question "do I know what to do with this image type at all (without necessarily doing the work)" seems orthogonal to that. Couldn't browsers just maintain a list of supported image types and synchronously check it whenever isTypeSupported is called?

If the browser entirely relies on the OS to provide the codec, it may not be possible to know statically what codecs are supported (particularly for newer formats). Instead, we may be forced to query OS apis that are adjacent to the actual decoding APIs. Often this involves the same async IPC to a privaledged sandboxed process. 

> If not, then have we considered changing the name of this API? IMHO it's surprising to give it the same name as an existing API without matching the signature of its return value.

Open to suggestions. I liked this name for its similarity actually. It is performing essentially the same function as its predecessor. I think any confusion would be pretty immediately resolved at dev-time.

index.src.html
chcunningham: We chatted on this, and my previous comment misunderstood your concern. 

> I'm not sure this is right, if we reach complete and and there are no more successive frames decoded and there's no error then by definition that frame is the 'complete frame'.

After chatting, we resolved to a different interpretation. For an output frame, being "complete" means the frame is the full fidelity output (i.e. the highest quality possible for formats that support progressive decoding). This is distinct from our concept of buffered data being "complete" (decoder.complete). 

Here are the scenarios we identified (and their outcomes) for when `decoder.complete = true`:
1. If encoded data is found to be malformed (e.g. the user gave an ArrayBuffer that cuts off in the middle of an encoded frame or the packetization of the encoded frames is mangled), trigger fatal shutdown of the decoder.
2. If we can decode a new successive frame (higher frame generation than previously output), resolve the promise with that frame. Set `result.complete` purely based on whether the output is the full quality output for a progressive image. 
3. if we can't decode a a new successive frame, run the "rejecet infeasible decode algorithm".

For scenario 1, I've added spec steps (not previously handled). I've also made also made the later "decoder produces an error" steps fatal. In short... if the data is identified to be malformed during demuxing or decoing, shut it all down.

index.src.html
chcunningham: I've had second thoughts. Purely a style thing, but in hindsight I like that the pending promises are all rejected with whatever the initial reason is with follow up operations being rejected because the codec is now in an invalid state (i.e. "closed")

index.src.html
chcunningham: We do want EncodingError for fatal error (now true in latest commit, "Fatally Reject Bad Data" algorithm). Note that the types of errors being discussed above are not fatal (merely infeasible requests).

chcunningham:
@padenot @aboba - this PR is over a month old. I've iterated on the design only very slightly since it's original posting (incorporating feedback from others). If no objection, I will go ahead and merge by EOD Wednesday 4/28.

index.src.html
padenot: > If the browser entirely relies on the OS to provide the codec, it may not be possible to know statically what codecs are supported (particularly for newer formats). Instead, we may be forced to query OS apis that are adjacent to the actual decoding APIs. Often this involves the same async IPC to a privaledged sandboxed process.

In my experience, this is mostly true for video, in the sense that it's well possible that the device that allows (say) power or cpu-efficient decoding is simply physically removed, and so that it's impossible to store the capabilities somewhere on the browser for synchronous access.

Do we have any evidence of a similar constraint for images? In our experience, hardware decoding for images has this problem where the setup time drawfs the decoding time, and the setup time need to happen per image (with a possible edge case when lots of images have the same dimensions/format maybe?).

dalecurtis: For HEIF I would expect implementors will require hardware support to enable the functionality. Early in process startup, this may not yet be known.

padenot:
We have quite a few feedback items already, and this is being circulated through various teams internally at Mozilla. Because of the type and number of comments we have, I think it's best to fix a few immediate things here (not a lot), merge, and then I'll open a series of issues on this repo, tagged with an appropriate tag (`image-decoding`?), so that people that mostly care about images can quickly follow only the issue about this. Would that be appropriate? I think it would make the discussions a bit more structured and workable.

That said, taking more than a month to review a fundamental new way to decode image on the web while the space hasn't really moved for years, while quite a few image formats are being added currently, all with the peculiarities and possible optimizations to make this a really compelling option for authors and unlocking new classes of apps is not exactly a long time.

dalecurtis:
Thanks Paul! I look forward to your reports. Your proposal sgtm.

While the PR may have only been out for a month, the API hasn't changed materially around the primary use case from the initial explainer and that's been circulating for over a year.

chcunningham:
Clicked by mistake!

chcunningham:
@padenot 

> I think it's best to fix a few immediate things here (not a lot)

To clarify, do I understand correctly that you will soon private a short list of things to consider for immediate fixes?

padenot:
> To clarify, do I understand correctly that you will soon private a short list of things to consider for immediate fixes?

I was going to add a regular review here like usual for a couple of things yes.

index.src.html
```diff
@@ -2260,14 +2290,805 @@
...
+ImageDecoder Interface {#imagedecoder-interface}
+------------------------------------------------
+
+<pre class='idl'>
+<xmp>
+[Exposed=(Window,DedicatedWorker)]
+interface ImageDecoder {
+  constructor(ImageDecoderInit init);
+
+  readonly attribute boolean complete;
```
padenot, line 2328: This needs an event associated with the value change. It was also mentioned a few minutes ago in the editor's call that this could be removed ?

index.src.html
```diff
@@ -2260,14 +2290,805 @@
...
+  readonly attribute ImageTrackList tracks;
```
padenot, line 2329: Is this a live object, or a static copy at time of retrieval? This needs to be specced.

index.src.html
padenot: Do you mean HEIC? HEIF is supported in software today by Chrome and Firefox.

dalecurtis: Yes, sorry. I keep forgetting that last letter :) In my head AVIF is what we call what Chrome and Firefox have.

index.src.html
dalecurtis: It could be, but it feels ugly if we do. Right now it's used to do this when ImageDecoder is used with a ReadableStream:

```
fetch('example_image').then(r => {
  let decoder =
      new ImageDecoder({data: r.body, type: r.headers.get("Content-Type")});

  let currentFrameIndex = 0;
  function decodeMoreFrames(result) {
    if (decoder.complete)
      return;
    decoder.decode({frameIndex: ++currentFrameIndex}).then(decodeMoreFrames);
  }

  decoder.decode({frameIndex: currentFrameIndex}).then(decodeMoreFrames);
});
```

The alternative is:
```
  function decodeMoreFrames(result) {
    decoder.decode({frameIndex: ++currentFrameIndex})
        .then(decodeMoreFrames)
        .catch(e => {
          if (e instanceof RangeError)
            return;
          throw e;
        });
  }
```

I'm okay with adding another promise-property or event, but we'd still need the boolean for synchronous tests. How does a promise-property like 'completed' sound? Too close to the boolean?

chcunningham: I think the decode-until-exception is workable, but unusual. I think a promise is better than an event here, per https://www.w3.org/2001/tag/doc/promises-guide#one-time-events. 

`complete` and `completed` are pretty close, but I don't hate it. I was tempted to name the promise completePromise, but I'm not finding [precedent for that style](https://source.chromium.org/search?q=f:idl$%20%22Promise;%22%20%20-f:test%20f:blink&sq=). Another option would be to name the boolen isComplete. Lots of [precedent for that](https://source.chromium.org/search?q=f:idl$%20%22boolean%20is%22%20-f:ppapi%20f:blink&sq=) . @padenot, do you like that idea?

index.src.html
chcunningham: It is a live object. Now made clear in the attribute definition.

index.src.html
padenot: I agree that using exceptions for control flow in perfectly normal (non-error) circumstances is not great, and is generally frowned upon in software engineering. A promise property sounds like a compelling option to me.

In general (and there are notable exceptions, with clear justifications), a property has to have something event-based to avoid any temptation to busy loop (even though it wouldn't do anything here).

I was thinking that maybe a generator / iterator pattern would lend itself well to this task, but the current style allows random access, and a generator / iterator wouldn't (afaik).

index.src.html
padenot: Yes, same, but our image folks made the distinction when I asked :-)

Per the TAG (a precedent with the autoplay policy, that took ages to resolve), a promise is to be used only if the information can change during the lifetime of the document. If not, then it's on the implementer to have the information ready. For example, in Gecko, this is part of the information received when a process is created.

If we anticipate that the support for this image format can be dynamic, it's going to be necessary (but annoying to have to do this async, because of an edge case that is not shipped anywhere and where it's not clear if it's going to be shipped anywhere). It would be a strange precedent however.

index.src.html
```diff
@@ -2343,14 +2374,805 @@
...
+  readonly attribute unsigned long length;
+  readonly attribute long selectedIndex;
+  readonly attribute ImageTrack? selectedTrack;
+};
+</xmp>
+</pre>
+
+### Internal Slots ### {#imagetracklist-internal-slots}
+: <dfn attribute for=ImageTrackList>[[ready promise]]</dfn>
+:: The promise used to signal when the {{ImageTrackList}} has been poplated
```
padenot, line 3044: `s/poplated/populated/`.

padenot:
I went ahead and created a tag called [`image`](https://github.com/w3c/webcodecs/labels/image) so that folks that mostly care (or are specialized) about images can follow easily new developments. I also filed an initial set of issues with this tag.

index.src.html
dalecurtis: At least in Chrome, in the event of a gpu process crash, it is possible support for some formats is no longer available. 

Otherwise I agree, we're just making this async for symmetry and a hypothetical. Maybe @jernoble or @aboba want to chime in to avoid a decision that would preclude any formats they might want to support.

index.src.html
dalecurtis: FWIW, &lt;img&gt; only has a complete property: https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/complete --- but the example on MDN would clearly benefit from a completed promise :)

dalecurtis: I went with completed, but can change later if there are objections.

index.src.html

index.src.html
chcunningham: Latest commit adds that promise. Leaving open for @padenot to resolve if this addresses his concern.

index.src.html
jrmuizel: @dalecurtis what formats aren't supported in Chrome when the GPU process crashes?

index.src.html
padenot: Yep!

index.src.html
jrmuizel: Ok, I think it's unlikely that a browser would choose to support an image format and not have a software fallback, but even if such a browser did exist the web author would still need to deal with decoding support going away after they've already checked for support with`isTypeSupported()`.

dalecurtis: Practically we don't have software fallback on Android for H.264 today -- so that at least is a real case. In the case where the codec goes away during usage the decoder would trigger a decoding error. The same would occur if the loss occurred between construction and the first decode call.

I think authors would have to handle this case regardless of if iTS is sync or not though. For security reasons (e.g., malicious invocation of the platform decoder), even if software fallback is available, it's unclear that automatic fallback is the right operation.

Ultimately my initial implementation was always synchronous, so if everyone wants this to be sync and isn't swayed by the hypotheticals I don't mind switching it back. I believe @chcunningham only suggested it for symmetry with the rest of the WebCodecs APIs.

chcunningham (commented):
@padenot @aboba - I've split off the one outstanding issue of async isTypeSupported() into #213 for followup. Should have better visibility/readability there. Going ahead with merge since this update is otherwise agreed on and we have other issues referencing it.

Issue #129, opened 2021-02-01
Clarify VideoFrame|AudioFrame clone/close/transfer behavior


chcunningham:
Current wording says this is a "deep copy", but makes a note that authors can implement with reference counting under the hood. This has [confused at least one user](https://github.com/WICG/web-codecs/issues/119#issuecomment-756998992) who expected that the "copy" was fine to hold onto. 

The text should formalize the reference counting behavior, as it is observable. Additionally, the text should use said behavior in describing new Serailzable and Transferrable characteristics of VideoFrame (what happens when you send to a worker).

chcunningham:
Title update to include AudioFrame. The two APIs should match. Working on a PR this week.

Issue #157, opened 2021-03-26
Simplifying and optimizing VideoFrame readback 


chcunningham:
(Full credit to @sandersdan for the following analysis and design. I'm just turning his doc into a bug.)

The `VideoFrame.planes` API allows reading back individual planes, but it has drawbacks:

* The implementation may map and unmap the underlying frame for each plane, which can be slow when all planes are read.
* It’s overly complicated to read all planes (partners have filed bugs showing its easy to get wrong)
* Compared to its complexity, it doesn’t actually offer much control.

This issue proposes a simpler API for copying all planes at once.

# Concepts #
## Region ##
A region is a `{top, left, width, height}` dictionary. `VideoFrame` should be updated to use this structure:

* Add `codedRegion`, which is always `{top: 0, left: 0, width: codedWidth, height: codedHeight}`.
  * We could deprecate `codedWidth` and `codedHeight`, but this only seems to complicate constructing `VideoFrames`.
* Add `visibleRegion`, replacing the `crop` parameters.

The default region is `visibleRegion`.

## elementSize ##
The number of bytes in a sample. Each `Plane` has an `elementSize` attribute.

## Stride ##
The number of bytes from the start of one row to the start of the next row, must be at least `elementSize * region.width`.

The default stride is exactly `elementSize * region.width`.

## Layout ##
A `layout` is a list of `{offset, stride}` dictionaries, one for each plane. `offset` is the location in a buffer that a plane starts.

The default `offset` is tightly packed (each plane immediately follows the previous), and either all offsets should be provided or none should be.

# API #
```
// Returns the size in bytes of a buffer that can hold a copy of |region|.
VideoFrame.allocationSize({region});
Plane.allocationSize({region, stride});

// Copies all planes into an output buffer |dst|.
VideoFrame.readInto(dst, {region, layout});  // returns |layout|
Plane.readInto(dst, {region, stride});       // returns |stride|

// Returns a new frame containing a copy of a region. Holding the new frame for an
// extended period of time will not stall decoding.
VideoFrame.copy({region});
```

## Changes to the planes API ##
* `Plane`
  * Add codedRegion, visibleRegion.
  * Add elementSize.
  * Rename columns to codedWidth and rows to codedHeight.
  * Deprecate length. Apps should call allocationSize().
  * Deprecate stride. It’s not relevant anymore.
* `PlaneInit`
  * Deprecate rows.
  * Add offset.

# Examples # 
[...]

PR #162, opened 2021-04-02
Rename AudioFrame->AudioData, drop AudioBuffer, add ref counting semantics.

### LINKED ISSUE #168: AudioFrame needs another name
An audio frame (or sample-frame, or sometimes just frame when talking about audio) is a term that is already in use with a specific meaning in general, on the web platform and even already in web codecs in other locations. It needs renaming.

It means "the ensemble of single audio samples that happen at a specific time, across multiple audio channels", which is unfortunately the opposite of its meaning in the Web Codecs spec, which is a group of continuous audio frames that has a duration. 

An explanation [here](https://developer.mozilla.org/en-US/docs/Web/API/Web_Audio_API/Basic_concepts_behind_Web_Audio_API#audio_buffers_frames_samples_and_channels), but this is a pretty generic term.

fwiw, Gecko uses the term "audio segment".

chcunningham:
Partially addresses #129. I'll send a follow up shortly to do the same for VideoFrame (will land at the same time, but trying to keep the diffs in my PRs small for easy review).

Fixes #168 (Rename AudioFrame -> AudioData).
Progresses #179 (Drop dependency on AudioBuffer).
Progresses #184 (Back attributes with slots).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/pull/162.html" title="Last updated on May 5, 2021, 3:53 AM UTC (5b33d7f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/162/209a2a0...5b33d7f.html" title="Last updated on May 5, 2021, 3:53 AM UTC (5b33d7f)">Diff</a>


chcunningham:
Some context for usage of "acquire the content" in https://github.com/WebAudio/web-audio-api-v2/issues/119#issuecomment-812238655

FYI @hoch @rtoy

index.src.html
```diff
@@ -1888,49 +1929,77 @@
...
+    2. Assign |init|.{{AudioFrameInit/buffer}} to {{AudioFrame/buffer}}.
+    3. Let |resource| be a new [=frame resource=].
+    4. [=Acquire the content=] [[WEBAUDIO]] of {{AudioFrame/buffer}} and
+        assign the acquired content to |resource|.
+
+        NOTE: The acquired content stored in |resource| is an immutable
+            snapshot {{AudioFrameInit/buffer}}'s content. Mutating
+            {{AudioFrame/buffer}} after construction (e.g. via
+            {{AudioBuffer/copyToChannel()}}) will have no effect on the
+            |resource|.
```
padenot, line 1954: I think you can more simply state that assigning `AudioFrameInit.buffer` to `AudioFrame.buffer` "acquires the content" of `buffer`, and that's it, not that you assign the "acquired content". But it might be more flexible to do this "acquire the content" operation only when calling `decode()` with a particular `AudioFrame`: only then it's actually sent to another thread and authors must be unable to tamper with it, etc. This would be consistent with what the Web Audio API does.

padenot (changes requested):
What we need again, but a few important questions.

index.src.html
```diff
@@ -1866,6 +1871,39 @@
...
+`[[resource reference]]` points to the same [=frame resource=] as the original
+frame.
+
+{{VideoFrame}}.{{VideoFrame/close()}} and {{AudioFrame}}.{{AudioFrame/close()}}
+will clear their [[resource reference]] slot, releasing the reference their
+[=frame resource=]
+
+A [=frame resource=] must outlive the frames that reference it. When no frames
+reference a given [=frame resource=], the resource must be destroyed
+(freeing associated [=system resources=]).
```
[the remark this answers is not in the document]
aboba: What does "A [=frame resource=] must outlive the frames that reference it" mean? Is this referring to the concept of reference frames?

index.src.html
```diff
@@ -1866,6 +1871,39 @@
...
+for reference counting (`frame.clone()` and `frame.close()`).
+
+### Reference Counting ### {#raw-media-memory-model-reference-counting}
+
+A <dfn>frame resource</dfn> is storage for the actual pixel data or the audio
+sample data described by a {{VideoFrame}} or {{AudioFrame}}.
+
+The {{AudioFrame}} {{AudioFrame/[[resource reference]]}} and {{VideoFrame}}
+{{VideoFrame/[[resource reference]]}} internal slots hold a reference to a
+[=frame resource=].
```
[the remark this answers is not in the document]
aboba: I had assumed it was `Resource*`, not  `Resource**`. But I think that is an implementation detail that is tangential to the main point.

index.src.html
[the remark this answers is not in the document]
chcunningham: I've moved (and edited) the final sentence about resource destruction into a non-normative note:. 

I left a version of the first sentence, stating that frames must not be destroyed while they are still referenced. I think this is important to mandate. If it is implied, it at least doesn't hurt to emphasize. I did massage the wording a bit to reduce potential confusion w/ "reference frames" as a media concept.

index.src.html
chcunningham: I'm of two minds on this one...

Mind A:
On the call we discussed whether this might move to encode(). I mentioned that this idea works but it breaks symmetry with other IO types (VideoFrame and Encoded*Chunk) which are immutable. This PR is otherwise making the two interfaces even more symmetrical. I think symmetry makes the API more intuitive.

Mind B:
Having an "immutable" AudioFrame who's buffer appears to be mutable (e.g. copyFromChannel()) is not very intuitive. We could instead "acquire the content" at `encode()` and otherwise let folks mutate their AudioBuffer all they like. If we pursue this, I'll want to remove `AudioFrame.clone()` entirely. The AudioFrame never acquired the content, so there isn't a "frame resource" to refcount. The content is owned by AudioBuffer. If folks want to take additional JS references on AudioBuffer, that is easy enough. We could continue to define `AudioFrame.close()`, which just nulls out the buffer giving us a chance for immediate cleanup rather than waiting for GC. Even that seems unimportant.

Having written that, I guess I'm leaning toward option 'B' as being less surprising and more flexible. @hoch @rtoy @dalecurtis @sandersdan - any concerns? 

If we take this route, I would drop this PR and move the entire frame resource / refcounting concept into the VideoFrame PR, making it VideoFrame-specific.

chcunningham (commented):
Thanks @padenot @aboba

index.src.html
dalecurtis: I'm wondering if we even want buffer as part of the AudioFrameInit structure. Since we allow VideoFrame(CanvasImageSource, VideoFrameInit) maybe it should be AudioFrame(buffer, optional AudioFrameInit).

If we want this to work similar to the VideoFrame mechanism than it is indeed taking a Snapshot in time of |buffer|, at the moment of creation buffer becomes immutable henceforth. I think the interface benefits from symmetry here.

index.src.html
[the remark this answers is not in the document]
aboba: Thanks. The potential confusion w/ "reference frames" has been removed.

index.src.html
chcunningham: > I'm wondering if we even want buffer as part of the AudioFrameInit structure. Since we allow VideoFrame(CanvasImageSource, VideoFrameInit) maybe it should be AudioFrame(buffer, optional AudioFrameInit).

I think VideoFrame is the weird one actually. The constructor shape made more sense when we were using the same init dicts between the two constructors, but now that we've split those I lean toward putting even more things into the distinct VideoFrameInit and VideoFramePlanesInit dicts, making both constructors take a single argument. But this is out of scope for this CL IMO.

> If we want this to work similar to the VideoFrame mechanism than it is indeed taking a Snapshot in time of |buffer|, at the moment of creation buffer becomes immutable henceforth. I think the interface benefits from symmetry here.

Ah, I forgot we were taking snapshots there too. This puts me somewhat at ease. With that + your edit, I'm back to favoring the proposal as written.

chcunningham:
@padenot @aboba I think we at least have agreement on the API shape and its intended meaning. Would you mind approving the merge here and we can keep pushing on wording/style stuff in follow up GH issues?

chcunningham:
> @padenot @aboba I think we at least have agreement on the API shape and its intended meaning. Would you mind approving the merge here and we can keep pushing on wording/style stuff in follow up GH issues?

@padenot @aboba  friendly ping

index.src.html
```diff
@@ -1810,6 +1815,42 @@
...
+{{AudioFrame}}.{{AudioFrame/clone()}} return new frame objects who's
+`[[resource reference]]` points to the same [=frame resource=] as the original
+frame.
+
+{{VideoFrame}}.{{VideoFrame/close()}} and {{AudioFrame}}.{{AudioFrame/close()}}
+will clear their [[resource reference]] slot, releasing the reference their
+[=frame resource=]
+
+A [=frame resource=] must remain alive as long as it continues to be referenced
+by a `[[resource reference]]`.
```
padenot, line 1848: "at least as", we don't want this to be overly prescriptive. For example, the prose above disallow delaying collection so that it's batched. It's kind of the opposite of what is in the note, which is what we want really.

index.src.html
chcunningham: Discussed on editors call. Resolved to go ahead w/ plans to break from AudioBuffer as described in https://github.com/w3c/webcodecs/issues/179

chcunningham:
Broadly: we are all in agreement about having clone()/close().

index.src.html

chcunningham:
@padenot @aboba I believe the latest commits address all outstanding comments.

index.src.html
```diff
@@ -1810,73 +1818,269 @@
...
+  readonly attribute AudioSampleFormat sampleFormat;
+  readonly attribute float sampleRate;
+  readonly attribute unsigned long numberOfFrames;
+  readonly attribute unsigned long numberOfChannels;
+  readonly attribute unsigned long allocationSize;
+  readonly attribute unsigned long long duration;
   readonly attribute unsigned long long timestamp;
-  readonly attribute AudioBuffer? buffer;
+
+  undefined copyFromChannel(BufferSource destination, unsigned long channelNumber);
```
padenot, line 1878: Something that occurred to me right after the call last Friday is that this API is not very good for interleaved data, it's really wasteful: each byte is touched twice to copy the full buffer.

index.src.html
```diff
@@ -1810,73 +1818,269 @@
...
+:: 32-bit float samples with interleaved channel arrangement.
+
+: <dfn enum-value for=AudioSampleFormat>S16P</dfn>
+:: 16-bit signed integer samples with planar channel arrangement.
+
+: <dfn enum-value for=AudioSampleFormat>S32P</dfn>
+:: 32-bit signed integer samples with planar channel arrangement.
+
+: <dfn enum-value for=AudioSampleFormat>FLTP</dfn>
+:: 32-bit float samples with planar channel arrangement.
```
padenot, line 2083: We'll need to explain what interleaved and planar are. Also how does those data types hold audio: for example, 16-bit is linear audio from in `[-65536, 65536]`, but the `FLT` variants only use [-1.0, 1.0] in general (but it can easily go outside this). And also `S32` is only expected to use 24-bits probably?

index.src.html
```diff
@@ -1810,73 +1818,269 @@
...
+different channels as either interleaved or planar.
 
-    NOTE: This section needs work. We should use the name and semantics of
-        VideoFrame destroy(). Similarly, we should add clone() to make a deep
-        copy.
-  </dd>
-</dl>
+<xmp class='idl'>
+enum AudioSampleFormat {
+  "U8",
```
padenot, line 2054: Do we want to continue RIFF WAVE's tradition of using u8 but signed for all other integer widths? What about endianess?

padenot (commented):
A couple of types, and some real questions, but generally I think this goes in the right direction.

index.src.html
dalecurtis: Yeah I would expect this to be something like copyFromPlane and interleaved data would only have one plane.

chcunningham (commented):
Addressing feedback.

index.src.html
padenot: This is probably fine for now. I was thinking that maybe we could expose single-byte samples as `i8` because it's more comfortable to do DSP type stuff when the bias is at 0, but it's quite established really.

index.src.html
```diff
@@ -1841,73 +1849,269 @@
...
+    destination buffer.
 
+    When invoked, run these steps:
+    1. If the value of |frame|'s {{AudioData/[[detached]]}} internal slot is
+        `true`, throw an {{InvalidStateError}} {{DOMException}}.
+    2. Let |allocationSize| be the number of bytes allocated to hold this
+        {{AudioData}}'s [=media resource=], as described by the getter steps of
+        {{AudioData/allocationSize}}.
+    3. If |allocationSize| is greater than `destination.byteLength`, throw a
+        {{TypeError}}.
```
padenot, line 2036: Isn't that usually `RangeError` [from ES](https://tc39.es/ecma262/#sec-native-error-types-used-in-this-standard-rangeerror) ? It would have been `IndexSizeError`, but that's deprecated.

index.src.html
```diff
@@ -1841,73 +1849,269 @@
...
+  readonly attribute AudioSampleFormat sampleFormat;
+  readonly attribute float sampleRate;
+  readonly attribute unsigned long numberOfFrames;
+  readonly attribute unsigned long numberOfChannels;
+  readonly attribute unsigned long allocationSize;
+  readonly attribute unsigned long long duration;
   readonly attribute unsigned long long timestamp;
-  readonly attribute AudioBuffer? buffer;
+
+  undefined copyTo([AllowShared] BufferSource destination, unsigned long planeNumber);
```
padenot, line 1909: I'm thinking maybe an offset and a frame count would make it a lot easier to send data to the WASM heap, and would help to address [this feedback](https://github.com/w3c/webcodecs/issues/176) from game developers (while not exactly that if I remember our conversation, it's really close).

After demuxing properly (with all the edge cases, easier said than done!), you have quite often a decoder delay value in frames, and a padding value in frames (is WAV the _only_ codec that doesn't have this? Certainly among the codecs available on the Web). Then imagine you're doing all your work in a custom audio engine written in C++, so the immediate thing you're doing is to copy the audio frames out to the WASM heap.

Say you're doing AAC, commonly it's 2112 frames of padding (not aligned to the common number of 1024 frames per packets), so you want to drop the two first packet entirely, and then copy to your heap with an offset of 64 frames starting on the third packet. Similarly, for the last packet, you only need to copy part of the packet, that's a function of a duration you find in the mp4 container usually (not always).

Here with the current API, you need to copy out to some intermediary buffer, and then copy from an offset to end final location, which is wasteful.

I believe adding two parameters

```
 undefined copyTo([AllowShared] BufferSource destination, unsigned long planeNumber, offset = 0, frameCount = 0);
```

with following meaning:
- `offset` is an offset in frames in the plane - throw `RangeError` if bigger than this `AudioData`'s `numberOfFrames`
- `frameCount` is the number of frames to copy - throw `RangeError` if bigger than `numberOfFrames - offset` or `destination.length`. The Web Audio API clamps in a similar situation, but I'd rather be explicit in this situation. The special value of 0 can be decided to mean `min(numberOfFrames - offset, destination.length)` if we want, that's ergonomic.

Most of the time you're not using those parameters though, hence the meaning of the defautl values.

dalecurtis: I'm not a fan of APIs which take an index like `planeNumber'. It sounds instead like we should have a planes array on the AudioData object, so folks can say data.planes[i].copyTo(dest, offset, frameCount). This would provide some symmetry with the VideoFrame API as well. This also makes it clear that you're not simply indexing into a channel array.

dalecurtis: I'm not a fan of APIs which take an index like `planeNumber'. It sounds instead like we should have a planes array on the AudioData object, so folks can say data.planes[i].copyTo(dest, offset, frameCount). This would provide some symmetry with the VideoFrame API as well. This also makes it clear that you're not simply indexing into a channel array.

chcunningham: @padenot , re offset + framecount + RangeErrors: all SGTM. Out of time tonight, but I'll send a new commit ASAP. I vote we create a `AudioDataReadToOptions` dict that lists these members alongside a required member for `planeNumber`. This also should help w/ @dalecurtis's concern that the simple argument looks like a channel index. 

Re: `planes` symmetry... not sure on that. We may yet delete `planes` from VideoFrame ([still up in the air](https://github.com/w3c/webcodecs/issues/157#issuecomment-832214060), and adding a new Plane interface with a single copyTo() method feels like a lot of ceremony IMO.

chcunningham:
This PR still needs some work to address the open issues, but I'd like to go ahead and merge because its blocking 2 other VideoFrame PRs that have approval and a 3rd that is under discussion. I commit to send a new PR to address @padenot's [last review](https://github.com/w3c/webcodecs/pull/162#pullrequestreview-651288952) tomorrow.

index.src.html
padenot: WFM, thanks!

padenot: https://github.com/w3c/webcodecs/issues/223 opened for followup, linking to the valuable informations in this thread.

Issue #166, opened 2021-04-05
Add definitions / diagrams for subtle video frame concepts


chcunningham:
Including coded size, crop size, display size, stride...

chcunningham:
Triage note: marking 'editorial', as there is broad consensus on what these terms / concepts mean, this issue merely tracks the need to document that better.

padenot:
I agree, from a high-level POV, but this needs to be precisely specified, because we all know small details matter when shipping APIs on the Web.

That said, it's not hard to spec I assume, granted we use the commonly accepted definitions, an attempt below:

The coded width is the size of a horizontal pixel line, that is bigger or equal to the stride. The coded height is the size of a vertical pixel line, that is bigger or equal to the height of the picture (but is that useful ?). The coded height x the coded width allows knowing how much memory an image occupies in practice. Then this memory interpreted by the metadata found on `VideoFrame`: a crop rectangle is applied (trimming off what's outside the cropped region), and then scaling is applied via `displayWidth` and `displayHeight`. Scaling is defined as being the linear transform we all know.

It can well be that the type of scaling needs to be specified, maybe by using https://html.spec.whatwg.org/#resizequality ?

chcunningham:
Thanks @padenot. I think those definitions match my expectation. @sandersdan  to double check.

One small nit 

> The coded width is the size of a horizontal pixel line, that is bigger or equal to the stride. 

I think this should be <= to the stride. 

Re: display scaling, I defer to @sandersdan.

sandersdan:
> The coded width is the size of a horizontal pixel line, that is bigger or equal to the stride.

There is no inherent relationship because `codedWidth` is measured in samples and `stride` is measured in bytes. `codedWidth` is also a measurement of the whole frame while `stride` is per-plane.

What is always true is that `planeCodedWidth * planeSampleBytes <= planeStride`.

There is still some discussion happening re: display size in #94. The current proposal is to treat the display size as a display aspect ratio, and then apply the usual 'scale up in exactly one dimension to match' approach that `<video>` uses. (Perhaps in the future we would add a `pixelWidth`/`pixelHeight` for cases that prefer using a pixel aspect ratio.)

padenot:
We can only work on this after https://github.com/w3c/webcodecs/issues/165 then. Thanks for the comment outlining my mistake. I'll work on a PR to add this precisely with all the comments taken into account.

Bman7714:
[...]

Issue #165, opened 2021-04-05
Add more complete descriptions for PixelFormats


chcunningham:
For instance, for I420 we could mention
- the number of planes is 3
- they are Y, U, and V
- their horizontal pixel sampling rate is 1, 2, and 2, respectively.
- their vertical pixel sampling rate is ""

This is good for interop and documentation. It also allows us to give more complete steps to validate PlaneInit dictionaries.

chcunningham:
Triage note: marking 'editorial' as we have broad consensus on the meaning of well known pixel formats (certainly true for the single I420 currently in the spec), and this issue merely tracks the work to document it's meaning.

padenot:
Yes, but this needs to be specified normatively and unambiguously before any implement ships. I _think_ we all agree what it is though:

`(w x h)` pixels of luma, and then immediately `(w x h / 4)` pixels for the U chroma plane and then immediately`(w x h / 4)` pixels for V Chroma plane. `w` and `h` are  the coded width and height, not necessarily the display width and height.

In this format, a component for a pixel is 8-bits in all 3 planes.

Because of the API, I don't think it's necessary to specify whether the Y, U and V planes linear and contiguous in memory when the buffer is in regular memory: it's always access via `readInto` (modulo names changes).

If an implementation wants to ship any other format it needs to be specified unambiguously in the same fashion.

padenot:
Also Y U and V are in this precise order.

chcunningham:
I believe we agree on the definition given above. @sandersdan to double check me.

sandersdan:
> Also Y U and V are in this precise order.

Just to clarify, that is how the APIs handle the planes but does not describe their layout in memory (which as you mentioned is arbitrary).

I'll add only that there are also general restrictions I would like to apply to all formats that also apply to I420:

- coded sizes must be a multiple of the largest subsampling factor: I420 coded sizes must be even.
- crop coordinates must be a multiple of the largest subsampling factor: I420 crops must have even origin and size values.

We could go further and specify byte ordering (eg. top-left at lowest memory address, increasing along rows up the the stride size, then wrapping to the left of the next row). Again this is probably the same for all formats, but it can cause confusion due to GL using y-flipped coordinates relative to this definition.

padenot:
Thanks for the comment. I agree with everything and will write a PR with all of this.

chcunningham:
@padenot good to close now that #288 is merged?

Issue #47, opened 2020-04-07
Negotiating colorspaces


sandersdan:
Ideally decoded content would be in the same colorspace as the encoded content, and colorspace negotiation is "just" a metadata management problem. Android MediaCodec works differently:

- MediaCodec requires us to supply HDR metadata up front, and provides output Textures. To support a decoder like this, we would need to specify colorspace information up-front, and that means it has to be a per-codec-profile description.
- Thus colorspace is a property of a codec as much as it is a property of a frame.
- A decoded frame could be in a different colorspace from the encoded media.
- There is no obvious way to handle transcoding generically.

Open questions:

- Is there a workaround for MediaCodec? If so, can we assume that all future decoders would also allow unprocessed access to the decoded frames?
- Should we be implementing colorspace conversions for encoding, or is this always something the app should do? (Related: should encoders handle scaling content or is this also an app concern?)

chcunningham:
> MediaCodec requires us to supply HDR metadata up front, and provides output Textures. 

[Reading here](https://developer.android.com/reference/android/media/MediaCodec#raw-video-buffers), it sounds as if YUV is still available for Texture (Surface) output. Perhaps this would allow us to forgo providing the HDR metadata?

sandersdan:
Summary of discussion today: HDR on Android is an unsolved problem; only tunneled mode is specified, and we can't use tunneled mode for WebCodecs. We do not know what capabilities may be exposed in the future.

As a user of a decoder I imagine I would be frustrated if I could not get unmodified output. We could make the behavior opt-in, then apps that need unmodified output would fail to create a codec.

padenot:
Having access to the raw image and the associated color-space is much more powerful than any alternative. Can we have more info about why Android does it the way it does it? It sounds weird from a first look.

Sorry if I missed discussions and this, I've had to skip a couple meetings.

sandersdan:
I believe Android did this as a shortcut and simplification; the entire decode-tonemap-render path is merged into one unit that can be tuned for a specific device, and existing media paths don't need to understand HDR.

I don't know whether the tonemapping part is actually implemented in an inseparable way on real devices, but certainly I imagine there will be a time in the future where the decoding part is always a separable component. Hopefully the API to use such a decoder will be clearly defined by then.

chcunningham:
triage notes: marking 'extension' as the expected outcome is to simply add members to config dicts and attributes to VideoFrame describing color space.

padenot:
[...]

PR #167, opened 2021-04-05
Add refcounting semantics to VideoFrame

### LINKED ISSUE #129: Clarify VideoFrame|AudioFrame clone/close/transfer behavior

chcunningham:
Making it symmetric w/ AudioFrame (PR #162)

Adds clone/close
Adds [[resource reference]]
Updates constructors accordingly (and fix cruft/obsolete steps).
Removes destroy (replaced by close).
    
Fixes #129
Notes issues #165 and #166 for follow up.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/pull/167.html" title="Last updated on May 5, 2021, 3:56 AM UTC (4829102)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/167/5b33d7f...4829102.html" title="Last updated on May 5, 2021, 3:56 AM UTC (4829102)">Diff</a>


aboba (approved):

index.src.html
```diff
@@ -2208,90 +2246,115 @@
...
+        11. Assign |frame|.{{VideoFrame/codedHeight}} to
+            {{VideoFrame/codedHeight}}.
+        12. Assign |frame|.{{VideoFrame/displayWidth}} to
+            {{VideoFrame/displayWidth}}.
+        13. Assign |frame|.{{VideoFrame/displayHeight}} to
+            {{VideoFrame/displayHeight}}.
+        14. Assign |frame|.{{VideoFrame/duration}} to {{VideoFrame/duration}}.
+        15. Assign |frame|.{{VideoFrame/timestamp}} to
+            {{VideoFrame/timestamp}}.
+    2. Return |clone|.
```
padenot, line 2345: This style is overly verbose. It would be clearer to say to copy all attributes to the new `clone` except `x` and `y` and then spell out the steps for those.

index.src.html
```diff
@@ -2208,90 +2246,115 @@
...
+        {{ImageBitmap/resizeHeight}} is present and is 0, then return |p|
+        rejected with an {{InvalidStateError}} {{DOMException}}.
+    3. If the <a>this'</a> {{VideoFrame/[[detached]]}} internal slot is set to
+        `true`, then return |p| rejected with an {{InvalidStateError}}
+        {{DOMException}}.
+    4. Let |imageBitmap| be a new {{ImageBitmap}} object.
+    5. Set |imageBitmap|'s bitmap data to a copy of the {{VideoFrame}} pixel
+        data, at the frame's intrinsic width and intrinsic height (`i.e`., after
+        any aspect-ratio correction has been applied), cropped to the source
+        rectangle with formatting.
```
padenot, line 2290: This will need to be spelled out at some point.

index.src.html
```diff
@@ -2208,90 +2246,115 @@
...
+    6. If the origin of |imageBitmap|'s image is not same origin with entry
+        settings object's origin, then set the origin-clean flag of
+        |imageBitmap|'s bitmap to `false`.
+    7. Run this step in parallel:
+      1. Resolve p with imageBitmap.
```
padenot, line 2295: I don't understand this last bit. What is the intent?

index.src.html
```diff
@@ -2049,111 +2067,131 @@
...
-        `frame.codedHeight`.
-    7. If `frameInit.duration` is present, initialize it to `frame.duration`.
-        Otherwise, default `frame.duration` to null.
-    8. If `frameInit.timestamp` is present, initialize it to `frame.timestamp`.
-        Otherwise default `frame.timestamp` to null.
+1. If the value of |imageBitmap|'s' {{PlatformObject/[[Detached]]}} internal
+    slot is `true`, then throw an {{InvalidStateError}}
+    {{DOMException}}.
+2. Let |resource| be the [=frame resource=] containing the pixel data for
+    |imageBitmap|.
```
padenot, line 2086: This is referred to as `bitmap data` in HTML.

index.src.html
```diff
@@ -2049,111 +2067,131 @@
...
+3. Let |resourceReference| be a reference to |resource|.
```
padenot, line 2087: How does it work if this data is then later made unavailable, for example because `transferFromImageBitmap` has been called ?

index.src.html
```diff
@@ -2049,111 +2067,131 @@
...
+    12. Assign |init|.{{VideoFramePlanesInit/cropHeight}} to
+        {{VideoFrame/cropHeight}}.
+    13. Assign |init|.{{VideoFramePlanesInit/displayWidth}} to
+        {{VideoFrame/displayWidth}}.
+    14. Assign |init|.{{VideoFramePlanesInit/displayHeight}} to
+        {{VideoFrame/displayHeight}}.
+    15. Assign |init|.{{VideoFramePlanesInit/timestamp}} to
+        {{VideoFrame/timestamp}}.
+    16. Assign |init|.{{VideoFramePlanesInit/duration}} to
+        {{VideoFrame/duration}}.
```
padenot, line 2192: There an opportunity to clarify here, either by sharing steps or using English instead of repetitive steps.

padenot (changes requested):
This needs clarification about ownership, and other fixes, but in general is what we need.

index.src.html
chcunningham: I condensed a few steps, but I also realized I was missing some defaulting behavior when things weren't provided. After fixing that, the opportunity to condense mostly disappeared (most steps have specific default logic that doesn't generalize).

High level: coded size is required (now by IDL), which cascades to crop size if not provided, and crop size cascades to display size if not provided.

index.src.html
chcunningham: Done. https://html.spec.whatwg.org/multipage/imagebitmap-and-animations.html#cropped-to-the-source-rectangle-with-formatting

index.src.html
chcunningham: I think the idea is just to clarify that we aren't resolving synchronously. This wording matches https://html.spec.whatwg.org/multipage/imagebitmap-and-animations.html#dom-createimagebitmap

index.src.html
```diff
@@ -2340,7 +2401,11 @@
     {{InvalidStateError}}.
 2. If {{Plane/length}} is greater than |`dst.byteLength`|, throw a
     {{TypeError}}.
-3. Copy the {{Plane/[[plane buffer]]}} into <var ignore>dst</var>.
+3. Let |resource| be the [=frame resource=] referneced by
```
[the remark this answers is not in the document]

index.src.html
chcunningham: Done.

chcunningham (commented):
Thanks @padenot and @aboba ! Nearly everything addressed.

index.src.html
chcunningham: @sandersdan friendly nudge to check me on the above.

chcunningham:
@padenot , similar to the audio issue, I think we at least have agreement on the API shape and its intended meaning. Would you mind approving the merge here and doing follow up issues for any lingering wording/style concern?

chcunningham:
> @padenot , similar to the audio issue, I think we at least have agreement on the API shape and its intended meaning. Would you mind approving the merge here and doing follow up issues for any lingering wording/style concern?

@padenot friendly ping.

index.src.html
sandersdan: I am not very familiar with the spec details at the level of resource references. `ImageBitmaps` are immutable, so it's always a valid to implement "copy" by reference, and this is compatible with `transferFromImageBitmap()` as long as exclusive access isn't needed (it's never needed, at least in Chrome's implementation).

It's probably fine to just say that we copy the bitmap data, the same as `createImageBitmap(ImageBitmap)` does.

index.src.html
```diff
@@ -2343,7 +2399,11 @@
     {{InvalidStateError}}.
 2. If {{Plane/length}} is greater than |`dst.byteLength`|, throw a
     {{TypeError}}.
-3. Copy the {{Plane/[[plane buffer]]}} into <var ignore>dst</var>.
+3. Let |resource| be the [=frame resource=] refrenced by
+    {{Plane/[[parent frame]]}}'s {{VideoFrame/[[resource reference]]}}.
+4. Let |plane bytes| be the region of bytes in [=frame resource=] coresponding
+    to this plane.
+5. Copy the |plane bytes| into <var ignore>dst</var>.
```
padenot, line 2406: You should be able to drop the `ignore` if you make `dst` a `<var>` above, no?

index.src.html
```diff
@@ -2211,90 +2260,99 @@
...
-    object's origin, then set the origin-clean flag of |imageBitmap|'s bitmap to
-    `false`.
-7. Run this step in parallel:
-  1. Resolve p with imageBitmap.
+    When invoked, run these steps:
+    1. Let |p| be a new Promise.
+    2. If either |options|'s {{ImageBitmapOptions/resizeWidth}} or
+        {{ImageBitmap/resizeHeight}} is present and is 0, then return |p|
+        rejected with an {{InvalidStateError}} {{DOMException}}.
+    3. If the <a>this'</a> {{VideoFrame/[[detached]]}} internal slot is set to
```
padenot, line 2297: nit: `is true`

index.src.html
```diff
@@ -2211,90 +2260,99 @@
...
+        `true`, then return |p| rejected with an {{InvalidStateError}}
+        {{DOMException}}.
+    4. Let |imageBitmap| be a new {{ImageBitmap}} object.
+    5. Set |imageBitmap|'s bitmap data to a copy of the {{VideoFrame}} pixel
+        data, at the frame's intrinsic width and intrinsic height (`i.e`.,
+        after any aspect-ratio correction has been applied),
+        [=ImageBitmap/cropped to the source rectangle with formatting=].
```
padenot, line 2304: What about all the other options? In particular, we really want to make sure colors are correct here, we can't just copy the bytes I think.

index.src.html
```diff
@@ -2052,111 +2072,140 @@
...
+        {{VideoFrame/cropTop}} respectively.
+    10. If |init|.{{VideoFramePlaneInit/cropWidth}} [=map/exists=], assign it
+        to {{VideoFrame/cropWidth}}. Otherwise, assign
+        {{VideoFrame/codedWidth}} to {{VideoFrame/cropWidth}}.
+    11. If |init|.{{VideoFramePlaneInit/cropHeight}} [=map/exists=], assign it
+        to {{VideoFrame/cropHeight}}. Otehrwise, assign
+        {{VideoFrame/codedHeight}} to {{VideoFrame/cropHeight}}.
+    12. If |init|.{{VideoFramePlaneInit/displayWidth}} [=map/exists=], assign
+        it to {{VideoFrame/displayWidth}}. Otherwise, assign
+        {{VideoFrame/cropWidth}} to {{VideoFrame/displayWidth}}.
```
padenot, line 2200: This also could benefit from factoring, it's a bit laborious to review in this form.

index.src.html
padenot: I've just re-read a bunch of stuff about this, and I agree with the last sentence in Dan's message above.

index.src.html
```diff
@@ -2052,111 +2072,140 @@
...
-    translation. When this occurs, we should allow planes to be null and format
-    to be "opaque" to avoid early optimization. We should make conversion
-    explicit and user controlled by offering a `videoFrame.convertTo(format)`
-    that returns a Promise containing a new VideoFrame for which the
-    copies/translations are performed.
-
-<dfn constructor for=VideoFrame title="VideoFrame(imageBitmap, frameInit)">
-  VideoFrame(imageBitmap, frameInit)
+<dfn constructor for=VideoFrame title="VideoFrame(imageBitmap, init)">
+  VideoFrame(imageBitmap, init)
```
padenot, line 2085: We really need to do something about color spaces here. What's the best issue to track this?

padenot (changes requested):

index.src.html
chcunningham: We're all in agreement. Our understanding is that this doesn't require any change to the language / interfaces (implied behavior).

index.src.html
chcunningham: I agree, it's verbose. Let's punt on refactoring until a later PR (CIS PR modifies this heavily).

index.src.html
chcunningham: We're definitely missing proper handling of color. Discussed in https://github.com/w3c/webcodecs/issues/47

index.src.html
```diff
@@ -2052,111 +2072,140 @@
...
 
-NOTE: this section needs work. Current wording assumes a VideoFrame can always
-    be easily represented using one of the known pixel formats. In practice, the
-    underlying UA resources may be GPU backed or formatted in such a way that
-    conversion to an allowed pixel format requires expensive copies and
-    translation. When this occurs, we should allow planes to be null and format
-    to be "opaque" to avoid early optimization. We should make conversion
-    explicit and user controlled by offering a `videoFrame.convertTo(format)`
-    that returns a Promise containing a new VideoFrame for which the
-    copies/translations are performed.
```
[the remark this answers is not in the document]
chcunningham: Yes, intentionally removed. The original issue was that I assumed you could always pick a pixel format from our enum. This PR adds a failure mode to set format = "" when no suitable format can be found.

index.src.html
chcunningham: Will fix

chcunningham (commented):
Thanks @padenot - I'll send a fix for the nit. Otherwise, we've agreed to merge and handle remainder in follow ups.

index.src.html
chcunningham: Nice. Done.

index.src.html

index.src.html
chcunningham: Discussed in editors call:  I borrowed the language here from createImageBitmap(). It seems the options are applied when the ImageBitamp is created, but baked into the the bitmap data from that point forward. This particular step in our algorithm is still useful, as it describes that we are cropping the full image (from (0,0) -> (width, height).

chcunningham (commented):
Thanks, all comments addressed

padenot (approved):
Good for merging now, thanks!

chcunningham:
This one depends on #162, so I'll merge it as soon as that one goes in.

chcunningham:
I'll merge back to main manually.

Issue #127, opened 2021-01-29
Should the spec require copying init.data when constructing EncodedAudioChunk and EncodedVideoChunk?


OllieJones:
Sections [9.1.1](https://wicg.github.io/web-codecs/#encodedaudiochunk-constructors) and [9.2.1](https://wicg.github.io/web-codecs/#encodedvideochunk-constructors) of the draft WebCodecs spec say:

> 2. Assign a copy of init.data to chunk.data.

I suggest the words **a copy of** be removed from the spec. It seems like leaving those words in requires unnecessary copying of compressed bitstream data.  

Leaving these words out would allow a client of these constructors to make the decision about copying that data. It's possible for a client to pass a `.subarray()` of some ArrayBuffer. If that ArrayBuffer is volatile for some reason, it should be up to the client to copy the data. 

Of course, if a decoder implementation requires the compressed bitstream data to be copied, the implementation can do so.  But promising to do so adds a required copying step.

How about this wording?

> 2. Assign init.data to chunk.data. The client must not allow init.data to change after using this ctor.

sandersdan:
The goal here is to reduce the total number of copies. We can't just use an ArrayBuffer directly as it could be mutated, and if a decoder is currently reading from it that could result in TOCTTOU-style issues.

@chcunningham previously we discussed removing `data` and providing `copyInto()` instead, thus making encoded chunks immutable.

An alternative to immutable chunks is to have our decoders always copy the data before decoding it, but I don't really see an advantage in having mutable chunks.

chcunningham:
> We can't just use an ArrayBuffer directly as it could be mutated, and if a decoder is currently reading from it that could result in TOCTTOU issues.

One option under consideration is to transfer the provided ArrayBuffer when constructing the encoded chunk. This would mitigate TOCTTOU issues by detaching the users reference to the data.

> @chcunningham previously we discussed removing data and providing copyInto() instead, thus making encoded chunks immutable.

I think this is probably the best we can do without read-only ArrayBuffers (https://github.com/WICG/reducing-memory-copies/issues/1). It does impose a new copy for users of the encoder interface (e.g. presently they could just pass our data directly to WebTransport). 

> An alternative to immutable chunks is to have our decoders always copy the data before decoding it, but I don't really see an advantage in having mutable chunks.

I lean toward immutable. The alternative doesn't resolve the issue #80.

chcunningham:
I'm closing this as I think its work is tracked in duplicate issues. The plan is:
[...]

PR #174, opened 2021-04-09
Make Encoded*Chunk interfaces immutable.

### LINKED ISSUE #80: EncodedChunks confusingly provide access to constructed ".data" -- despite the decoder making an internal copy.
As folks are starting to play with the API, at least one person has hit a footgun where they modified the encoded chunk after sending it to the decoder. We should just copy before control returns to JS to avoid this.

Obviously this prevents a zero copy on encoded data path, but generally copying the input data isn't expensive.

chcunningham:
Provides a readInto() method for copying the data out.

Fixes #80. Also, see #127.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/pull/174.html" title="Last updated on May 13, 2021, 10:53 PM UTC (877b646)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/174/731a471...877b646.html" title="Last updated on May 13, 2021, 10:53 PM UTC (877b646)">Diff</a>


chcunningham:
FYI @sandersdan

aboba (commented):
Line 1742: s/repsenting/representing/

chcunningham:
> Line 1742: s/repsenting/representing/

Done! Thanks.

aboba (approved):

index.src.html
```diff
@@ -1720,7 +1720,9 @@
   constructor(EncodedAudioChunkInit init);
   readonly attribute EncodedAudioChunkType type;
   readonly attribute unsigned long long timestamp;  // microseconds
-  readonly attribute ArrayBuffer data;
+  readonly attribute unsigned long byteLength;
+
+  undefined readInto(ArrayBufferView dst);
```
padenot, line 1725: `copyTo` is easier to understand because it says what it does. The naming is also in line with what the Web Audio API has, and someone close to `TypedArray.copyWithin`.

index.src.html
```diff
@@ -1766,7 +1780,9 @@
   readonly attribute EncodedVideoChunkType type;
   readonly attribute unsigned long long timestamp;  // microseconds
   readonly attribute unsigned long long? duration;  // microseconds
-  readonly attribute ArrayBuffer data;
+  readonly attribute unsigned long byteLength;
+
+  undefined readInto(ArrayBufferView dst);
```
padenot, line 1785: Same.

padenot (commented):
Looks good, but I'm not sure about the naming. Did you name it this way because of some precedent in some other API ?

index.src.html
chcunningham: @sandersdan - lets chat

sandersdan: `readInto()` is what I've proposed for `VideoFrame` readback. There the name expresses two things: it's not necessarily just a copy, and the destination is an app-provided buffer rather than a new one.

This use in chunks only has the second property.

In use:

```js
let layout = await frame.readInto(buf);
chunk.readInto(buf);

let layout = await frame.copyTo(buf);
chunk.copyTo(buf);
```

In context prefer `readInto()` for frame readback, but `copyTo()` for chunks (and I don't like `readTo()` at all). I'm not sure what the most consistent set of choices is.

chcunningham: > it's not necessarily just a copy

IIUC, the point you're making here is in cases like texture backed frames where read is doing things like unswizzling etc... 

I follow that this is not just a copy. Is this an important signal to the JS user? If so, I guess we split the baby, keeping readInto() for VideoFrame, but copyTo() for chunks. Otherwise, we can make them both copyTo()... but read on. 

> WebAudio has copyToBuffer()

This was wrong. WebAudio [actually has](https://webaudio.github.io/web-audio-api/#audiobuffer) 

```
interface AudioBuffer {
  ...
  undefined copyFromChannel (Float32Array destination,
                             unsigned long channelNumber,
                             optional unsigned long bufferOffset = 0);
  undefined copyToChannel (Float32Array source,
                           unsigned long channelNumber,
                           optional unsigned long bufferOffset = 0);
};
```

So there, copyTo*() is actually a mutation API. It's copyFrom*() that is the "read". But we definitely don't want copyFrom(buffer) here... as that is even more hinting at the wrong direction. WebAudio only gets by because its copyFromChannel(). And I just copied that name in PR #162.

Best I've come up with so far: copyOut(buffer). "Out" is pretty unambiguous about the direction. We can have VideoFrame name it readOut(buffer) if that distinction matters. 

Aside: in other discussion, we've floated the idea of adding a copy() API to VideoFrame that makes a deep copy. We'd have to name this something else if we change readInto() -> to be some sort of copy*().

chcunningham: Internally, my idea of `copyOut()` didn't get much love, while `copyTo()` and `copyInto()` were both deemed acceptable. 

On this thread, `copyTo()` has been preferred to `copyInto()`. I'm supportive of that. Note that it is backward from WebAudio (copyToChannel() is a mutator) , but I think WebAudio might be in the wrong. It's good for the name to be intuitive as you write/read the code from left to right. That is `chunk.copyTo(buffer)` reads as "chunk: copy your contents to buffer". This is consistent with Streams (pipeTo()) and TextEncoder (encodeInto()). 

So I'm joining the `copyTo()` camp. Any objections?

Re: `readInto()`, we would make this simply `readTo()` or `copyTo()`. I lean `copyTo()` here as well. I take @sandersdan's point about `readInto()` suggesting that it's more than just a copy (i.e. it may be an expensive async readback, involving unswizzling etc)... but I don't think "read" really hints at that very strongly ([making frame.readInto() async](https://github.com/w3c/webcodecs/pull/208#discussion_r625508233) is probably the stronger hint). @sandersdan any objection?

sandersdan: I interpret "read" as potentially expensive since the term for copying an image from GPU memory is "readback", and also by association with WebGL `readPixels()` (which presumably got its name for the previous reason).

No objections to any of these name, though.

chcunningham: Understood. 

In summary
1. we're aligned on copyTo() for all non-VideoFrame things (including chunks)
2. we're ambivalent about copyTo() vs readTo() for VideoFrame. 

Re point 2, @padenot would you like to be the tie breaker?

padenot: I like @sandersdan's argument of using `read` because of `readback` and `glRead{,n}Pixels` (not only in WebGL, also in regular GL).

I also still like 1. 

But this discussion is about `Encoded*Chunk` so this discussion doesn't need to block merging this. It's not the same operation at all.

PR #186, opened 2021-04-21
Minor changes to width/height config members.

### LINKED ISSUE #94: It's hard to set size(width/height) information in `VideoDecoderConfig`
(I'm sorry if this issue is duplicated one.)
According to Spec, `VideoDecoder` requires size(codec and crop) information to configure decoder.
But I think that it's hard to set without analyzing raw stream when I received only raw stream.

Like H.26x codec includes coded size information in SPS as RBSP. So decoder may do without passing size information.

`VideoFrame` already has information to render appropriately. However I would be happy if there is some notification callback to get size information before decoding a first frame.

chcunningham:
For VideoDecoderConfig, the coded size attributes are no longer
required. They only serve as a hint for the UA to create an approriate
underlying decoder (e.g. hardware accelerated for higher resolutions)
and are not strictly required. Fixes #94.

For VideoEncoderConfig, cropWidth and cropHeight are renamed to simply
be width and height. The crop prefix was added in #113 as I realized
that we were describing the "crop" size of the encoded VideoFrames. But
I now think this risks giving the impression that the encoder is going
to crop frames. Being intuitive is more important than being pedantic.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/pull/186.html" title="Last updated on Apr 21, 2021, 5:55 AM UTC (059860b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/186/be1f1cc...059860b.html" title="Last updated on Apr 21, 2021, 5:55 AM UTC (059860b)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -1356,8 +1356,8 @@ VideoDecoderConfig{#video-decoder-config}
 dictionary VideoDecoderConfig {
   required DOMString codec;
   BufferSource description;
-  required unsigned long codedWidth;
-  required unsigned long codedHeight;
+  unsigned long codedWidth;
+  unsigned long codedHeight;
   unsigned long cropLeft;
   unsigned long cropTop;
   unsigned long cropWidth;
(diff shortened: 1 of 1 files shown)


aboba (approved):

chcunningham:
@padenot I'll go ahead and land this very minor edit w/ just Bernard's approval. Happy to address any feedback you may have in a follow up.

PR #193, opened 2021-04-26
Switch to w3c/spec-prod and automate publication to /TR

tidoust:
The code currently relies on Travis CI to update the gh-pages branch. This update replaces that with the w3c/spec-prod action and enables automatic publication of the specs as working drafts to /TR whenever a change gets pushed to the main branch.

The update also switches to the new NOTE-ED and NOTE-WD statuses in Bikeshed, which makes it possible to drop the "Text macro" workaround, see discussion in:
https://github.com/tabatkins/bikeshed/pull/1999#issuecomment-819321482


chcunningham (approved):
Looks great, thanks @tidoust!

tidoust:
Grmpf, the job actually fails to even start because it considers that the reference to the `secrets` context in the `matrix` is invalid. Not sure how to fix this yet, see https://github.com/w3c/spec-prod/issues/55#issuecomment-827373030

PR #196, opened 2021-04-27
[spec-prod] Dereference secret name in step

tidoust:
The `secrets` context is not available at the `matrix` level. See discussion in:
https://github.com/w3c/spec-prod/issues/55#issuecomment-827576189

Workaround is to list the secret name at that level and to dereference it in the actual step.

### DIFF
--- a/.github/workflows/pr-push.yml
+++ b/.github/workflows/pr-push.yml
@@ -23,17 +23,17 @@ jobs:
         include:
           - source: index.src.html
             destination: index.html
-            echidna_token: ${{ secrets.ECHIDNA_TOKEN_WEBCODECS }}
+            echidna_token: ECHIDNA_TOKEN_WEBCODECS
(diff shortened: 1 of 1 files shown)


tidoust:
@chcunningham This should fix the job. I note that the use of a `matrix` in the YML file effectively means that the multiple instances of the job (one for each entry in the matrix) run each time. Hence the "3 successful checks" here.

chcunningham (approved):
Looks good to me. Thanks for the quick fix!

tidoust:
Success... almost ;)

Publication to `gh-branch` succeeded for the 3 documents. Publication of WebCodecs to /TR succeeded as well although job still reported a failure for some reason. Publication of the registry and AVC documents to /TR failed due to a shortname mismatch. I'll look into it.

chcunningham:
Thanks!

PR #208, opened 2021-04-30
Simplify VideoFrame Readback

chcunningham:
Define new concepts of Regions and Layout as described in #157. 
* Regions make it easier to articulate what part of the frame is desired for reading out (e.g. only the "visible" (non padded) region).
* Layouts make it easier to describe the position and padding of a plane within a buffer. Useful during Plane construction as well as when reading out Planes to a BufferSource.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/pull/208.html" title="Last updated on Jun 11, 2021, 10:32 PM UTC (9591101)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/208/942a4aa...9591101.html" title="Last updated on Jun 11, 2021, 10:32 PM UTC (9591101)">Diff</a>


chcunningham:
@padenot @aboba This makes the interface changes but does not yet include spec steps. Uploaded as-is for discussion during (or prior) to tomorrow's meeting.

chcunningham:
Editors call: 
@padenot generally looks good, going to loop in graphics folks shortly
@bernard, no objection

padenot:
>  @padenot generally looks good, going to loop in graphics folks shortly

(I've contacted them).

index.src.html
```diff
@@ -2523,9 +2513,9 @@
 };
```
chcunningham, line 2513: We propose deleting `plane.readInto()`, as it is now replaced by `frame.readInto()`. 

The original `frame.planes[i].readInto()` was much less efficient in the common case of needing to read all planes (for mapped YUV frames, this required us to map/unmap 3 times!) and much more verbose to do common things like read out just the visible region (no padding). 

The only missing feature of the new path is the ability to read back just one plane if you desire that (e.g. say you just want Y of YUV to run some ML analysis like face tracking). This can be easily addressed by adding a `discard` field to the layout dict...so the YUV case becomes:

```
let options = {layout: [{},                // y plane, default layout
                        {discard: true},   // u plane, discarded
                        {discard: true}]}; // v plane, discarded
frame.readInto(buff, options);
```

chcunningham (commented):
I added some additional comments for design discussion.

index.src.html
chcunningham: Furhtermore, with `planes.readInto()` deleted, we propose deleting `frame.planes` entirely. It's remaining attributes only served as metadata for sizing a BufferSource to provide to `plane.readInto()`. This metadata is no longer needed with the simpler regions-based `frame.readInto()`. Folks can instead determine the minimum size for their destination buffer using using `frame.allocationSize()`.

index.src.html
```diff
@@ -2110,15 +2083,29 @@
...
+  readonly attribute VideoFrameRegion visibleRegion;
+
   readonly attribute unsigned long displayWidth;
   readonly attribute unsigned long displayHeight;
   readonly attribute unsigned long long? duration;
   readonly attribute unsigned long long? timestamp;
 
+  // Returns the minimum size (in bytes) of a buffer that readInto() could
+  // read this frame into with the given options.
+  unsigned long allocationSize(
```
chcunningham, line 2100: @sandersdan I think we probably out to go with byteLength. Either name works, but TypedArray uses byteLength.

sandersdan: To me that's a bit strange, I see`byteLength` as an intrinsic property of a `TypedArray` while for `VideoFrame` it's a configurable detail of `readInto()`.

Ideally a name would convey the relationship between the methods but I don't have an idea that does that.

Perhaps `byteLength()` being a method is enough to make the meaning clear.

index.src.html
chcunningham: Sorry, lets discuss this in the issue #157 for better visibility.

chcunningham (commented):
> I added some additional comments for design discussion.

On second thought, lets move the design discussion to #157 for better visibility/readability. Please see [this recent comment](https://github.com/w3c/webcodecs/issues/157#issuecomment-832214060).

>> @padenot generally looks good, going to loop in graphics folks shortly
>
> (I've contacted them).

Thanks. Standing by. Please give this issue priority. Of all our v1 launch blockers, this one has the most open questions.

padenot:
> Thanks. Standing by. Please give this issue priority. Of all our v1 launch blockers, this one has the most open questions.

I'm relaying the comments I've received, paraphrasing to add a bit more context:

> The changes look like a good idea overall. A few thoughts:
> 
> Most graphics APIs use `Region` to mean a complex shape (like a set of rectangles, or polygons). Can we just use the more-common `Rect` for top/left/width/height? Could we just use the existing `DOMRect` instead of defining a new dictionary type?
> 
> `PlaneInit` now takes a `BufferSource` and an offset, what happens if someone constructs a `VideoFrame` with a set of planes that each refer to different buffers? The idea of having a per-plane offset suggests that we expect all planes to be within a single buffer, so I think `PlaneInit` should be a single `BufferSource` plus a `Sequence` of `PlaneLayouts`. 
> 
> Alternatively we should truly allow for separate buffers per-plane, which in theory should be possible (and describable by our internal C++ descriptors in Gecko's implementation), but doesn't seem to happen in practice (from the existing decoders in Gecko).
> 
> As is being discussed, adding an option to do an async readback seems like a really good idea.

chcunningham:
Thanks @padenot 

> Most graphics APIs use Region to mean a complex shape (like a set of rectangles, or polygons). Can we just use the more-common Rect for top/left/width/height? Could we just use the existing DOMRect instead of defining a new dictionary type?

The main thing we disliked about DOMRect was that it's values are floating point. There are media APIs that handle the crop as floating point values, but this only makes sense for rendering, not for copying raw bytes.

We definitely aren't attached to the name "Region". VideoFrameRect or similar would be fine. We also slightly prefer the top/left naming to DOMRect's x/y, but that's not too important. 

> PlaneInit now takes a BufferSource and an offset, what happens if someone constructs a VideoFrame with a set of planes that each refer to different buffers? The idea of having a per-plane offset suggests that we expect all planes to be within a single buffer, so I think PlaneInit should be a single BufferSource plus a Sequence of PlaneLayouts.
> 
> Alternatively we should truly allow for separate buffers per-plane, which in theory should be possible (and describable by our internal C++ descriptors in Gecko's implementation), but doesn't seem to happen in practice (from the existing decoders in Gecko).

The current shape is flexible: each plane can use the same buffer or a different buffer. The offset refers to whatever they give (should default to zero). 

But, we agree that using the same buffer will be more common. We're open to taking that path if you prefer. If so, we propose the following single-buffer constructor
`VideoFrame(BufferSource buffer, VideoFramePlaneInit init)`

And we would add a new `sequence<PlaneLayout> layout;` member to  `VideoFramePlaneInit` to describe the offset/stride of each plane in the single `buffer`.

> As is being discussed, adding an option to do an async readback seems like a really good idea.

Awesome.

index.src.html
chcunningham: editors call: Resolved to call this allocationSize()

chcunningham:
@padenot I still need to define some spec steps to tie everything together, but this should now at least be at a point where the definitions are all there and crop stuff is fully removed. I'll finish it up Thursday afternoon.

chcunningham:
@padenot 

The latest commit does everything except impl steps for vf.allocationSize() and vf.copyTo(). Listing those steps is so far looking like a large amount of work. Still pondering the approach. 

Re: #165, I think we at least want each VF pixel format to specify
- num planes
- sample size of plane[i]
- bytes per sample of plane[i]

Maybe more than that, but those were the dfns I found myself wanting so far. Chat tomorrow.

chcunningham:
@padenot @aboba this is now completely specified and ready for review. 

@sandersdan you might double check the new VF algos.

@padenot, above I wrote: 
> Re: #165, I think we at least want each VF pixel format to specify
> 
> * num planes
> * sample size of plane[i]
> * bytes per sample of plane[i]

Having now written the algorithms, I would add to that list
* "plane" 
* stride

index.src.html
```diff
@@ -2927,27 +2998,15 @@
...
-
-            ISSUE: The spec should define explicit rules for each
-                {{PixelFormat}} and reference them in the step above. See
-                [#165](https://github.com/w3c/webcodecs/issues/165).
-
-        3. Assign |planes| to {{VideoFrame/planes}}.
-    3. Otherwise (|resource| does not use a recognized {{PixelFormat}}):
-        1. Assign `""` to {{VideoFrame/format}}.
-        2. Assign `null` to {{VideoFrame/planes}}.
+    2. If |resource| uses a recognized {{PixelFormat}}, assign the
```
padenot, line 3001: We'll link this after my PR.

index.src.html
```diff
@@ -2958,80 +3017,232 @@
...
+    4. Let |numPlanes| be the number of planes as defined by
+        {{VideoFrame/[[format]]}}
+    5. While |planeIndex| is less than |numPlanes|:
+
+        NOTE: The following steps validate |rect| is sample-aligned for this
+            frame's {{VideoFrame/[[format]]}}.
+
+        1. Let |plane| be the Plane identified by |planeIndex| as defined by
+            {{VideoFrame/[[format]]}}.
+        2. Let |sampleWidth| be the horizontal pixel size of each subsample
```
padenot, line 3057: We'll link this as well.

index.src.html
```diff
@@ -2958,80 +3017,232 @@
...
+            for |plane|.
+        3. Let |sampleHeight| be the vertical pixel size of each subsample
+            for |plane|.
+        4. If |sourceRect|.{{VideoFrameRect/left}} and
+            |sourceRect|.{{VideoFrameRect/width}} cannot be evenly divided by
+            |sampleWidth|, throw a {{TypeError}}.
```
padenot, line 3063: I see this formulation "is not a multiple of" more often for alignment stuff. I checked HTML and that seems to be the case for imagedata stuff.

index.src.html
```diff
@@ -2958,80 +3017,232 @@
...
+        2. Let |sampleBytes| be the number of bytes per sample for |plane|.
+        3. Let |sampleWidth| be the horizontal pixel size of each subsample
+            for |plane|.
+        4. Let |sampleHeight| be the vertical pixel size of each subsample
+            for |plane|.
+        5. Let |sampleWidthBytes| be the product of multiplying |sampleWidth| by
+            |sampleBytes|.
+        6. Let |computedLayout| be a new [=computed plane layout=].
+        7. Assign quotient of dividing |parsedRect|.{{VideoFrameRect/top}} by
+            |sampleHeight| to |computedLayout|'s [=computed plane layout/sourceTop=].
```
padenot, line 3092: `Set sourceTop to the result of the integer division of top by sampleHeight` read better imo.

index.src.html
```diff
@@ -2958,80 +3017,232 @@
...
+        sample-aligned as determiend by {{VideoFrame/[[format]]}}.
+    * If {{VideoFrameCopyToOptions/layout}} [=map/exists=], a {{PlaneLayout}}
+        must be provided for all planes.
+</div>
+
+: <dfn dict-member for=VideoFrameCopyToOptions>rect</dfn>
+:: The {{VideoFrameRect}} describing the rectangle of pixels to copy from the
+    {{VideoFrame}}. If unspecified, the {{VideoFrame/visibleRect}} will be used.
+
+    NOTE: The coded rectangle can be easily specified by passing
```
padenot, line 3184: `s/easily//`, let's not assume difficulty in specs.

padenot (commented):
Looks good, a few questions though.

index.src.html
```diff
@@ -2819,7 +2816,92 @@
...
+        {{DOMException}}.
+    3. Let |parsedOptions| be the result of running the [=Parse
+        VideoFrameCopyToOptions=] algorithm with |options|.
+    4. If |parsedOptions| is an exception, return a promise rejected with
+        |parsedOptions|.
+    5. If `destionation.byteLength` is less than |parsedOptions|' [=parsed
+        copyto options/allocationSize=], return a promise rejected with a
+        {{TypeError}}.
+    6. Let |p| be a new {{Promise}}.
+    7. Let |copyStepsQueue| be the result of starting a new [=parallel queue=].
```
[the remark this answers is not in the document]
chcunningham: Editors Call: maybe add a note clarifying that these promises don't have to resolve in order of call.

chcunningham:
Editors call: good to merge % comments

chcunningham:
VideoFrame(BufferSource buffer, sequence<PlaneLayout> layout, VFPlaneInit init)

but this drops ability to add alpha without a copy. options to extend (if needed):

VideoFrame((BufferSource or sequence<BufferSource>) buffer, sequence<PlaneLayout> layout, VFPlaneInit init)

or We could add it to VFPlaneInit

index.src.html
chcunningham: Done, here and below.

index.src.html
chcunningham: Fixed here and below.

index.src.html
chcunningham: Perfect!

index.src.html
chcunningham: Ack

index.src.html
```diff
@@ -2839,55 +2920,56 @@
...
-          {{VideoFramePlaneInit/cropWidth}}, and
-          {{VideoFramePlaneInit/cropHeight}}
-          be the crop region of the decoded video frame |output| in
-          pixels, prior to any aspect ratio adjustments.
-      5. Let |displayWidth| and |displayHeight| be the display size of
-          the decoded frame in pixels.
-      6. If |displayAspectWidth| and |displayAspectHeight| are provided,
+  1. Let |frame| be a new {{VideoFrame}}, constructed as follows:
+      1. Assign `false` to {{VideoFrame/[[detached]]}}.
+      2. Let |resource| be the [=media resource=] described by |output|.
```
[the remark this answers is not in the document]
chcunningham: I'd say it as "resource ownership is shared by all the frames that reference the resource. it may additionally be shared by a codec that is using/emitting the resource"

But I worry I maybe misunderstand your intent? We document the sharing in the Memory Model section. We also have this note in decode():

```
NOTE: Authors should call close() on output VideoFrames immediately when frames are no longer needed. The underlying media resources are owned by the VideoDecoder and failing to release them (or waiting for garbage collection) may cause decoding to stall.
```

LMK if more is needed, will send a quick follow up.

index.src.html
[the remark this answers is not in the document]
chcunningham: Note added

```
NOTE: Promises that are returned by several calls to
        {{VideoFrame/copyTo()}} are not guaranteed to resolve in the order they
        were returned.
```

index.src.html
```diff
@@ -2744,52 +2747,46 @@
...
+    convenience for use with {{VideoFrame/allocationSize()}} and
+    {{VideoFrame/copyTo()}}.
+
+    The {{VideoFrame/codedRect}} getter steps are:
+    1. Let |rect| be a new {{VideoFrameRect}}, initialized as follows:
+        1. Assign `0` to {{VideoFrameRect/left}} and {{VideoFrameRect/top}}.
+        2. Assign {{VideoFrame/[[coded width]]}} and
+            {{VideoFrame/[[coded height]]}} to {{VideoFrameRect/width}} and
+            {{VideoFrameRect/height}} respectively.
+    2. Return |rect|.
```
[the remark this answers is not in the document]
chcunningham: I think this one is pretty clear since we called our attributes "top" and "left", but I'm happy to add a note if you still like.

chcunningham (commented):
Thanks for reviewing! I left the link comments unresolved so it would be easy to find them later. A few minor comments also still open w/ reply questions, but I'll send a quick follow up patch for those if needed.

PR #216, opened 2021-05-04
Drop Previous Version link from ED

tidoust:
Spec-prod attempts to override the "Previous version" link in the documents before generation with Bikeshed, but this is not a real override, leading to specs with two Previous Version links, see:
https://github.com/tabatkins/bikeshed/issues/2041

In turn, this makes publication fail when spec gets published multiple times on the same day (because "from biblio" then reports the same URL as the "This Version" link).

Dropping the "Previous version" link from the ED fixes the issue. The link is not useful in an ED in any case.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/pull/216.html" title="Last updated on May 4, 2021, 9:34 AM UTC (da659e4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/216/209a2a0...da659e4.html" title="Last updated on May 4, 2021, 9:34 AM UTC (da659e4)">Diff</a>

### DIFF
--- a/avc_codec_registration.src.html
+++ b/avc_codec_registration.src.html
@@ -7,7 +7,6 @@ Level: none
 Group: mediawg
 ED: https://w3c.github.io/webcodecs/avc_codec_registration.html
 TR: https://www.w3.org/TR/webcodecs-avc-codec-registration/
-Previous Version: from biblio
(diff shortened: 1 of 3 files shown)


chcunningham (approved):
Thanks!

PR #217, opened 2021-05-04
Add job to deploy samples to gh-pages branch

tidoust:
The job runs on pushes to the main branch that touch the "samples" folder and the deploy script itself. It deploys the contents of that folder to the "samples" folder in the "gh-pages" branch.

The job may also be run manually from the GitHub admin UI.


chcunningham (approved):
thank you so much!

PR #225, opened 2021-05-05
Update sample links and add readme.md

tguilbert-google:
WebCodecs is hosted at https://w3c.github.io/webcodecs/, which means that hrefs starting with '/samples/' need to start with '/webcodecs/samples/' instead.

This PR also adds a convenience link to the samples from the github /samples/readme.md.


chcunningham (approved):
LGTM

PR #228, opened 2021-05-06
Define AudioDataCopyToOptions with frameOffset and frameCount

### LINKED ISSUE #179: Audio decoder output to regular buffers and not `AudioBuffer`
Quite a few programs or library don't really use the Web Audio API with the native nodes etc., and rely on either the `AudioWorklet` or even a `ScriptProcessorNode` to play their audio. All the DSP happens in user-code (probably WASM), because that's code that is already written, and the Web is "just" another target in this context.

It seems like we could add an output mode for our audio decoders that would write to a user-provided buffer, to minimize copy.

This would also sidestep the issue where (for now) the `AudioBuffer` object is always planar f32, and quite a few authors would like integers for memory footprint reasons, and also because that's what codec libraries frequently output, no need for a useless conversion.

Forcing the use of `AudioBuffer`s (that have quite a few constraints) when any user of WASM probably has their own audio engine seem like a sub-optimal solution for them. This would also simplify the spec and the code I guess.

### LINKED ISSUE #223: Reading back a PCM data from an `AudioData` object
The discussion starts here https://github.com/w3c/webcodecs/pull/162/files#r625845109 (hopefully that's a stable link?).

chcunningham:
Following up on this TODO
https://github.com/w3c/webcodecs/pull/162/files#r625845109

Fixes #179 
Fixes #223

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/pull/228.html" title="Last updated on May 26, 2021, 9:40 PM UTC (c2a1c9c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/228/47196de...c2a1c9c.html" title="Last updated on May 26, 2021, 9:40 PM UTC (c2a1c9c)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -1960,7 +1960,7 @@ interface AudioData {
   readonly attribute long long duration;     // microseconds
   readonly attribute long long timestamp;    // microseconds
 
-  unsigned long byteLength(AudioDataCopyToOptions options);
+  unsigned long allocationSize(AudioDataCopyToOptions options);
   undefined copyTo([AllowShared] BufferSource destination, AudioDataCopyToOptions options);
   AudioData clone();
   undefined close();
@@ -2064,16 +2064,17 @@ dictionary AudioDataInit {
     3. Return the product of |durationInSeconds| and |microsecondsPerSecond|.
 
 ### Methods ###{#audiodata-methods}
-: <dfn method for=AudioData>byteLength(|options|)</dfn>
+: <dfn method for=AudioData>allocationSize(|options|)</dfn>
 :: Returns the number of bytes required to hold the samples as described by
     |options|.
 
     When invoked, run these steps:
-    1. Let |copyFrameCount| be the result of running the
-        [=Compute Copy Frame Count=] algorithm with |options|.
+    1. Let |copyElementCount| be the result of running the
+        [=Compute Copy Element Count=] algorithm with |options|.
     2. Let |bytesPerSample| be the number of bytes per sample, as defined by
         the {{AudioData/[[sample format]]}}.
-    3. Return the product of multiplying |bytesPerSample| by |copyFrameCount|.
+    3. Return the product of multiplying |bytesPerSample| by
+        |copyElementCount|.
 
 : <dfn method for=AudioData>copyTo(|destination|, |options|)</dfn>
 :: Copies the samples from the specified plane of the {{AudioData}} to the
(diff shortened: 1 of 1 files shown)


index.src.html
```diff
@@ -2049,24 +2049,36 @@
...
+        |options|.{{AudioDataCopyToOptions/frameOffset}} from |frameCount|.
+    5. If |options|.{{AudioDataCopyToOptions/frameCount}} [=map/exists=]:
+        1. If |options|.{{AudioDataCopyToOptions/frameCount}} is greater than
+            |copyFrameCount|, throw a {{RangeError}}.
+        2. Otherwise, assign |options|.{{AudioDataCopyToOptions/frameCount}}
+            to |copyFrameCount|.
+    6. Let |bytesPerSample| be the number of bytes per sample, as defined by
+        the {{AudioData/[[sample format]]}}.
+    7. If the product of multiplying |bytesPerSample| by |copyFrameCount| is
+        greater than `destination.byteLength`, throw a {{RangeError}}.
```
padenot, line 2075: As said during the call, this is multiplying the sample size with a frame count, there needs to be a multiplication by the channel count if the source is interleaved. Maybe this is also an opportunity to do the conversion? We might want to do this as a separate method, not sure.

index.src.html
```diff
@@ -2102,6 +2114,27 @@
...
+<xmp class='idl'>
+dictionary AudioDataCopyToOptions {
+  required unsigned long planeNumber;
+  unsigned long frameOffset = 0;
+  unsigned long frameCount;
+};
+</xmp>
+
+: <dfn dict-member for=AudioDataCopyToOptions>planeNumber</dfn>
+:: The number identifying the plane to copy from.
```
padenot, line 2128: This is more like an index, not just a number.

padenot (commented):
It will be easier to decide how this needs to work when the output format is specced precisely.

dalecurtis:
+1 to adding the sample format to the copyTo options or also having a convertTo(format, options). These are all easy to convert, but as an implementation quality issue a NotSupportedError could be thrown if an implementation doesn't support it for now.

index.src.html
chcunningham: I've changed step 2 at the top to make it such that copyFrameCount includes all channels in the case of single-plane interleaved formats. I think that means step 7 here is now correct.

I do like the idea of conversion via this method, but I'd like to send a follow up PR for that, probably after I've ironed out the AudioSampleFormat definitions as you mentioned.  One question though: if we allow conversion here, should we make the semantics of options.frameOffset and options.frameCount be post-conversion? That is, convert first, then offset into the converted buffer?

chcunningham: I've changed the earlier steps at the top to make it such that, for single plane interleaved formats, this copyFrameCount actually includes frames across all channels. I think this makes the wording in step 7 correct (and reflects my original intent in this case too). LMK if it works. 

I do like the idea of conversion here, but I'd like to do that in a follow up PR. Potentially after I've expanded the definitions of AudioSampleFormats as you mentioned. One question though: if we add conversion, do we apply the options.frameOffset and options.frameCount pre or post converting? I'm thinking post, but not certain.

index.src.html

index.src.html
```diff
@@ -2075,37 +2064,37 @@
     3. Return the product of |durationInSeconds| and |microsecondsPerSecond|.
 
 ### Methods ###{#audiodata-methods}
-: <dfn method for=AudioData>
-      copyTo(|destination|, |options|)
-    </dfn>
+: <dfn method for=AudioData>byteLength(|options|)</dfn>
```
chcunningham, line 2067: I've made this a method because I realized it wasn't very useful as an attribute. The idea is to tell folks who want to call copyTo() how big their buffer needs to be. 

This has nice symmetry with what's proposed in the VideoFrame readback PR. Its named videoFrame.allocationSize() there, but we've agreed to change that to byteLength() as well.

padenot: Not a fan of the name, but the concept is nice and I agree this is better as a method.

`byteLengthForCopy` or something.

padenot (changes requested):

index.src.html
chcunningham: editors call: lets go back to allocationSize() (same for VF)

index.src.html
chcunningham: > The offset and length is pre-conversion, but since it's an element count, it doesn't change post-conversion.

Yes, should be true now.

> If one wants to place frames at a specific offset in an existing buffer this can be done using a typed array view, no ?

I agree

index.src.html
```diff
@@ -2126,6 +2115,25 @@
 
 ### Algorithms ### {#audiodata-algorithms}
 
+: <dfn>Compute Copy Frame Count</dfn> (with |options|)
+:: Run these steps:
+    1. Let |frameCount| be the number of frames in the plane identified by
+        |options|.{{AudioDataCopyToOptions/planeIndex}}.
+
+        NOTE: For interleaved formats with a single plane, |frameCount| will
+            be the total number of frames accross all channels.
```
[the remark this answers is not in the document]
chcunningham: Thanks, hopefully fixed! Algo now returns number of elements, including multiplication by channel count for interleaved formats.

index.src.html
chcunningham: now named allocationSize()

chcunningham (commented):
Thanks Paul

padenot (approved):

padenot:
Looks good, thanks! Can you rebase it so I can rebase #256 and merge it?

chcunningham:
Editors call: @aboba says LGTM

chcunningham:
Thanks! Rebased and proceeding to merge.

PR #229, opened 2021-05-06
Fix WebIDL

tidoust:
Optional dictionary arg must have a default value of `{}`, see related text in
WebIDL in:
https://heycam.github.io/webidl/#dfn-optional-argument-default-value

"If the type of an argument is a dictionary type or a union type that has a
dictionary type as one of its flattened member types, and that dictionary type
and its ancestors have no required members, and the argument is either the final
argument or is followed only by optional arguments, then the argument must be
specified as optional and have a default value provided."

Also, interface definitions must have an "Exposed" extended attribute.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/pull/229.html" title="Last updated on May 6, 2021, 7:07 AM UTC (d2ab661)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/229/ed5ae3c...d2ab661.html" title="Last updated on May 6, 2021, 7:07 AM UTC (d2ab661)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -2719,7 +2719,7 @@ interface ImageDecoder {
   readonly attribute Promise<undefined> completed;
   readonly attribute ImageTrackList tracks;
 
-  Promise<ImageDecodeResult> decode(optional ImageDecodeOptions options);
+  Promise<ImageDecodeResult> decode(optional ImageDecodeOptions options = {});
   undefined reset();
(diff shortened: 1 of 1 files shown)


chcunningham (approved):
Thanks for cleaning up after me ;)

PR #248, opened 2021-05-17
Make timestamps/durations int64 microseconds

### LINKED ISSUE #122: change units of timestamp or rename to timestampMicros
Nearly every existing web API dealing with timestamps uses [DOMTimeStamp](https://developer.mozilla.org/en-US/docs/Web/API/DOMTimeStamp) or [DOMHighResTimeStamp](https://developer.mozilla.org/en-US/docs/Web/API/DOMHighResTimeStamp), both of which are defined to be in units of milliseconds. However, the timestamp attribute found throughout [the WebCodecs spec](https://wicg.github.io/web-codecs/) has units of microseconds. This is quite confusing and has resulted in several bugs in my use of WebCodecs.

I suggest either changing the units of timestamp to DOMHighResTimeStamp or renaming the attribute timestampMicros to make it clear that it is different from other timestamps.

I'll also point out that [HTMLMediaElement.currentTime](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/currentTime) is in units of seconds. The timestamp attribute of WebCodecs is analogous to the currentTime of a media element, so using seconds as the unit could also be an option.

chcunningham:
Fixes #122

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/pull/248.html" title="Last updated on May 17, 2021, 3:43 AM UTC (bfc9465)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/248/ec00a4f...bfc9465.html" title="Last updated on May 17, 2021, 3:43 AM UTC (bfc9465)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -1783,7 +1783,7 @@ EncodedAudioChunk Interface {#encodedaudiochunk-interface}
 interface EncodedAudioChunk {
   constructor(EncodedAudioChunkInit init);
   readonly attribute EncodedAudioChunkType type;
-  readonly attribute unsigned long long timestamp;  // microseconds
+  readonly attribute long long timestamp;    // microseconds
   readonly attribute unsigned long byteLength;
 
   undefined copyTo(ArrayBufferView dst);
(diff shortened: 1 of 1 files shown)


padenot (approved):
:+1:

chcunningham:
Thanks! @aboba I'll go ahead and land since this is a totally mechanical change. But late feedback welcome.

PR #250, opened 2021-05-19
Remove crop* from VideoDecoderConfig

tguilbert-google:
VideoDecoderConfig's cropHeight, cropWidth, cropTop and cropLeft are all deprecated, and need to be removed.

Remaining references to crop* on VideoFrame and VideoFramePlaneInit will be removed in upcoming PRs dealing with VisibleRegions.

This should address issue #94.

***
<a href="https://pr-preview.s3.amazonaws.com/tguilbert-google/webcodecs/pull/250.html" title="Last updated on May 26, 2021, 10:12 PM UTC (48e8fdb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/250/0db2aae...tguilbert-google:48e8fdb.html" title="Last updated on May 26, 2021, 10:12 PM UTC (48e8fdb)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -120,7 +120,7 @@ Definitions {#definitions}
 :: Resources including CPU memory, GPU memory, and exclusive handles to specific
     decoding/encoding hardware that may be allocated by the User Agent as part
     of codec configuration or generation of {{AudioData}} and {{VideoFrame}}
-    objects. Such resources may be quickly exhuasted and should be released
+    objects. Such resources may be quickly exhausted and should be released
     immediately when no longer in use.
 
 : <dfn>Progressive Image</dfn>
@@ -1440,6 +1440,10 @@ run these steps:
     {{VideoDecoderConfig/displayHeight}} = 0, return `false`.
 4. Return `true`.
 
+NOTE: A user agent may also reject a config whose {{VideoDecoderConfig/codedWidth}}
+  or {{VideoDecoderConfig/codedHeight}} exceeds reasonnable bounds, determined by
+  which {{VideoDecoder/[[codec implementation]]}} are available to the user agent.
+
 <dl>
   <dt><dfn dict-member for=VideoDecoderConfig>codec</dfn></dt>
   <dd>Contains a codec string describing the codec.</dd>
@@ -1465,14 +1469,19 @@ run these steps:
         adjustments.
   </dd>
 
+  NOTE: {{VideoDecoderConfig/codedWidth}} and {{VideoDecoderConfig/codedHeight}}
+    are only used as hints when selecting a {{VideoDecoder/[[codec implementation]]}}.
+
   <dt><dfn dict-member for=VideoDecoderConfig>displayWidth</dfn></dt>
   <dd>
-    Width of the VideoFrame when displayed.
+    Width of the VideoFrame when displayed. Defaults to {{VideoFrame/cropWidth}}
+    if not present.
   </dd>
 
   <dt><dfn dict-member for=VideoDecoderConfig>displayHeight</dfn></dt>
   <dd>
-    Height of the VideoFrame when displayed.
+    Height of the VideoFrame when displayed. Defaults to {{VideoFrame/cropHeight}}
+    if not present.
   </dd>
 
   <dt><dfn dict-member for=VideoDecoderConfig>hardwareAcceleration</dfn></dt>
--- a/index.src.html
+++ b/index.src.html
@@ -1495,12 +1495,8 @@ dictionary VideoDecoderConfig {
   BufferSource description;
   [EnforceRange] unsigned long codedWidth;
   [EnforceRange] unsigned long codedHeight;
-  [EnforceRange] unsigned long cropLeft;
-  [EnforceRange] unsigned long cropTop;
-  [EnforceRange] unsigned long cropWidth;
-  [EnforceRange] unsigned long cropHeight;
-  [EnforceRange] unsigned long displayWidth;
-  [EnforceRange] unsigned long displayHeight;
+  [EnforceRange] unsigned long displayAspectWidth;
+  [EnforceRange] unsigned long displayAspectHeight;
   HardwareAcceleration hardwareAcceleration = "allow";
 };
 </xmp>
--- a/index.src.html
+++ b/index.src.html
@@ -688,12 +688,12 @@ Algorithms {#videodecoder-algorithms}
             be the crop region of the decoded video frame |output| in
             pixels, prior to any aspect ratio adjustments.
         5. If |displayAspectWidth| and |displayAspectWidth| are provided,
-            assign the value of the decoded video frame in pixel to either
+            assign the value of the decoded video frame size in pixel to either
             {{VideoFramePlaneInit/displayWidth}} or
             {{VideoFramePlaneInit/displayHeight}}, and scale up the other
             dimension to match the ratio of |displayAspectWidth| and
-            |displayAspectHeight|, as described in
-            [=intrinsic width|HTMLVideoElement=].
+            |displayAspectHeight|, as described by the definitions of
+            [=intrinsic width|intrinsic size=].
         6. Otherwise, let {{VideoFramePlaneInit/displayWidth}} and
             {{VideoFramePlaneInit/displayHeight}} be the display size of
             the decoded video frame in pixels.
@@ -1486,19 +1486,19 @@ run these steps:
 
   <dt><dfn dict-member for=VideoDecoderConfig>displayAspectWidth</dfn></dt>
   <dd>
-    Width of the VideoFrame's aspect ratio when displayed.
+    Horizontal component of the VideoFrame's aspect ratio when displayed.
   </dd>
 
   <dt><dfn dict-member for=VideoDecoderConfig>displayAspectHeight</dfn></dt>
   <dd>
-    Height of the VideoFrame's aspect ratio when displayed.
+    Vertical component of the VideoFrame's aspect ratio when displayed.
   </dd>
 
-NOTE: User agents are expected to scale up a {{VideoFrame|VideoFrame's}} {{VideoFrame/displayWidth}} or
- {{VideoFrame/displayHeight}} to match the ratio of
- {{VideoDecoderConfig/displayAspectWidth}} and
- {{VideoDecoderConfig/displayAspectHeight}}, as defined in the
- [=intrinsic width|HTMLVideoElement=] spec.
+Note: {{VideoFrame/displayWidth}} and {{VideoFrame/displayHeight}} can both be
+  different from {{VideoDecoderConfig/displayAspectWidth}} and
+  {{VideoDecoderConfig/displayAspectHeight}}, but they should have identical
+  ratios, after being scaled according to defintions of
+  [=intrinsic width|intrinsic size=].
 
   <dt><dfn dict-member for=VideoDecoderConfig>hardwareAcceleration</dfn></dt>
   <dd>
--- a/index.src.html
+++ b/index.src.html
@@ -656,14 +656,15 @@ Algorithms {#videodecoder-algorithms}
         1. Let |timestamp| and |duration| be the
             {{EncodedVideoChunk/timestamp}} and {{EncodedVideoChunk/duration}}
             from the {{EncodedVideoChunk}} associated with |output|.
-        2. Let |displayAspectWidth| and |displayAspectHeight| be the
-            {{VideoDecoderConfig/displayAspectWidth}} and
-            {{VideoDecoderConfig/displayAspectHeight}} of the
-            {{VideoDecoder/[[active decoder config]]}} respectively.
-        3. Let |frame| be the result of running the [=Create a VideoFrame=]
+        2. Let |displayAspectWidth| and |displayAspectHeight| be undefined.
+        3. If {{VideoDecoderConfig/displayAspectWidth}} and
+            {{VideoDecoderConfig/displayAspectHeight}} [=map/exist=] in the
+            {{VideoDecoder/[[active decoder config]]}}, assign their values to
+            |displayAspectWidth| and |displayAspectHeight| respectively.
+        4. Let |frame| be the result of running the [=Create a VideoFrame=]
             algorithm with |output|, |timestamp|, |duration|, |displayAspectWidth|
             and |displayAspectHeight|.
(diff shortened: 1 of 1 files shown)


sandersdan:
Note: regions proposal is https://github.com/w3c/webcodecs/issues/157, PR is https://github.com/w3c/webcodecs/pull/208.

index.src.html
```diff
@@ -1477,40 +1465,14 @@
         adjustments.
```
chcunningham, line 1465: @sandersdan, for codedHeight and codedWidth, should we add notes to these that say how they're used (as a hint to guide selection of the [[codec impl]])?

index.src.html
```diff
@@ -1477,40 +1465,14 @@
...
-  <dd>
-    The height in pixels to include in the crop, starting from cropLeft.
-        Defaults to codedHeight if not present.
-  </dd>
-
   <dt><dfn dict-member for=VideoDecoderConfig>displayWidth</dfn></dt>
   <dd>
-    Width of the VideoFrame when displayed. Defaults to cropWidth if not
-        present.
+    Width of the VideoFrame when displayed.
```
chcunningham, line 1470: Here and below, I would still add mention of the default, which is now {{VideoFrame/cropWidth}}

chcunningham (changes requested):
Generally looks good.

We're missing something (not your fault): we don't actually properly use config.displayWidth|Height to set the videoFrame.displayWidth|height in the decoder output algo. To do this, we need to save the active config in a slot, similar to how we did for the encoder. @sandersdan, my read of the issue convo is we prefer config display sizing to override inband sizing if provided. Sound right?

index.src.html
sandersdan: I think that's the only remaining use, yes. It might be more than a hint in that an implementation should be able to outright reject unsupported sizes.

index.src.html
sandersdan: Note that the interpretation is as an aspect ratio, and the actual display size of a frame is the same as for `<video>` (github.com/whatwg/html/commit/2e94aa64fcf9adbd2f70d8c2aecd192c8678e298)--one dimension of the visible size is scaled up until the ratio is the same as the display size.

sandersdan:
> we prefer config display sizing to override inband sizing if provided. Sound right?

That's correct.

index.src.html
tguilbert-google: I added a note after the algorithm, but it might be better placed under the dict-member definitions.

The note itself might too ambiguous, should upper bounds checking be part of the algorithm?

tguilbert-google (commented):
Updated, but did not address the point about the active config slot.

index.src.html
tguilbert-google: I removed the note following discussions with @chcunningham. This goes in the "unsupported" errors, and not the "invalid errors".

index.src.html
```diff
@@ -678,7 +687,14 @@
...
             be the crop region of the decoded video frame |output| in
             pixels, prior to any aspect ratio adjustments.
-        5. Let {{VideoFramePlaneInit/displayWidth}} and
+        5. If |displayAspectWidth| and |displayAspectWidth| are provided,
+            assign the value of the decoded video frame in pixel to either
+            {{VideoFramePlaneInit/displayWidth}} or
+            {{VideoFramePlaneInit/displayHeight}}, and scale up the other
+            dimension to match the ratio of |displayAspectWidth| and
+            |displayAspectHeight|, as described in
+            [=intrinsic width|HTMLVideoElement=].
```
tguilbert-google, line 696: Is there a better way to say this? Or at least to reference the spec?

index.src.html
```diff
@@ -1477,41 +1481,24 @@
...
-    Width of the VideoFrame when displayed. Defaults to cropWidth if not
-        present.
-  </dd>
-
-  <dt><dfn dict-member for=VideoDecoderConfig>displayHeight</dfn></dt>
-  <dd>
-    Height of the VideoFrame when displayed. Defaults to cropHeight if not
-        present.
-  </dd>
+NOTE: User agents are expected to scale up a {{VideoFrame|VideoFrame's}} {{VideoFrame/displayWidth}} or
```
tguilbert-google, line 1497: Is this not useful here, or should the definition of the members be expanded?

tguilbert-google: Updated in the latest commit.

index.src.html
tguilbert-google: I think this is addressed.

index.src.html
tguilbert-google: I added a reference to that spec, and refer to it in a few places.

index.src.html
chcunningham: I would say "as described by [=intrinsic size=] in [[!HTML]]". But I would probably drop the reference. It describes the historical behavior for <video>, but it doesn't clearly point to a dfn for us to use in our algo steps.

My edit:

a. Let displayWidht and displayHeight be the display size of the decoded frame in pixels
b. if displayAspectWidth and displayAspectHeight are provided, increase displayWidth or displayHeight until the ratio of displayWidth : displayHeight equals that of displayAspectWidth : displayAspectHeight.
c. Assign displayWidth and displayHeight to {{VideoFramePlaneInit/displayWidth}} and {{VideoFramePlaneInit/displayHeight}} respectively.

index.src.html
```diff
@@ -1442,15 +1452,9 @@
...
-3. If {{VideoDecoderConfig/cropWidth}} = 0 or {{VideoDecoderConfig/cropHeight}}
-    = 0, return `false`.
-4. If {{VideoDecoderConfig/cropTop}} + {{VideoDecoderConfig/cropHeight}} >=
-    {{VideoDecoderConfig/codedHeight}}, return `false`.
-5. If {{VideoDecoderConfig/cropLeft}} + {{VideoDecoderConfig/cropWidth}} >=
-    {{VideoDecoderConfig/codedWidth}}, return `false`.
-6. If {{VideoDecoderConfig/displayWidth}} = 0 or
-    {{VideoDecoderConfig/displayHeight}} = 0, return `false`.
-7. Return `true`.
+3. If {{VideoDecoderConfig/displayAspectWidth}} = 0 or
```
chcunningham, line 1455: Can you add a check that if one is provided, the both are

index.src.html
```diff
@@ -1477,41 +1481,24 @@
...
-        aspect ratio adjustments. Defaults to zero if not present.
-  </dd>
-
-  <dt><dfn dict-member for=VideoDecoderConfig>cropTop</dfn></dt>
-  <dd>
-    The number of pixels to remove from the top of the VideoFrame, prior to
-        aspect ratio adjustments. Defaults to zero if not present.
-  </dd>
+  NOTE: {{VideoDecoderConfig/codedWidth}} and {{VideoDecoderConfig/codedHeight}}
+    are only used as hints when selecting a {{VideoDecoder/[[codec implementation]]}}.
```
chcunningham, line 1485: nit: drop 'only', as they may be used to reject support.

index.src.html
```diff
@@ -1477,41 +1481,24 @@
...
-  <dt><dfn dict-member for=VideoDecoderConfig>displayHeight</dfn></dt>
-  <dd>
-    Height of the VideoFrame when displayed. Defaults to cropHeight if not
-        present.
-  </dd>
+Note: {{VideoFrame/displayWidth}} and {{VideoFrame/displayHeight}} can both be
+  different from {{VideoDecoderConfig/displayAspectWidth}} and
+  {{VideoDecoderConfig/displayAspectHeight}}, but they should have identical
+  ratios, after being scaled according to defintions of
+  [=intrinsic width|intrinsic size=].
```
chcunningham, line 1501: I'd replace this link w/ a link to the create videoframe algo.

index.src.html
chcunningham: For those following along, see my comment for in the issue
https://github.com/w3c/webcodecs/issues/94#issuecomment-845483687

index.src.html
tguilbert-google: Done.

index.src.html
tguilbert-google: I'm not sure I follow. The aspect ratio is not marked as optional, so this already makes sure we have both dAW and dAH in all cases.

index.src.html
tguilbert-google: Done

index.src.html
tguilbert-google: Done.

index.src.html
chcunningham: dictionary members are implicitly optional. you can mark them "required", but we like optional for these.

index.src.html
chcunningham: sorry, just edit'ed this comment - my first attempt trailed off because I confused the markdown editor w/ a video tag

index.src.html
```diff
@@ -678,7 +687,14 @@
             {{VideoFramePlaneInit/cropHeight}}
             be the crop region of the decoded video frame |output| in
             pixels, prior to any aspect ratio adjustments.
-        5. Let {{VideoFramePlaneInit/displayWidth}} and
+        5. If |displayAspectWidth| and |displayAspectWidth| are provided,
```
padenot, line 690: Change the second `displayAspectWidth` by `Height`.

index.src.html
```diff
@@ -1442,15 +1452,9 @@
...
-    = 0, return `false`.
-4. If {{VideoDecoderConfig/cropTop}} + {{VideoDecoderConfig/cropHeight}} >=
-    {{VideoDecoderConfig/codedHeight}}, return `false`.
-5. If {{VideoDecoderConfig/cropLeft}} + {{VideoDecoderConfig/cropWidth}} >=
-    {{VideoDecoderConfig/codedWidth}}, return `false`.
-6. If {{VideoDecoderConfig/displayWidth}} = 0 or
-    {{VideoDecoderConfig/displayHeight}} = 0, return `false`.
-7. Return `true`.
+3. If {{VideoDecoderConfig/displayAspectWidth}} = 0 or
+    {{VideoDecoderConfig/displayAspectWidth}} = 0, return `false`.
```
padenot, line 1456: Ditto

index.src.html
```diff
@@ -1477,41 +1481,24 @@
...
-        aspect ratio adjustments. Defaults to zero if not present.
-  </dd>
-
-  <dt><dfn dict-member for=VideoDecoderConfig>cropTop</dfn></dt>
-  <dd>
-    The number of pixels to remove from the top of the VideoFrame, prior to
-        aspect ratio adjustments. Defaults to zero if not present.
-  </dd>
+  NOTE: {{VideoDecoderConfig/codedWidth}} and {{VideoDecoderConfig/codedHeight}}
+    are used as hints when selecting a {{VideoDecoder/[[codec implementation]]}}.
```
padenot, line 1485: It's not really a hint, it's just something that helps picking the right decoder.

index.src.html
```diff
@@ -1477,41 +1481,24 @@
...
 
-  <dt><dfn dict-member for=VideoDecoderConfig>cropWidth</dfn></dt>
+  <dt><dfn dict-member for=VideoDecoderConfig>displayAspectWidth</dfn></dt>
   <dd>
-    The width in pixels to include in the crop, starting from cropLeft.
-        Defaults to codedWidth if not present.
+    Horizontal component of the VideoFrame's aspect ratio when displayed.
```
padenot, line 1489: Maybe use another word than "component", as this is the term used for something else ? Maybe simply "dimension" ?

padenot (commented):
Looks good, just a few mistakes that are trivial to fix, and a couple of suggestions for improved clarity. Thanks!

index.src.html
tguilbert-google: Ah, thank you! Updated.

index.src.html
tguilbert-google: Seems good, updated.

index.src.html
tguilbert-google: Removed.

index.src.html
tguilbert-google: Done.

index.src.html
tguilbert-google: Done.

index.src.html
tguilbert-google: This is a nice clean way of saying this. TY!

index.src.html
```diff
@@ -1477,41 +1452,24 @@
...
-        aspect ratio adjustments. Defaults to zero if not present.
-  </dd>
-
-  <dt><dfn dict-member for=VideoDecoderConfig>cropWidth</dfn></dt>
-  <dd>
-    The width in pixels to include in the crop, starting from cropLeft.
-        Defaults to codedWidth if not present.
-  </dd>
+  NOTE: {{VideoDecoderConfig/codedWidth}} and {{VideoDecoderConfig/codedHeight}}
+    are when selecting a {{VideoDecoder/[[codec implementation]]}}.
```
chcunningham, line 1456: nit: probably missing a word. "are _used_ when selecting..."

index.src.html
```diff
@@ -653,36 +656,14 @@
...
-            {{VideoFramePlaneInit/cropWidth}}, and
-            {{VideoFramePlaneInit/cropHeight}}
-            be the crop region of the decoded video frame |output| in
-            pixels, prior to any aspect ratio adjustments.
-        5. Let {{VideoFramePlaneInit/displayWidth}} and
-            {{VideoFramePlaneInit/displayHeight}} be the display size of
-            the decoded video frame in pixels.
-    4. Return a new {{VideoFrame}}, constructed with |pixelFormat|,
-        |planes|, and |init|.
+        2. Let |displayAspectWidth| and |displayAspectHeight| be the
```
chcunningham, line 659: Needs handling for when dispalyAspectWidth/Height are not provided

2. Let |displayAspectWidth| and |displayAspectHeight| be `undefined`.
3. If {{VideoDecoderConfig/displayAspectWidth}} and {{VideoDecoderConfig/displayAspectHeight}} [=map/exist=] in the {{VideoDecoder/[[active decoder config]]}}, assign their values to |displayAspectWidth| and |displayAsepctHeight| respectively.

chcunningham (approved):
LGTM % final nits.

padenot (approved):

chcunningham:
Editors call: @aboba LGTM

PR #251, opened 2021-05-20
Fix semi-broken link to AVC WebCodecs Registration

tidoust:
The link to the AVC spec is fine in the ED of the registry spec, but breaks when the spec is published to /TR because the AVC spec gets published to a different folder.

This update switches link to use an absolute URL.


chcunningham (approved):

chcunningham:
Thanks!

PR #271, opened 2021-06-07
Use slots instead of directly referencing the attributes

padenot:
This does not change any behaviour. I did not do `VideoFrame` yet because it's moving too much, I'll do it after.

This is split in multiple commit, to make it a lot easier to review.


padenot:
> VideoFrame should mostly have slots defined already, but maybe I misunderstand what you're planning? You might wait for that one until after the #208 lands... if you're planning lots of churn the merge could get ugly.

Simply a proofreading to make sure we're good and fix any typos, like the `ImageDecoder/complete` above, almost nothing.

PR #274, opened 2021-06-08
Add AlphaMode to keep/discard alpha channel

### LINKED ISSUE #207: Alpha support
I've been meaning to send a PR for the latest thinking from the Chrome explorations... I'll try to have it out this week. 

Generally, my proposal will be:
* add a `attribute bool hasAlpha` to `VideoFrame`. The semantics for planar formats would be that the last plane (highest index) is the alpha plane. More on this in https://github.com/w3c/webcodecs/issues/200#issuecomment-829535780
* Define an `enum AlphaOption { keep, discard }`, and add attributes/members using this enum to `VideoFrameInit` and `VideoEncoderConfig`. Choose defaults based on expected use cases. 
  * For `VideoEncoderConfig.alpha`, default to `=discard`. Alpha encoding is a thing, but relatively rare. In the event that the `VideoFrame` is backed by a texture, we can use `=discard` to optimize readback to a alpha-less format that is supported by the `VideoEncoder` (e.g. NV12)
  * For `VideoFrameInit.alpha`, default to `=keep`. Seems least disruptive in the event you weren't planning to encode the frame (which is already covered above).
* For `VideoEncoder`, if configured with `alpha=keep` (and if supported), we emit the alpha data as side data in the chunk output callback. This follows the path of existing underlying encoders (e.g. ffmpeg's `AV_PKT_DATA_MATROSKA_BLOCKADDITIONAL`). 
  * The output callback already emits chunk metadata in the form of the `VideoDecoderConfig` needed to decode the chunk. PR #187 is expanding that into a metadata bucket for homing chunk SVC metadata. I propose to just add alpha side data there as well... TBD some naming bikeshed. 

Note: we haven't yet thought about how to do *decoding* of alpha. Something to consider.

tguilbert-google:
This PR adds the hasAlpha attribute to VideoFrame, and the AlphaOption to keep/discard alpha data when initializing or encoding VideoFrames.

Fixes #207

***
<a href="https://pr-preview.s3.amazonaws.com/tguilbert-google/webcodecs/pull/274.html" title="Last updated on Jul 21, 2021, 4:27 AM UTC (5890816)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/274/02d54e9...tguilbert-google:5890816.html" title="Last updated on Jul 21, 2021, 4:27 AM UTC (5890816)">Diff</a>


aboba (commented):
Looking at `EncodedVideoChunkMetadata` (and recent discussions about extension) makes me wonder whether we shouldn't separate it out better (e.g. put `temporalLayerId` into its own dictionary). 

See: https://github.com/w3c/webrtc-encoded-transform/issues/9

chcunningham:
editors call:
- is alpha core for RTC use case as wll -> yes
- should we separate SVC metadata into its own sub-bucket -> sure (svcMetadata)

index.src.html
```diff
@@ -2530,6 +2548,11 @@
   [EnforceRange] unsigned long long duration;  // microseconds
   [EnforceRange] long long timestamp;          // microseconds
 };
+
+enum AlphaOption {
```
chcunningham, line 2552: Since this is used both by VideoFrame and VideoEncoderConfig, I would move this definition to its own section under configs (parallel to hardware acceleration) and add a few short <dfn> blurbs about the values.

index.src.html
```diff
@@ -2819,6 +2842,18 @@
     The {{VideoFrame/duration}} getter steps are to return
     {{VideoFrame/[[duration]]}}.
 
+: <dfn attribute for=VideoFrame>hasAlpha</dfn>
+:: Whether or not the {{VideoFrame}} has an alpha channel. If this is `false`,
```
chcunningham, line 2846: @sandersdan do you like "channel" or "component" for this? Same same?

index.src.html
```diff
@@ -2819,6 +2842,18 @@
...
+    the frame is considered to be opaque, and must be rendered as such.
```
chcunningham, line 2847: I'd expand more on the contrasting false vs true behavior. Something like

When false 
- the frame must be treated as fully opaque (for example, by any method which takes a CanvasImageSource)
- for planar formats, an alpha plane must not be present
- for interleaved formats, blocks of samples will contain an 'X' channel where the alpha channel may otherwise have been. The position of the 'X' channel is defined by [[format]]. The values of the 'X' channel are undefined and must be ignored when rendering.

When true 
- the frame must be considered transparent for the pixels and degrees indicated by the alpha values.
- for planar formats, the alpha plane must appear as an additional plane following the non-alpha planes defined by [[format]]
- for interleaved formats, the alpha channel will appear at the 'A' position as defined by [[format]]

Just a draft, could probably use additional editing.

index.src.html
```diff
@@ -2819,6 +2842,18 @@
...
+    For planar {{PixelFormat}}s, the last {{Plane}} (highest index) in
+    {{VideoFrame/planes}} is an alpha plane. For interleaved {{PixelFormat}}s,
+    the alpha channel is interleaved within each block of related samples,
+    according to that {{PixelFormat}}'s specifications.
+
+ISSUE: {{VideoFrame}}s with an interleaved {{PixelFormat}} should omit copying
```
chcunningham, line 2853: After further discussion, I think we can go ahead and solve this here and now as described above. "The values of the 'X' channel are undefined and must be ignored when rendering." Implementers are free (and probably will) just copy out the alpha data as is with this additional signal that it should be ignored.

index.src.html
```diff
@@ -2914,11 +2949,14 @@
             {{Plane/stride}}, {{Plane/rows}}, {{Plane/length}}.
         4. Append |plane| to |planes|.
     6. Assign |planes| to |frame|.{{VideoFrame/planes}}.
-    7. If {{VideoFrameInit/duration}} [=map/exists=] in |init|, assign it to
+    7. If |otherFrame|.{{VideoFrame/hasAlpha}} is `true` and
+        |init|.{{VideoFrameInit/alpha}} is `"discard"`, run the
+        [=Discard alpha=] algorithm with |frame|.
```
chcunningham, line 2954: I think we might drop the discard alpha algo entirely. as-is, it seems to mutate the underlying resource for both the original and new frame. But also, I think given how the dust is settling for interleaved, per dfn of hasAlpha above, we can allow the two frames to simply share the same resource and rely on the definition of hasAlpha + format to do the right thing. For this method, you could then just say something like:

if alpha = discard, assign `false` to hasAlpha. otherwise, assign otherframe.hasAlpha to hasAlpha.

index.src.html
```diff
@@ -2530,6 +2548,11 @@
...
   [EnforceRange] long long timestamp;          // microseconds
 };
+
+enum AlphaOption {
+  "keep",
+  "discard",
+};
 </xmp>
 
 ### Internal Slots ###{#videoframe-internal-slots}
```
chcunningham, line 2558: Can you make an internal slot that is [[has alpha]] and make the hasAlpha attr simply proxy that. For example:
https://w3c.github.io/webcodecs/#dom-audiodecoder-state

chcunningham (commented):
Nice work.

chcunningham:
@aboba wrote
> Looking at `EncodedVideoChunkMetadata` (and recent discussions about extension) makes me wonder whether we shouldn't separate it out better (e.g. put `temporalLayerId` into its own dictionary).

@tguilbert-google could you also tackle this (i.e. metadata.temporalLayerId -> metadata.svc.temporalLayerId)? Could be a follow up.

index.src.html
sandersdan: Channel is correct. A component is part of a sample but a plane has many samples.

sandersdan:
I see that the plan has changed from using just  RGBA and BGRA to also include RGBX and BGRX. This doesn't seem very consistent to me:

- I420/I420A can only be distinguished by `hasAlpha`.
- RGBX/RGBA can be distinguished by name alone.

It seems like if we are including 'X' RGB formats, then we should also include 'A' YUV formats. In that case `hasAlpha` is not required at all, it's just a convenience attribute that reflects a property that can be derived from the format name alone.

The original plan of not having 'X' formats was also consistent, in the opposite direction.

I don't see any reasoning for this change?

chcunningham:
In my veiw the plan was never super firm. I was [originally opposed](https://github.com/w3c/webcodecs/issues/207#issuecomment-829701994), and was later [unsure](https://github.com/w3c/webcodecs/issues/207#issuecomment-829710877).

I don't love RGBA + hasAlpha=false. To me that seems confusing given precedent that 'A' means alpha is there while 'X' means its ignored.

My earlier sense from googling this stuff was that 'X' variants of RGB are perhaps more prevalent than 'A' variants of YUV formats. But maybe that's not the case? Seems at least [Ffmpeg](https://source.chromium.org/chromium/chromium/src/+/main:third_party/ffmpeg/libavutil/pixfmt.h;drc=d770909b13329118ed55f852c863b1bd7b8cdef2;l=101) and [Gpac](https://doxygen.gpac.io/group__cst__grp.html) have both. Chrome has both I420 and I420A. I'm open to pursue 'A' and 'X'. Would this then remove the need for frame.hasAlpha entirely?

tguilbert-google:
> @aboba wrote
> 
> > Looking at `EncodedVideoChunkMetadata` (and recent discussions about extension) makes me wonder whether we shouldn't separate it out better (e.g. put `temporalLayerId` into its own dictionary).
> 
> @tguilbert-google could you also tackle this (i.e. metadata.temporalLayerId -> metadata.svc.temporalLayerId)? Could be a follow up.

Yes I can. I will make this a separate PR.

tguilbert-google:
This latest iteration of the spec gets rid of the the `Discard alpha` algorithm, in favor of not exposing alpha when `hasAlpha` is false, even if the underlying media resource has an alpha channel.

If we got rid of `hasAlpha`, would we need to be more specific about how to handle differences in the format of the underlying media resource and that of the video frame.

In that case, it would be important to specify which formats map to which with `AlphaOption.discard`.

chcunningham:
> I don't know how confusing it is. "Opaque" RGBA is quite common while RGBX is not something you are likely to see outside of media. A lot of code is able to handle them identically.
>
> Encoding this information in the format doubles the number of formats, but it also makes the formats more precise. I don't have a strong preference myself.

I also don't have a strong preference. Considering these points, it seems that this is slightly simplifying for users who don't care about the distinction. They don't need list 'x' variants in switch statements on pixel format. For planar formats, I can also imagine common code (e.g. if you do copyTo and just assume the returned layout has 3 planes, you're probably fine). Maybe that's enough reason to go with just hasAlpha, no 'x', no i420a, etc?

@padenot @aboba curious to get your thoughts.

index.src.html
```diff
@@ -2515,6 +2558,7 @@
 dictionary VideoFrameInit {
   unsigned long long duration;  // microseconds
   long long timestamp;          // microseconds
+  AlphaOption alpha = "keep";
 };
 
 dictionary VideoFramePlaneInit {
```
chcunningham, line 2564: If the outcome of pending discussion is that we keep hasAlpha, we should probably add hasAlpha to this structure as well as a signal for how to interpret the data (e.g. you now have an extra plane for planar formats)

chcunningham:
Editors call: @padenot @aboba having further look to consider the questions above.

chcunningham:
(Editors call)

Question: If we do it via boolean, do we then allow for alpha combined with all pixel formats. Is that always sane? Does it create need for more elaborate error handling?

Related questions:
- capabilities signalling: how to know your encoder doesn't support alpha. or a pixel format generally.

@padenot having closer look. @sandersdan, thoughts on the above

dalecurtis:
> * capabilities signalling: how to know your encoder doesn't support alpha. or a pixel format generally.

The `alpha` field is part of the VideoEncoderConfig. So if `alpha == keep` is set isConfigSupported() can reject it. Alpha encoding is always a separate (generally sw encoder), so it's mostly an implementation quality thing.

chcunningham:
> Question: If we do it via boolean, do we then allow for alpha combined with all pixel formats. Is that always sane? Does it create need for more elaborate error handling?

Discussed this with @sandersdan. He concludes that it is technically sane for any format to have "alpha", but this does create the mentioned headaches around UA support and error handling. This means `frame.hasAlpha` seems more error prone, and more complex, while not more useful. Hence, we prefer to drop that attribute and go with having alpha always described in the format string. Planar formats that don't suffix with an 'A' must not have alpha (I420 -> no alpha, I420A -> has alpha). Interleaved formats should signal an empty alpha channel using an X, and a non-empty alpha channel using an A (e.g. RGBX vs RGBA).

@padenot LMK if this sounds good and we'll update the PR accordingly.

padenot:
> @padenot LMK if this sounds good and we'll update the PR accordingly.

I agree with this assessment. Would we expose all permutations at first, or would we only list what is going to be used in practice ? Is everything going to be used? I've never heard of `NV12A` or anything like that, does that even exist? `ffmpeg` doesn't have anything like that.

chcunningham:
> Would we expose all permutations at first, or would we only list what is going to be used in practice?

I would only expose the useful permutations with the implication being that implementers must support whatever we expose. List you've added in #288 looks good for now. I don't think we want to expose the full set of formats in ffmpeg, but that small list should cover most common cases and is probably already supported in every UA.

padenot:
>  I don't think we want to expose the full set of formats in ffmpeg

For sure. I was just using `ffmpeg` format list as a rough check of if a format even exists at all (e.g. NV12 + alpha), because `ffmpeg` is known to be quite a general tool.

index.src.html
```diff
@@ -1815,6 +1825,12 @@
     {{HardwareAcceleration}}.
   </dd>
 
+<dt><dfn dict-member for=VideoEncoderConfig>alpha</dfn></dt>
+  <dd>
+    Whether the alpha component of the {{VideoFrame}} inputs should be kept
+    or discarded prior to encoding.
```
alvestrand, line 1831: Is it stated somewhere how the encoder should behave when it gets a video frame with alpha option of either "keep" or "discard"?
The intuitive behavior (for me) says that "discard" in the encoder overrules a "keep" on the video frame, but this might need to be said explicitly somewhere.

chcunningham: You've got the right idea. @tguilbert-google, wdyt of adding more clarifying detail?

index.src.html
tguilbert-google: This is done.

index.src.html
tguilbert-google: I've added a clarifying sentence. Thanks for pointing this out!

chcunningham:
Editors call: @aboba no concerns. I'll take other pass today, but probably LGTM

chcunningham (approved):
Nice!

chcunningham:
Nit: re-title to something like "Add AlphaMode to keep/discard alpha channel"

padenot (approved):

PR #282, opened 2021-06-16
Add resource reclamation section

### LINKED ISSUE #121: Fire QuotaExceeded if codec resources are exhausted/reclaimed
The UA (and underlying platform) will often have a limit for how many codecs can be simultaneously instantiated. Particularly for hardware codecs. We could throw a QuotaExceeded exception via the error callback when this occurs. 

Additionally, the UA should have the ability to reclaim codec resources if it can identify they're not being used. Maybe this could be done silently (no use -> no sideffects -> why notify). If the app picks up usage later, we could silently reallocate the resources and chug along. If the resources are no longer available, throw the same QuotaExceeded described above. 

Similar consideration: UAs (at least Chromium) may suspended video decoding for `<video>` when tabs are put into the background (and resumed when the tab re-enters foreground). Similar techniques could be used with WebCodecs, which would again throw the QuotaExceeded error (or some other error). This is tricky though: we can't reliably know that VideoDecoder (and kin) isn't doing useful work in the background. We might consider a reserving the right to suspend backgrounded Window & DedicatedWorker VideoDecoders, but offer a different lifetime contract for SharedWorker and ServiceWoker. We haven't thought through the SharedWorker and ServiceWorker usage yet (I'll soon reduce the exposed= to just DedicateWorker in the IDL).

tguilbert-google:
This PR fixes #121.

Areas of this PR that should be scrutinized (amongst others):
- Definition of an inactive codec
- Definition of a background codec
- Whether to fire QuotaExceededError, or a new exception/error.

***
<a href="https://pr-preview.s3.amazonaws.com/tguilbert-google/webcodecs/pull/282.html" title="Last updated on Aug 13, 2021, 4:09 AM UTC (0446a6f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/282/2ef514e...tguilbert-google:0446a6f.html" title="Last updated on Aug 13, 2021, 4:09 AM UTC (0446a6f)">Diff</a>


index.src.html
```diff
@@ -4133,6 +4136,40 @@
...
+When resources are constrained, a User Agent might want to pro-actively reclaim
+codecs. This is particularly true in the case where hardware codecs are limited,
+and shared accross all web pages.
+
+A User Agent is allowed to [=reclaim a codec=], freeing the resources of a
+<dfn>reclaimable codec</dfn>, which is a defined as a codec that is either an
+[=inactive codec=] or a [=background codec=].
+
+An <dfn>inactive codec</dfn> is a codec that has not received a call to
+`encode()` or `decode()` in the past 1 second, and has no pending outputs.
```
sandersdan, line 4151: What's a pending output exactly? I've used that term before to mean that there is no input queued or sent to the codec for which we expect a future output, but that's a per-codec thing that's not easy to state both simply and precisely.

The main difficulty in definition is frame reordering; a codec might be waiting for a future `encode()` or `decode()` before it can output a frame it has already processed, yet `flush()` might not be sufficient to force the output. If you count that as 'inactive' then it's not so hard to define.

index.src.html
```diff
@@ -4133,6 +4136,40 @@
...
+
+A <dfn>background codec</dfn> is a codec whose {{ownerDocument}}'s
+{{Document/hidden}} attribute is {{hidden|"hidden"}}.
+
+When selecting which [=reclaimable codec=] to evict, the User Agent should
+prioritize, in order, the following criteria:
```
sandersdan, line 4157: Is strict ordering necessary? I realize it's "should" not "must", but in my experience heuristics get bumped up in priority all the time.

index.src.html
```diff
@@ -4133,6 +4136,40 @@
...
+    1. What type of resource is the codec holding on to? User Agents must only
```
sandersdan, line 4158: Don't like 'must' here.

index.src.html
```diff
@@ -4133,6 +4136,40 @@
...
+    2. Is the codec both a [=background codec=] and an [=inactive codec=]? These
+        codecs should be prioritized for reclamation, as they are least likely
+        to be disruptive to user experience.
+    2. Otherwise, prefer reclaiming codecs that are [=background codec=]s,
+        before [=inactive codec=]s.
+    4. Any other heuristics, determined by the User Agent. For example,
+        heuristics might prioritize reclaiming the oldest codecs, the ones with
+        the least recent activity, or the least frequent activity.
+
+To <dfn>reclaim a codec</dfn>, a User Agent must run the appropriate
```
sandersdan, line 4169: Maybe this can be bumped higher, before the selection criteria? It seems like the most important part to me.

index.src.html
tguilbert-google: User Agents will naturally try to make smart choices, so the ordering might not be relevant.

Updated to should, and switch from ordered list to bulleted items. I also reworded some things.

index.src.html
tguilbert-google: Done.

index.src.html
tguilbert-google: Removed.

I am purposely leaving "resource" undefined, meaning that this could be HW codec slots, memory, CPU, power. Should these be called out?

index.src.html
tguilbert-google: You're right, this is tricky, and I hadn't thought of reordered frames.

I've tentatively updated the wording to "and has not called its `output()` callback in the past 1 second"... I'm not sure if this is what we want, but the definition as a whole makes some sense to me at least now. "No inputs or outputs for 1 second".

index.src.html
```diff
@@ -4133,6 +4136,42 @@
 : <dfn event for=ImageTrack>change</dfn>
 :: Fired at the {{ImageTrack}} when the {{ImageTrack/frameCount}} is altered.
 
+Resource Reclamation{#resource-reclamation}
+==============================================
+
+When resources are constrained, a User Agent might want to pro-actively reclaim
```
chcunningham, line 4142: nits: 

- s/might want to/may/
- proactively

index.src.html
```diff
@@ -4133,6 +4136,42 @@
...
+codecs. This is particularly true in the case where hardware codecs are limited,
+and shared accross all web pages.
+
+A User Agent is allowed to [=reclaim a codec=], freeing the resources of a
```
chcunningham, line 4146: s/is allowed to/may/

index.src.html
```diff
@@ -4133,6 +4136,42 @@
...
+<dfn>reclaimable codec</dfn>, which is a defined as a codec that is either an
+[=inactive codec=] or a [=background codec=].
+
+An <dfn>inactive codec</dfn> is a codec that has not received a call to
+`encode()` or `decode()` in the past 1 second, and has not called its
```
chcunningham, line 4151: 1 second is lower than I expected. I think our timeout in `<video>` has historically been 3 seconds?

index.src.html
chcunningham: Here and elsewhere I would refer to the existing dfn of (codec) system resources
https://w3c.github.io/webcodecs/#system-resources

index.src.html
```diff
@@ -4133,6 +4136,42 @@
...
+`output()` callback in the past 1 second.
```
chcunningham, line 4152: nit: here and elsewhere, surround numbers / literals w/ backticks to create the monospace font affect

index.src.html
```diff
@@ -4133,6 +4136,42 @@
...
+Resource Reclamation{#resource-reclamation}
+==============================================
+
+When resources are constrained, a User Agent might want to pro-actively reclaim
+codecs. This is particularly true in the case where hardware codecs are limited,
+and shared accross all web pages.
+
+A User Agent is allowed to [=reclaim a codec=], freeing the resources of a
+<dfn>reclaimable codec</dfn>, which is a defined as a codec that is either an
+[=inactive codec=] or a [=background codec=].
```
chcunningham, line 4148: After this sentence, should we explicitly ban the reclaiming of active & foreground codecs. e.g. 

`A user agent must not reclaim a codec that is not meeting one or both of these criteria.`

chcunningham (commented):
Generally looks good. A few simple questions.

index.src.html
aboba: Maybe "A User Agent may only [=reclaim a codec=] that is either an [=inactive codec=] or a [=background codec=].  An [=active codec=] or [=foreground codec=] MUST NOT be reclaimed."

tguilbert-google: I've made another pass. I am dropping the [=reclaimable codec=] term, as not all that useful.

I tried to avoid defining [=active codec=] and [=foreground codec=], since they are only used once. I'm happy to change this to add explicit definitions.

index.src.html
```diff
@@ -4133,6 +4136,43 @@
...
+A User Agent may only [=reclaim a codec=] that is either an
+[=inactive codec=], a [=background codec=], or both. A User Agent MUST NOT
+reclaim a codec that is both active and in the foreground, e.g any codec that is
+neither an [=inactive codec=] nor a [=background codec=].
+
+An <dfn>inactive codec</dfn> is a codec that has not received a call to
+`encode()` or `decode()` in the past `3 seconds`, and has not called its
+`output()` callback in the past `3 seconds`.
+
+A <dfn>background codec</dfn> is a codec whose {{ownerDocument}}'s
```
tguilbert-google, line 4155: Does {{ownerDocument}} also work for the workers case?

index.src.html
```diff
@@ -4144,6 +4147,43 @@
...
+        reclaim codecs that will alleviate the pressure of scarce resources.
+    - How disruptive might reclaiming the codec be? User Agents should consider
+        whether the codec is a [=background codec=], an [=inactive codec=], or
+        both. Codecs that are both are least likely disrupt to user experience
+        when reclaimed. Otherwise, User Agents might prefer reclaiming codecs
+        that are [=background codec=]s, before [=inactive codec=]s in
+        foreground pages.
+    - Any other heuristics, determined by the User Agent. For example,
+        heuristics might prioritize reclaiming the oldest codecs, the ones with
+        the least recent activity, or the least frequent activity.
```
padenot, line 4186: This is going to be a web compat nightmare.

index.src.html
padenot: You can refer to the document that is in the owner set of the worker if there is one, or reccurse if there is none (nested workers).

padenot (commented):
Quite a few comments on this that we need to talk about.

index.src.html
```diff
@@ -4144,6 +4147,43 @@
...
+==============================================
+
+When resources are constrained, a User Agent may proactively reclaim codecs.
+This is particularly true in the case where hardware codecs are limited, and
+shared accross all web pages.
+
+A User Agent may only [=reclaim a codec=] that is either an
+[=inactive codec=], a [=background codec=], or both. A User Agent MUST NOT
+reclaim a codec that is both active and in the foreground, e.g any codec that is
+neither an [=inactive codec=] nor a [=background codec=].
```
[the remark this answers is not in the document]
chcunningham: See reply in https://github.com/w3c/webcodecs/issues/121#issuecomment-868180014

index.src.html
```diff
@@ -4144,6 +4147,43 @@
...
+
+An <dfn>inactive codec</dfn> is a codec that has not received a call to
+`encode()` or `decode()` in the past `3 seconds`, and has not called its
+`output()` callback in the past `3 seconds`.
```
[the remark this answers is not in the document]
chcunningham: How about 10 seconds?

index.src.html
chcunningham: In my initial read of this I agreed, but then I re-read the intro and felt better.

> A User Agent may only reclaim a codec that is either an inactive codec, a background codec, or both. A User Agent MUST NOT reclaim a codec that is both active and in the foreground, e.g any codec that is neither an inactive codec nor a background codec.

This caused me to read the highlighted sentence as purely informational  description (perhaps it should be a note) of the narrow flexibility afforded by the above rule and how UAs might prioritize codecs for reclamation. For instance, the UA may choose to reclaim only one of two inactive codecs if one codec was holding onto a more precious resource than the other. The UA may also choose to reclaim both innactive codecs or neither. But this sentence does not allow the UA to reclaim an _active foreground_ codec.

padenot: This needs to be specced a bit more tightly, https://github.com/w3c/webcodecs/issues/121#issuecomment-875750242 is my latest thinking, transcribed by @chcunningham.

tguilbert-google (commented):
I've added restrictions to background reclamations. I think these new ones cover the RTC case @chcunningham mention in [this comment](https://github.com/w3c/webcodecs/issues/121#issuecomment-868180014)

I've removed all language specifying how a User Agent might want to prioritize codecs reclamation within the pool of acceptable codecs.

index.src.html
```diff
@@ -4549,40 +4549,38 @@
...
-    - How disruptive might reclaiming the codec be? User Agents should consider
-        whether the codec is a [=background codec=], an [=inactive codec=], or
-        both. Codecs that are both are least likely disrupt to user experience
-        when reclaimed. Otherwise, User Agents might prefer reclaiming codecs
-        that are [=background codec=]s, before [=inactive codec=]s in
-        foreground pages.
-    - Any other heuristics, determined by the User Agent. For example,
-        heuristics might prioritize reclaiming the oldest codecs, the ones with
-        the least recent activity, or the least frequent activity.
+A User Agent MUST only [=reclaim a codec=] that is either an
```
chcunningham, line 4561: LGTM to the substance here, but I would propose a bit of reorg for readability. As-is, my nit is that what is/isn't allowed seems defined in two places with definitions for active/background in the middle.

I would structure like this

(new paragraph) The rules governing when a codec may be reclaimed depend on whether the codec is an active or inactive codec and/or a background codec. 

An active codec is a codec that.... 

A background codec is a codec that...

A user agent must only reclaim a codec that is either an ...

Additionally, User Agents MUST NOT reclaim [=background=] codecs in the
following scenarios:...

index.src.html
```diff
@@ -4549,40 +4549,38 @@
...
+<dfn lt="inactive codec|inactive">inactive codec</dfn> is any codec that does
+not meet the definition of an [=active codec=].
+
+A <dfn lt="background codec|background">background codec</dfn> is a codec whose
+{{ownerDocument}} (or [=owner set=]'s {{Document}}, for codecs in workers) has a
+{{Document/hidden}} attribute equal to {{hidden|"hidden"}}.
+
+Additionally, User Agents MUST NOT reclaim [=background=] codecs in the
+following scenarios:
+    - [=Active=] {{AudioEncoder}}s or {{VideoEncoder}}s, to prevent long running
```
chcunningham, line 4577: here and below, I'd put the "to prevent..." bits in NOTE: blocks under the bullet. Separates the info from the requirement. Probably implies some grammar massaging too to keep the flow.

index.src.html
```diff
@@ -4549,40 +4549,38 @@
...
+        encode tasks from being interrupted.
+    - [=Active=] {{AudioDecoder}}s or {{VideoDecoder}}s, if there is
+        respectively an [=active=] {{AudioEncoder}} or {{VideoEncoder}} in the
+        same context, to prevent breaking long running transcoding tasks.
```
chcunningham, line 4581: probably, s/context/global object/ (linkable dfn). @padenot check me on this (vs using "realm")

index.src.html
```diff
@@ -4549,40 +4549,38 @@
...
+    - [=Active=] {{AudioDecoder}}s when their respective tabs is actively
```
chcunningham, line 4582: @padenot is there some better way to say "tab is actively playing audio". Does WA have this notion?

chcunningham (commented):
Gernally looks good. @padenot a few questions for you

index.src.html
tguilbert-google: Done.

index.src.html
tguilbert-google: I've added Window and WorkerGlobalScope references. Is there a shorthand for both?

index.src.html
tguilbert-google: I've done one iteration of massaging the grammar.

index.src.html
chcunningham: I think "global object" or "realm" is the shorthand you're looking for. We've previously used global object ("Each global object has a separate control thread.").

tguilbert-google: Ah! I didn't realize we were already referencing it in this CL.

index.src.html
```diff
@@ -4946,6 +4949,49 @@
 : <dfn event for=ImageTrack>change</dfn>
 :: Fired at the {{ImageTrack}} when the {{ImageTrack/frameCount}} is altered.
 
+Resource Reclamation{#resource-reclamation}
+==============================================
+
+When resources are constrained, a User Agent MAY proactively reclaim codecs.
+This is particularly true in the case where hardware codecs are limited, and
+shared accross all web pages.
```
padenot, line 4957: Not only web pages.

index.src.html
```diff
@@ -4946,6 +4949,49 @@
...
+{{DOMException}}.
+
+The rules governing when a codec may be reclaimed depend on whether the codec is
+an [=active=] or [=inactive=] codec and/or a [=background=] codec.
+
+An <dfn lt="active codec|active">active codec</dfn> is a codec that has
+received a call to `encode()` or `decode()` in the past `10 seconds`, or has
+called its `output()` callback in the past `10 seconds`. An
+<dfn lt="inactive codec|inactive">inactive codec</dfn> is any codec that does
+not meet the definition of an [=active codec=].
```
padenot, line 4971: This is good, but we need to add another condition, when a frame is being encoded: it's well possible that there is 10s between calling `encode()` and the encoded frame being output. I tested encoding a 1080p file to av1 on my rather powerful Macbook with `rav1e`, a high quality encoding is > 10s per frame.

index.src.html
padenot: Not that I know of in terms of spec. That said in Gecko we prefer to use the term "audible", because sometimes we adopt a different behaviour if a page is technically playing audio (i.e. an audio stream is running), but it's been silent for a while.

padenot (approved):
Approved with the comments fixed, thanks!

index.src.html
tguilbert-google: Added a small clarification.

index.src.html
tguilbert-google: I added a special case for VideoEncoder, and added a note.

A concern @chcunningham and I shared was "how do we reclaim stalled VideoEncoders", which is why I went with "making progress in encoding" as the definition for activity

chcunningham (approved):
LGTM. Nice work!

PR #306, opened 2021-07-16
Define video color space interfaces and members

### LINKED ISSUE #47: Negotiating colorspaces

chcunningham:
Fixes #47

:)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/pull/306.html" title="Last updated on Aug 11, 2021, 5:54 AM UTC (41253c5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/306/b43d3df...41253c5.html" title="Last updated on Aug 11, 2021, 5:54 AM UTC (41253c5)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -1255,20 +1255,20 @@ Methods {#videoencoder-methods}
     5. [=Queue a control message=] to encode |frameClone|.
 
     [=Running a control message=] to encode the frame means performing these steps.
-    1. If {{VideoEncoder/[[active color space]]}} is `null`:
-        1. If |frame|.{{VideoFrame/colorSpace}} is not `null`, assign its value
-            to {{VideoEncoder/[[active color space]]}}.
-        2. ISSUE: what here? 709? Can we trust that frame.colorSpace is not null?
-    1. Attempt to use {{VideoEncoder/[[codec implementation]]}} to encode
+    1. If {{VideoEncoder/[[active color space]]}} is `null`, and
+        |frame|.{{VideoFrame/colorSpace}} is not `null`, assign
+        |frame|.{{VideoFrame/colorSpace}} to
+        {{VideoEncoder/[[active color space]]}}.
+    2. Attempt to use {{VideoEncoder/[[codec implementation]]}} to encode
         |frameClone| according to |options|.
-    2. If encoding results in an error, queue a task on the [=control thread=]
+    3. If encoding results in an error, queue a task on the [=control thread=]
         event loop to run the [=Close VideoEncoder=] algorithm with
         {{EncodingError}}.
-    3. Queue a task on the [=control thread=] event loop to decrement
+    4. Queue a task on the [=control thread=] event loop to decrement
         {{VideoEncoder/[[encodeQueueSize]]}}.
-    4. Let |encoded outputs| be a [=list=] of encoded video data outputs
+    5. Let |encoded outputs| be a [=list=] of encoded video data outputs
         emitted by {{VideoEncoder/[[codec implementation]]}}.
-    5. If |encoded outputs| is not empty, queue a task on the [=control thread=]
+    6. If |encoded outputs| is not empty, queue a task on the [=control thread=]
         event loop to run the [=Output EncodedVideoChunks=] algorithm with
         |encoded outputs|.
   </dd>
--- a/index.src.html
+++ b/index.src.html
@@ -1255,8 +1255,7 @@ Methods {#videoencoder-methods}
     5. [=Queue a control message=] to encode |frameClone|.
 
     [=Running a control message=] to encode the frame means performing these steps.
-    1. If {{VideoEncoder/[[active color space]]}} is `null`, and
-        |frame|.{{VideoFrame/colorSpace}} is not `null`, assign
+    1. If {{VideoEncoder/[[active color space]]}} is `null`, assign
         |frame|.{{VideoFrame/colorSpace}} to
         {{VideoEncoder/[[active color space]]}}.
     2. Attempt to use {{VideoEncoder/[[codec implementation]]}} to encode
--- a/index.src.html
+++ b/index.src.html
@@ -1274,6 +1274,13 @@ Methods {#videoencoder-methods}
     1. If {{VideoEncoder/[[active color space]]}} is `null`, assign
         |frame|.{{VideoFrame/colorSpace}} to
         {{VideoEncoder/[[active color space]]}}.
+
+        NOTE: For codec implementations that encode in-band color space metdata,
+                the above step implies that User Agents defer running the
+                [=control message steps=] associated with the most recent call
+                to {{VideoEncoder/configure()}} until the arrival of the first
+                frame to {{VideoEncoder/encode()}}.
+
     2. Attempt to use {{VideoEncoder/[[codec implementation]]}} to encode
         |frameClone| according to |options|.
     3. If encoding results in an error, queue a task on the [=control thread=]
@@ -3103,8 +3110,8 @@ dictionary VideoFrameBufferInit {
         {{VideoFrameBufferInit/duration}} to {{VideoFrame/[[timestamp]]}} and
         {{VideoFrame/[[duration]]}} respectively.
     7. Let |colorSpace| be `undefined`.
-    8. If {{VideoFrameBufferInit/colorSpace}} [=map/exists=], assign its value
-        to |colorSpace|.
+    8. If |init|.{{VideoFrameBufferInit/colorSpace}} [=map/exists=], assign its
+        value to |colorSpace|.
     9. Assign the result of running the [=VideoFrame/Pick Color Space=]
         algorithm, with |colorSpace| and {{VideoFrame/[[format]]}}, to
         {{VideoFrame/[[color space]]}}.
--- a/index.src.html
+++ b/index.src.html
@@ -163,7 +163,7 @@ Definitions {#definitions}
     track. The mechanism for indicating a primary track is format defined.
 
 : <dfn>RGB Format</dfn>
-:: A {{PixelFormat}} containing red, green, and blue color channels in
+:: A {{VideoPixelFormat}} containing red, green, and blue color channels in
     any order or layout (interleaved or planar), and irrespective of whether an
     alpha channel is present.
 
--- a/index.src.html
+++ b/index.src.html
@@ -1631,6 +1631,9 @@ dictionary VideoDecoderConfig {
   [EnforceRange] unsigned long displayAspectWidth;
   [EnforceRange] unsigned long displayAspectHeight;
   HardwareAcceleration hardwareAcceleration = "allow";
+
+  // Hint about the color space of the content.
+  VideoColorSpaceInit colorSpace;
 };
 </xmp>
 
@@ -1755,6 +1758,9 @@ dictionary VideoEncoderConfig {
   DOMString scalabilityMode;
   BitrateMode bitrateMode = "variable";
   LatencyMode latencyMode = "quality";
+
+  // Color space for metadata (in-band or output VideoDecoderConfig).
+  VideoColorSpaceInit colorSpace;
 };
 </xmp>
 
@@ -2768,6 +2774,10 @@ interface VideoFrame {
   readonly attribute unsigned long displayHeight;
   readonly attribute unsigned long long? duration;  // microseconds
   readonly attribute long long? timestamp;          // microseconds
+
+  // Color space of the data produced by copyTo().
+  readonly attribute VideoColorSpace? colorSpace;
+
   unsigned long allocationSize(
       optional VideoFrameCopyToOptions options = {});
   Promise<sequence<PlaneLayout>> copyTo(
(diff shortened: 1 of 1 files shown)


chcunningham:
For now it's just the basic IDL. Out of time tonight, but wanted to get this out for discussion tomorrow. The enums are just a start - more to come there (probably in a follow up) aligning with H.273.

@sandersdan FYI

chcunningham:
@sandersdan - looking at `VideoColorSpace.toJSON()` - can you remind me what the precedent was for this?

For changing color space (getting dict version of the interface), @padenot mentioned the pattern in WebRTC has been to use Object.assign(), for example 
image

sandersdan:
`Object.assign()` is very similar to `[Default] toJSON()`, but it doesn't work as (may be) intended if the interface includes properties that shouldn't be turned into fields. For `VideoColorSpace` this could come up if we add H.273 getters, for example.

`toJSON()` also allows `JSON.stringify()` to work.

index.src.html
```diff
@@ -2685,6 +2695,9 @@
   // Default display dimensions match visibleRect.
   [EnforceRange] unsigned long displayWidth;
   [EnforceRange] unsigned long displayHeight;
+
+  // Defaults to Rec709 or SRGB for planar and interleaved formats respectively.
```
sandersdan, line 2699: For YUV and RGB formats respectively.

index.src.html
```diff
@@ -3672,6 +3685,50 @@
...
+interface VideoColorSpace {
+  constructor(VideoColorSpaceInit init);
+
+  readonly attribute VideoColorPrimaries? primaries;
+  readonly attribute VideoTransferCharacteristics? transfer;
+  readonly attribute VideoMatrixCoefficients? matrix;
+  readonly attribute boolean? fullRange;
+
+  // Note: [Default] implementation would be correct.
+  VideoColorSpaceInit toJSON();
```
sandersdan, line 3702: Should be [Default] in the spec (that's what the note is about).

index.src.html
chcunningham: I've deleted this comment now, but see how I've implemented the "Pick Color Space" algo (choose sRGB if it's an RGB format, otherwise choose rec709).

index.src.html

index.src.html
```diff
@@ -1216,6 +1253,10 @@
     5. [=Queue a control message=] to encode |frameClone|.
 
     [=Running a control message=] to encode the frame means performing these steps.
+    1. If {{VideoEncoder/[[active color space]]}} is `null`:
+        1. If |frame|.{{VideoFrame/colorSpace}} is not `null`, assign it's value
+            to {{VideoEncoder/[[active color space]]}}.
+        2. ISSUE: what here? 709? Can we trust that frame.colorSpace is not null?
```
chcunningham, line 1259: @sandersdan thoughts? I we discussed it being null in some cases, like if you were creating from an internal videoframe that had some non-compatible representation? I wasn't able to quickly find cases in our impl where this actually occurs.

index.src.html
```diff
@@ -1216,6 +1253,10 @@
     5. [=Queue a control message=] to encode |frameClone|.
 
     [=Running a control message=] to encode the frame means performing these steps.
+    1. If {{VideoEncoder/[[active color space]]}} is `null`:
```
chcunningham, line 1256: This is different from what we implemented. We currently default to 709 if the user doesn't give a config.colorSpace. But that seemed needlessly unfriendly. Maybe I'm forgetting some good reason we did that.

index.src.html
```diff
@@ -3148,6 +3246,14 @@
         |frame|.{{VideoFrame/duration}}.
     8. Assign `init`.{{VideoFrameInit/timestamp}} to
         |frame|.{{VideoFrame/timestamp}}.
+    9. If |resource| has a known {{VideoColorSpace}}, assign it's value to
+        {{VideoFrame/[[color space]]}}.
+    10. Otherwise, assign {{VideoFrame/[[color space]]}} with the result of
+        running the [=VideoFrame/Pick Color Space=] algorithm, with
+        {{VideoFrame/[[format]]}}.
+
+        ISSUE: This implies frame.colorSpace would never be null. Should the
```
chcunningham, line 3255: More of the same question.

chcunningham (commented):
Mostly done. Just needs definitions for the enums. @sandersdan a few questions for you in here.

index.src.html
sandersdan: That's correct, an implementation could produce a VideoFrame with no JS-exposed color space. In Chrome that can happen easily with the current implementation because the WebCodecs implementation only supports a subset of color spaces (BT. 601, BT. 709, and sRGB), but if/when it's expanded to all of H.273 it would only happen for frames that are in a non-H.273 color space (currently not possible in Chrome).

index.src.html
sandersdan: I'm not fully convinced that we should take the color space from the frame, since it may force configuration to be deferred internally. I'm not sure what a normal encoding pipeline will look like, though; will apps prefer to do color conversion on the client and send consistent streams, or will they prefer to configure the streams to match the source and handle that on the server/pass through to receiving clients?

Given that apps can override the behavior by explicitly setting a color space, the risk is small.

index.src.html
```diff
@@ -732,10 +750,24 @@
...
+
+            NOTE: The codec implementation may detect a {{VideoColorSpace}} by
+                analyzing the bitstream. Detection is not required and may not
+                always be feasible. The exact method of detection is implementer
+                defined and codec-specific. Authors may override the
+                detected {{VideoColorSpace}} by providing a
+                {{VideoDecoderConfig/colorSpace}} in the {{VideoDecoderConfig}}.
+
+        5. If {{VideoDecoderConfig/colorSpace}} [=map/exists=] in the
+            {{VideoDecoder/[[active decoder config]]}}, assign it's value to
```
sandersdan, line 765: its

index.src.html
```diff
@@ -3148,6 +3246,14 @@
         |frame|.{{VideoFrame/duration}}.
     8. Assign `init`.{{VideoFrameInit/timestamp}} to
         |frame|.{{VideoFrame/timestamp}}.
+    9. If |resource| has a known {{VideoColorSpace}}, assign it's value to
```
sandersdan, line 3249: its

index.src.html
chcunningham: On VC just now we observed
- I anticipate most encoding pipelines will just want pass through of sender -> receiver without any color space normalization. With that expectation, in the absence of a config.colorSpace, it's best to trust the frame vs picking some default.
- Deferred configuration is fine (something we wanted to allow by design anyway), but good to note. I'll do that.

sandersdan:
Merge conflicts resolved, minor fixes applied.

@padenot PTAL.

index.src.html
```diff
@@ -1218,16 +1255,19 @@
     5. [=Queue a control message=] to encode |frameClone|.
 
     [=Running a control message=] to encode the frame means performing these steps.
-    1. Attempt to use {{VideoEncoder/[[codec implementation]]}} to encode
+    1. If {{VideoEncoder/[[active color space]]}} is `null`, assign
+        |frame|.{{VideoFrame/colorSpace}} to
```
sandersdan, line 1259: Adding back Chris' concern here: `VideoFrame.colorSpace` could be `null`, although this should only happens in cases where the UA produced the frame itself and wasn't able to map to a WebCodecs color space. Potentially we should remove the nullability here, since UAs could produce a `VideoColorSpace` with null members instead.

index.src.html
```diff
@@ -2768,6 +2829,8 @@
   readonly attribute unsigned long displayHeight;
   readonly attribute unsigned long long? duration;  // microseconds
   readonly attribute long long? timestamp;          // microseconds
+  readonly attribute VideoColorSpace? colorSpace;
```
sandersdan, line 2832: Review note: We could make this non-nullable and use a `VideoColorSpace` instance with null values instead.

chcunningham:
> Review note: We could make this non-nullable and use a VideoColorSpace instance with null values instead.

I think I prefer the parent nullable, as-is, but I'm flexible. 

Sorry, ran out time here. @sandersdan, here's an outline of todos:

1. Define the attribute getter steps (and associated slots) for the VideoColorSpace interface. Should be similar in size to https://w3c.github.io/webcodecs/#imagetrack 

2. For "Initialize Frame With Resource and Size (with init, frame, resource, width and height)", delete step 12, instead write: `Otherwise, assign `null` to [[color space]].` Per our convo that not every image source can be represented. Also, delete the `Issue:` block following this step

3. Fill out "Rendering Color" section under VideoColorSpace, replacing the placeholder Issue: block. The idea here is just to describe in english (not rigid numbered steps) the behavior of WebGL and createImageBitmap() wrt VideoFrame. 

4. Add brief definitions for VideoColorSpace enum values, referencing table entries in h.273 (let h.273 do the heavy lifting here).

sandersdan:
> Sorry, ran out time here. @sandersdan, here's an outline of todos:

TODOs fixed, PTAL.

index.src.html
```diff
@@ -734,10 +763,24 @@
...
             |displayAspectWidth| and |displayAspectHeight| respectively.
-        4. Let |frame| be the result of running the [=Create a VideoFrame=]
-            algorithm with |output|, |timestamp|, |duration|, |displayAspectWidth|
-            and |displayAspectHeight|.
-        5. Invoke {{VideoDecoder/[[output callback]]}} with |frame|.
+        4. Let |colorSpace| be the {{VideoColorSpace}} for |output| as detected
+            by the codec implementation. If no {{VideoColorSpace}} is detected,
+            let |colorSpace| be `undefined`.
+
+            NOTE: The codec implementation may detect a {{VideoColorSpace}} by
```
padenot, line 770: Please rephrase to remove rfc2119 word in a non-normative block.

index.src.html
```diff
@@ -1879,6 +1942,15 @@
...
     Configures latency related behaviors for this codec. See {{LatencyMode}}.
   </dd>
+
+  <dt><dfn dict-member for=VideoEncoderConfig>colorSpace</dfn></dt>
+  <dd>
+    Configures the {{VideoColorSpace}} to be used in the output
+    {{VideoDecoderConfig}}. May also be used to set in-band color information
+    while encoding. If not provided, the codec will attempt to use the
+    {{VideoFrame}}.{{VideoFrame/colorSpace}} for the first frame passed to
+    {{VideoEncoder/encode()}} following this {{VideoEncoderConfig}}.
```
padenot, line 1952: What happens if this attempt fail ?

index.src.html
```diff
@@ -3192,40 +3283,56 @@
...
+    2. Return |frame|.
+
+
+: <dfn for=VideoFrame>Pick Color Space</dfn> (with |overrideColorSpace| and |format|)
+:: 1. If |overrideColorSpace| is provided, return a new {{VideoColorSpace}}
+        constructed with |overrideColorSpace|.
+
+        NOTE: User Agents may replace `null` members of the provided
+            |overrideColorSpace| with  guessed values as determined by
+            implementer defined hueristics.
```
padenot, line 3325: `s/hueristics/heuristics/`.

That smells like a compatibility problem, can we not spell out some rules here? Isn't this likely to be a simple extrapolation from a field to the obvious others, as defined around line 181 ?

index.src.html
```diff
@@ -3488,6 +3598,16 @@
         field to the {{VideoFrame}} internal slot in |value| with the same name
         as the named field.
 
+### Rendering ### {#videoframe-rendering}
+
+When rendered, for example by {{CanvasDrawImage}}
+{{CanvasDrawImage/drawImage()}}, a {{VideoFrame}} should be converted to a
```
padenot, line 3604: `MUST` ? In what scenario would an implementation decide not to convert ?

index.src.html
sandersdan: Nothing in particular, following the steps the the output `colorSpace` value will be `null`.

index.src.html
sandersdan: In most cases these should be obvious extrapolations, yes. I'm not sure I have the expertise to prescribe a process though, and limiting the implementation to changing empty fields allows apps to be specific and avoid implementation-defined behavior if desired.

chcunningham (commented):
> TODOs fixed, PTAL.

Thank you! LGTM. 

I added a small note to address the [comment from earlier](https://github.com/w3c/webcodecs/pull/306#:~:text=chcunningham-,15%20days%20ago,-Author). 

 @padenot @aboba let us know if this looks good.

padenot (approved):

chcunningham:
EC: @aboba defers to Paul. LGTM

PR #307, opened 2021-07-16
Update AudioSampleFormat

### LINKED ISSUE #300: Should AudioSampleFormat enums values be lowercase?
Follow up to #256, brought up by @sandersdan.

It's standard for video pixel formats to be uppercase, but audio sample formats don't have such a strong convention.  JS enums are more commonly seen as lower case. @padenot, was there a specific intention with going to uppercase?

Any thoughts on using `u8p` or `u8P` instead of `U8P`?

tguilbert-google:
Fixes #300

- Uses lowercase for all AudioSampleFormat enums
- Uses `"-planar"` instead of `"p"`
- Removes `"s24"` and `"s24-planar"` formats.
- Adds a note recommending left-shift by 8 for converting from 24-bit to `"s32"`

***
<a href="https://pr-preview.s3.amazonaws.com/tguilbert-google/webcodecs/pull/307.html" title="Last updated on Jul 28, 2021, 12:27 AM UTC (011f41a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/307/2cbb1e2...tguilbert-google:011f41a.html" title="Last updated on Jul 28, 2021, 12:27 AM UTC (011f41a)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -2507,8 +2507,8 @@ An audio sample format describes the numeric type used to represent a
 single sample (e.g. 32-bit floating point) and the arrangement of samples from
 different channels as either [=interleaved=] or [=planar=]. The <dfn>audio
 sample type</dfn> refers solely to the numeric type and interval used to store
-the data, this is {{U8}}, {{S16}}, {{S24}}, {{S32}}, or {{FLT}} for respectively
-unsigned 8-bits, signed 16-bits, signed 32-bits, signed 32-bits, and 32-bits
+the data, this is {{u8}}, {{s16}}, {{s32}}, or {{f32}} for respectively
+unsigned 8-bits, signed 16-bits, signed 32-bits, and 32-bits
 floating point number. The [[#audio-buffer-arrangement|audio buffer
 arrangement]] refers solely to the way the samples are laid out in memory
 ([=planar=] or [=interleaved=]).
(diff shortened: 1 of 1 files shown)


chcunningham (approved):
LGTM. @padenot, PTAL. Particularly, double check the details re: 24-bit formats.

aboba (approved):

index.src.html
```diff
@@ -2665,6 +2649,10 @@
 but audio content using 24-bit samples is common, so 32-bits integers are
 commonly used to hold 24-bit content.
 
+{{AudioDecoder}}s decoding 24-bit audio chunks MUST convert their output to
+{{s32}} or {{s32-planar}}, by left-shifting the decoded samples by `8` before
```
padenot, line 2653: Add a note that explains what happens when the audio clips.

Mentions that we're talking about a bit shift on each sample: we're specifically not shifting the decoded samples by 8, we're bit-shifting each sample by 8.

Add a note that explain why this is OK, and the trade-off: any audio that clips cannot be decoded losslessly. Mention that reading back to f32 doesn't clip, and that authors that want this to work want to read-back to float. Mention that implementations SHOULD take care to not clip internally (we can discuss the normative requirement, of course).

dalecurtis: Taking over for @tguilbert-google since he's OOO this week.

The current text indicates that _AudioDecoders_ should always convert 24-bit audio to 32-bit audio via bit-shifting, so even if copyTo() is called later with float32, the contents will already have been clipped.

I've updated the text to be more inline with your thoughts. PTAL.

padenot (approved):

Issue #239, opened 2021-05-11
Is exposing https://w3c.github.io/webcodecs/#enumdef-hardwareacceleration a good idea


youennf:
I am wondering what HardwareAcceleration is supposed to be used for.

One potential use would be to always prefer power efficiency. But power efficiency does not mandate hardware acceleration.
Depending on the device, the codec, the resolution, software based codecs might be better suited. It is unclear how a web developer will be able to select hardwareAcceleration for that case except to let the UA decide with 'allow'.

Another possibility is to maximise compatibility and use 'deny'. In that case though, it means that web developer looses power efficiency in a lot of cases. A careful web developer will then probably want to enter the business of identifying which SW and HW codecs are in use on a device. This does not seem great and somehow contradicts the desire to not increase fingerprinting.

It seems UA is in general the best entity to decide what to use at any given point.
Instead of hard requirements, a web application could look at providing hints, though it is true hints tend to be difficult to define and implement consistently.

It also seems HardwareAcceleration is a potential fingerprinting vector though it is not marked as so in the spec.

chcunningham:
> I am wondering what HardwareAcceleration is supposed to be used for.

This is a feature request from several sophisticated apps that bring their own encoders/decoders implemented in WASM (with customized features and tuning). Such apps are interested in WebCodecs only when we can offer hardware acceleration. If we can only do software encoding/decoding, they prefer their WASM codec 

> It is unclear how a web developer will be able to select hardwareAcceleration for that case except to let the UA decide with 'allow'.

What part is unclear? `allow` works as you've described, and is the default value. 

> This does not seem great and somehow contradicts the desire to not increase fingerprinting.

Fingerprinting and codec features are often add odds. In such cases, including this one, we strive to offer the features folks are demanding without offering any extra information (leas power principal). This is why we have reduced the signal to "hardware accelerated", as opposed to exposing the name of the hardware or similar. 

> It also seems HardwareAcceleration is a potential fingerprinting vector though it is not marked as so in the spec.

Agree, I think that's a good thing for us to call out.

chcunningham:
Triage note: marking 'breaking', as removal of this attribute would clearly break behavior for folks that have come to rely on it. This is purely a triage note. My stance on implementing this change (opposed) is still as described above. 

Also marking 'editorial' for the requested additions to privacy considerations.

chcunningham:
@youennf, pls see above. If nothing more to discuss I'd like to close.

youennf:
[...]

PR #313, opened 2021-07-28
Revise hardwareAcceleration based on consensus decision.

dalecurtis:
* It's now a hint instead of being required.
* Values are noPreference, preferSoftware, and preferHardware.

Fixes: #239

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/pull/313.html" title="Last updated on Aug 16, 2021, 11:55 PM UTC (4ea4dc6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/313/79a1416...4ea4dc6.html" title="Last updated on Aug 16, 2021, 11:55 PM UTC (4ea4dc6)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -1740,9 +1740,7 @@ run these steps:
     return `false`.
 5. If {{VideoDecoderConfig/displayAspectWidth}} = 0 or
     {{VideoDecoderConfig/displayAspectHeight}} = 0, return `false`.
-6. If {{VideoDecoderConfig/hardwareAcceleration}} is provided and its value is
-    unsupported by the User Agent, return `false`.
-7. Return `true`.
+6. Return `true`.
 
 <dl>
   <dt><dfn dict-member for=VideoDecoderConfig>codec</dfn></dt>
@@ -1883,9 +1881,7 @@ To check if a {{VideoEncoderConfig}} is a <dfn>valid VideoEncoderConfig</dfn>,
     = 0, return `false`.
 3. If {{VideoEncoderConfig/displayWidth}} = 0 or
     {{VideoEncoderConfig/displayHeight}} = 0, return `false`.
-4. If {{VideoEncoderConfig/hardwareAcceleration}} is provided and its value is
-    unsupported by the User Agent, return `false`.
-5. Return `true`.
+4. Return `true`.
 
 <dl>
   <dt><dfn dict-member for=VideoEncoderConfig>codec</dfn></dt>
--- a/index.src.html
+++ b/index.src.html
@@ -1649,7 +1649,9 @@ run these steps:
     return `false`.
 5. If {{VideoDecoderConfig/displayAspectWidth}} = 0 or
     {{VideoDecoderConfig/displayAspectHeight}} = 0, return `false`.
-6. Return `true`.
+6. If {{VideoDecoderConfig/hardwareAcceleration}} is provided and its value is
+    unsupported by the User Agent, return `false`.
+7. Return `true`.
 
 <dl>
   <dt><dfn dict-member for=VideoDecoderConfig>codec</dfn></dt>
@@ -1781,7 +1783,9 @@ To check if a {{VideoEncoderConfig}} is a <dfn>valid VideoEncoderConfig</dfn>,
     = 0, return `false`.
 3. If {{VideoEncoderConfig/displayWidth}} = 0 or
     {{VideoEncoderConfig/displayHeight}} = 0, return `false`.
-4. Return `true`.
+4. If {{VideoEncoderConfig/hardwareAcceleration}} is provided and its value is
+    unsupported by the User Agent, return `false`.
+5. Return `true`.
 
 <dl>
   <dt><dfn dict-member for=VideoEncoderConfig>codec</dfn></dt>
--- a/index.src.html
+++ b/index.src.html
@@ -1901,7 +1901,7 @@ specialized hardware. {{HardwareAcceleration/prefer-hardware}} and
 respect these values when possible, User Agents may ignore these values in some
 or all circumstances for any reason.
 
-To avoid fingerprinting issues, User Agents SHOULD ensure rejection or
+To prevent fingerprinting issues, User Agents MUST ensure rejection or
 acceptance of a given {{HardwareAcceleration}} preference reveals no additional
 information on top of what is revealed by [[media-capabilities]] or other web
 exposed capability systems the User Agent may provide.
--- a/index.src.html
+++ b/index.src.html
@@ -1901,6 +1901,11 @@ specialized hardware. {{HardwareAcceleration/prefer-hardware}} and
 respect these values when possible, User Agents may ignore these values in some
 or all circumstances for any reason.
 
+To avoid fingerprinting issues, User Agents SHOULD ensure rejection or
+acceptance of a given {{HardwareAcceleration}} preference reveals no additional
+information on top of what is revealed by [[media-capabilities]] or other web
+exposed capability systems the User Agent may provide.
+
 <div class='note'>
   NOTE: Good examples of when a User Agent may ignore
   {{HardwareAcceleration/prefer-hardware}} or
--- a/index.src.html
+++ b/index.src.html
@@ -2024,7 +2024,7 @@ fingerprinting, they SHOULD ignore the {{HardwareAcceleration}} preference.
   circumstances where the User Agent determines an alternative setting would
   better serve the end user.
 
-  NOTE: Most authors will be best served by using the default of
+  Most authors will be best served by using the default of
   {{HardwareAcceleration/no-preference}}. This gives the User Agent flexibility
   to optimize based on its knowledge of the system and configuration. A common
   strategy will be to prioritize hardware acceleration at higher resolutions
--- a/index.src.html
+++ b/index.src.html
@@ -1630,7 +1630,7 @@ dictionary VideoDecoderConfig {
   [EnforceRange] unsigned long codedHeight;
   [EnforceRange] unsigned long displayAspectWidth;
   [EnforceRange] unsigned long displayAspectHeight;
-  HardwareAcceleration hardwareAcceleration = "noPreference";
+  HardwareAcceleration hardwareAcceleration = "no-preference";
   boolean optimizeForLatency;
 };
 </xmp>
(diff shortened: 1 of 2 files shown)


index.src.html
```diff
@@ -1885,30 +1885,41 @@
 ---------------------------------------------
 <xmp class='idl'>
 enum HardwareAcceleration {
-  "allow",
-  "deny",
-  "require",
+  "noPreference",
+  "preferHardware",
+  "preferSoftware",
```
sandersdan, line 1890: Per https://heycam.github.io/webidl/#idl-enums, recommend "no-preference", "prefer-hardware", "prefer-software".

sandersdan (approved):

padenot (commented):
We need to have a bit of prose somewhere that does the rejection in the case the UA decides to reject, right? I thought we had something in the current text but I can't find anything.

Probably just a rather vague line in the [valid `VideoDecoderConfig`](https://w3c.github.io/webcodecs/#valid-videodecoderconfig) steps or something.

dalecurtis:
> We need to have a bit of prose somewhere that does the rejection in the case the UA decides to reject, right? I thought we had something in the current text but I can't find anything.
> 
> Probably just a rather vague line in the [valid `VideoDecoderConfig`](https://w3c.github.io/webcodecs/#valid-videodecoderconfig) steps or something.

Yeah I thought we had one as well. Done.

index.src.html
```diff
@@ -1885,30 +1889,41 @@
 ---------------------------------------------
 <xmp class='idl'>
 enum HardwareAcceleration {
```
youennf, line 1891: Given we might introduce in the future other values (prefer-compression-efficiency on encoder side maybe), I would tend to rename the enum and the config field name itself to something like codecPreference.

index.src.html
```diff
@@ -1885,30 +1889,41 @@
...
-  "allow",
-  "deny",
-  "require",
+  "no-preference",
```
youennf, line 1892: As discussed in https://github.com/w3c/webcodecs/issues/239, I tend to prefer a more descriptive default value.
@dalecurtis mentionned renaming no-preference to prefer-efficiency.

index.src.html
```diff
@@ -1920,33 +1935,36 @@
...
   </dd>
-  <dt><dfn enum-value for=HardwareAcceleration>require</dfn></dt>
+  <dt><dfn enum-value for=HardwareAcceleration>prefer-hardware</dfn></dt>
   <dd>
-    Indicates that the User Agent must use hardware acceleration.
+    Indicates that the User Agent SHOULD prefer hardware acceleration. User
+    Agents may ignore this value for any reason.
 
-    NOTE: This will cause the configuration to be unsupported on platforms where
+    NOTE: This may cause the configuration to be unsupported on platforms where
```
youennf, line 1967: So this leaves implementations with the ability to mark some configurations as 'unsupported' because of the hint?
This seems to contradict the desire to not allow fingerprinting.
This might create compatibility issues between privacy-aware implementations and privacy-unaware implementations.
I think we should be more restrictive here to ensure leaking fingerprinting is not an option and keep good web compatibility.

index.src.html
dalecurtis: I defer final decision to @aboba @padenot, I prefer `no-preference`.

index.src.html
dalecurtis: A few points:
* We must have some sort of configuration rejection mechanism. Your argument applies to every field in the configuration dictionary even if we drop this hunk of text for `hardwareAcceleration`, so it seems impractical. I think this text strikes the best balance we can without becoming prescriptive in codecs (due to registry, literally impossible), resolutions, etc. 
* Cross-UA compatibility isn't a concern in the context of this property, since authors setting `prefer-hardware` or `prefer-software` will already have to be prepared for rejection for many reasons: lack of support, too many codecs in use, gpu disabled, resolution too small or too large, etc,
* You agreed to this here: https://github.com/w3c/webcodecs/issues/239#issuecomment-876667296 and that was what was presented to the WG, who have deferred to the editors consensus.

youennf: > * Your argument applies to every field in the configuration dictionary even if we drop this hunk of text for `hardwareAcceleration`, so it seems impractical.

The hint is there to sort the codec list, not to reduce the codec list.
As such, the hint should not have allow to change the outcome of supported or not, it should only allow to change the codec behavior if the config is supported.
If a script is able to test the same configuration with two different preferences and in one case, get an unsupported error, we are back to fingerprinting.

As for other configuration field, my understanding is that Media Capabilities should expose the same information.
Or the values make it not supported.
In case this cannot be supported and MC does not disclose that information, WebCodecs is leaking fingerprinting information. We should then do the usual fingerprinting analysis: list the mechanism, see whether it can be avoided...

Looking at video decoder as an example, only optimizeForLatency seems new from Media Capabilities.
It is a hint, and I do not see that it may trigger unsupported configuration.

> * You agreed to this here

I read too quickly the summary, I am sorry to realize that now.
In the last meeting I attended, @chcunningham and I talked about that exact possibility and I expressed my opposition.

> * Cross-UA compatibility isn't a concern in the context of this property

I disagree, cross-UA compatibility usually derives from specs that are not tight enough, which is the case here.

dalecurtis: > The hint is there to sort the codec list, not to reduce the codec list.

The codec list (keep in mind the configuration is manifold over profiles and resolutions) is already variable in size within the a context of a single-UA regardless of this property, hence my second point. There are no cross-UA compatibility issues that are not also single-UA issues an author would need to be aware of the take advantage of this property.

> In case this cannot be supported and MC does not disclose that information, WebCodecs is leaking fingerprinting information. 

There's no difference in what a given UA reveals between MediaCapabilities and WebCodecs now that this is a hint. Just as Media Capabilities does not define what codecs, resolutions, profiles, or implementations are supported (indeed the very point of the API is to answer that question!) and what it means to be power efficient, so to does the text here allow UA flexibility in making those determinations .

> I disagree, cross-UA compatibility usually derives from specs that are not tight enough, which is the case here.

I think you've misunderstood my comment. While your statement is true in some contexts, it is not true in the context of capabilities that are _standardized_ to vary across UAs and platforms like codecs.

dalecurtis: I've added a normative note in hope that will resolve your concerns:

"To avoid fingerprinting issues, User Agents SHOULD ensure rejection or
acceptance of a given {{HardwareAcceleration}} preference reveals no additional
information on top of what is revealed by [[media-capabilities]] or other web
exposed capability systems the User Agent may provide."

I'm happy to change that to a MUST if @padenot @aboba want.

index.src.html
aboba: no-preference is fine. I've also seen "balanced" used.

dalecurtis:
Bump @youennf @padenot . Thanks!

youennf:
I am fine with a MUST. I would try to be more straightforward in the way to state it.
Something like: To prevent any fingerprinting issue, the hardwareAccelaration preference value MUST have no effect on rejection of a given configuration.

dalecurtis:
@youennf the current text is accurate, I've changed the wording slightly (avoid => prevent, SHOULD => MUST), but per editors consensus rejection using the preference will be allowed if it doesn't expose any fingerprinting information.

youennf:
> rejection using the preference will be allowed if it doesn't expose any fingerprinting information.

Can you describe what is the advantage of the current PR wording?
For instance, what are the cases where using the preference would trigger/prevent rejection and would not expose fingerprinting information?

dalecurtis:
We've discussed this extensively on the issue, but to recap: the advantage is a clearer articulation of capabilities to developers (ideally resulting in a better end user experience) for UAs which have exposed equivalent information through Media Capabilities or other mechanisms.
 * Per the issue thread this is important to developers -- both Zoom and VLC have expressed a desire to manage deny-lists for robustness and/or select their own internal software codecs when necessary.
 * Data from @padenot and Zoom supports those requests. @padenot indicated that WASM is performant enough for low complexity operations that Firefox is replacing significant internal pieces with it. Zoom has confirmed independently that WASM is as performant or better for low resolutions (<=360p) using H264.

As for cases where it won't reveal additional fingerprinting, that's going to depend on the UA's implementation of other APIs. If a UA doesn't expose Media Capabilities or doesn't reveal hardware acceleration through the `powerEfficient` bit (or some other mechanism), a UA implementing WebCodecs will need to ignore the `hardwareAcceleration` value entirely.

For example, in Chromium software decoders are known based on major version, so no additional information is revealed with the `no-preference` or `prefer-software` values. MediaCapabilities reveals hardware presence via the `powerEfficient` bit for both encoding and decoding, so `IsConfigSupported()` provides zero additional information in our implementation. I expect this to be similar for most UAs, but if not they are required by this text to ensure zero impact on fingerprinting.

youennf:
> MediaCapabilities reveals hardware presence via the `powerEfficient` bit

In that case, let's be specific, being open-ended makes the whole sentence weak, I believe we can safely remove 'or other web exposed capability systems the User Agent may provide'.
FWIW, the probability for another capabilities mechanism is probably low and it is safer to revisit the issue if such a new mechanism happens.

dalecurtis:
Done, I've made the text clearer.

chcunningham:
@padenot @aboba friendly ping to take a look at the latest here

aboba (approved):

index.src.html
padenot: I find this rather descriptive. In passing this value, the author explicitly signals that they don't have any particular preference, and defers to the UA to make the decision on what decoder to use, based on the scenario characteristics (on battery power, video resolution, etc.).

padenot (approved):

chcunningham:
> > We need to have a bit of prose somewhere that does the rejection in the case the UA decides to reject, right? I thought we had something in the current text but I can't find anything.
> > Probably just a rather vague line in the [valid `VideoDecoderConfig`](https://w3c.github.io/webcodecs/#valid-videodecoderconfig) steps or something.
> 
> Yeah I thought we had one as well. Done.

This is covered by step 3 of the Check Configuration Support algo.
https://www.w3.org/TR/webcodecs/#config-support  

```
3. If the User Agent can provide a <a>codec</a> to support all entries of the
    |config|, including applicable default values for keys that are not
    included, return `true`.
```

It intentionally doesn't reference individual attributes of the config, since that would be a long list and the current text is pretty clear. This algo is invoked for both [IsConfigSupported()](https://www.w3.org/TR/webcodecs/#ref-for-check-configuration-support%E2%91%A2) and for the [support checks that take place during configure()](https://www.w3.org/TR/webcodecs/#ref-for-check-configuration-support%E2%91%A1).

I'll take the liberty of reverting @dalecurtis commit to check this in the config validation steps. The semantics of that algo are "is your input valid" which we want to distinguish from "supported". This algo is called earlier on in the synchronous steps of the methods that take configs and produces rejection via TypeError (invalid input).

chcunningham (approved):
LGTM. Looks like all comments are addressed. On naming, sticking w/ "no-preference" given majority approval.

PR #315, opened 2021-07-29
Add Vorbis to the registry

padenot:
Does what it says. I have an open question:

Are the three header packet required, or just the last one? The first one allows getting the channel number/sample-rate, what if it contradicts what has been passed in? Do we consider it kind of part of the container and expect authors to bit-shift their way to this piece of info to set up the decoder (which would work, but then it's clearly in the codec bytestream, so that's unclear), or do we state (like is in the proposed text here) that the values in the `AudioDecoderConfig` are not required, or ignored ? The second header is useless for Web Codecs in any case.

### DIFF
--- a/vorbis_codec_registration.src.html
+++ b/vorbis_codec_registration.src.html
@@ -15,7 +15,7 @@ Abstract: This registration is entered into the [[webcodecs-codec-registry]].
     It describes, for Vorbis, the (1) fully qualified codec strings, (2)
     the {{AudioDecoderConfig.description}} bytes, (3) the codec-specific
     {{EncodedAudioChunk}} [=EncodedAudioChunk/[[internal data]]=] bytes, and (4)
-    the meaning of {{EncodedAudioChunk}} [=EncodedAudioChunk/[[type]]=].
+    the values of {{EncodedAudioChunk}} [=EncodedAudioChunk/[[type]]=].
 
     The registration is not intended to include any information on whether a
     codec format is encumbered by intellectual property claims. Implementers and
@@ -34,11 +34,11 @@ Markup Shorthands:css no, markdown yes, dfn yes
 <pre class='anchors'>
 spec: WEBCODECS; urlPrefix: https://w3c.github.io/webcodecs/#
     type: attribute
-        for: EncodedAudioChunkType; text: key; url: dom-encodedaudiochunktype-key
         text: AudioDecoderConfig.description; url: dom-audiodecoderconfig-description
         text: AudioDecoderConfig.sampleRate; url: dom-audiodecoderconfig-samplerate
         text: AudioDecoderConfig.channelCount; url: dom-audiodecoderconfig-channelcount
     type: dfn
+        for: EncodedAudioChunkType; text: key; url: dom-encodedaudiochunktype-key
         for: EncodedAudioChunk; text: [[internal data]]; url: dom-encodedaudiochunk-internal-data-slot
         for: EncodedAudioChunk; text: [[type]]; url: dom-encodedaudiochunk-type-slot
     type: interface
@@ -92,11 +92,11 @@ NOTE: The comments header content is not used by [[WEBCODECS]].
 EncodedAudioChunk type {#encodedaudiochunk-type}
 ================================================
 
-The [=EncodedAudioChunk/[[type]]=] is not used in this codec, its value is
-ignored.
+The [=EncodedAudioChunk/[[type]]=] for an {{EncodedAudioChunk}} containing
+Vorbis is always "[=EncodedAudioChunkType/key=]".
 
 NOTE: Once the initialization has succeeded, any Vorbis packet can be decoded at
-any time, not necessarily in sequence, there is no concept of key-frames.
+any time without error, but this might not result in the expected audio output.
 
 Privacy and Security Considerations {#privacy-and-security-considerations}
 ==========================================================================
--- a/vorbis_codec_registration.src.html
+++ b/vorbis_codec_registration.src.html
@@ -1,0 +1,98 @@
+<pre class='metadata'>
+Title: Vorbis WebCodecs Registration
+Repository: w3c/webcodecs
+Status: NOTE-ED
+Shortname: webcodecs-vorbis-codec-registration
+Level: none
+Group: mediawg
+ED: https://w3c.github.io/webcodecs/vorbis_codec_registration.html
+TR: https://www.w3.org/TR/webcodecs-vorbis-codec-registration/
+Editor: Chris Cunningham, w3cid 114832, Google Inc. https://www.google.com/
+Editor: Paul Adenot, w3cid 62410, Mozilla https://www.mozilla.org/
+Editor: Bernard Aboba, w3cid 65611, Microsoft Corporation https://www.microsoft.com/
+
+Abstract: This registration is entered into the [[webcodecs-codec-registry]].
+    It describes, for Vorbis, the (1) fully qualified codec strings, (2)
+    the {{AudioDecoderConfig.description}} bytes, (3) the codec-specific
+    {{EncodedAudioChunk}} [=EncodedAudioChunk/[[internal data]]=] bytes, and (4)
+    the meaning of {{EncodedAudioChunk}} [=EncodedAudioChunk/[[type]]=].
+
+    The registration is not intended to include any information on whether a
+    codec format is encumbered by intellectual property claims. Implementers and
+    authors are advised to seek appropriate legal counsel in this matter if they
+    intend to implement or use a specific codec format. Implementers of
+    WebCodecs are not required to support the Vorbis codec.
+
+    This registration is non-normative.
+
+Markup Shorthands:css no, markdown yes, dfn yes
+!Participate: <a href="https://github.com/w3c/webcodecs">Git Repository.</a>
+!Participate: <a href="https://github.com/w3c/webcodecs/issues/new">File an issue.</a>
+!Version History: <a href="https://github.com/w3c/webcodecs/commits">https://github.com/w3c/webcodecs/commits</a>
+</pre>
+
+<pre class='anchors'>
+spec: WEBCODECS; urlPrefix: https://w3c.github.io/webcodecs/#
+    type: attribute
+        for: EncodedAudioChunkType; text: key; url: dom-encodedaudiochunktype-key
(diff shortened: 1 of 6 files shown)


padenot:
I think the failure in the job `codec_registry.src.html` is expected since the VORBIS entry is not public yet.

dalecurtis (approved):
lgtm, thanks!

Are vorbis packets bit-exact decodable in isolation? I.e., with MP3, Opus, and AAC we have the concept of preroll such that while every packet can be decoded independently the resulting waveform will vary based on order. I don't think you need to change this text or anything, I'm just curious.

vorbis_codec_registration.src.html
```diff
@@ -0,0 +1,98 @@
...
+NOTE: The comments header content is not used by [[WEBCODECS]].
+
+EncodedAudioChunk type {#encodedaudiochunk-type}
+================================================
+
+The [=EncodedAudioChunk/[[type]]=] is not used in this codec, its value is
+ignored.
+
+NOTE: Once the initialization has succeeded, any Vorbis packet can be decoded at
+any time, not necessarily in sequence, there is no concept of key-frames.
```
sandersdan, line 91: In my mind, this is the same as saying that all Vorbis chunks should have type `key`.

sandersdan:
> Do we consider it kind of part of the container and expect authors to bit-shift their way to this piece of info to set up the decoder (which would work, but then it's clearly in the codec bytestream, so that's unclear), or do we state (like is in the proposed text here) that the values in the AudioDecoderConfig are not required, or ignored ? The second header is useless for Web Codecs in any case.

I'm not familiar enough with Vorbis to have an opinion here, so just general guidance:

- I'm not certain why the `AudioDecoderConfig` `sampleRate` and `numberOfChannels` are required, but they are similar to the `VideoDecoderConfig` `codedWidth` and `codedHeight`, which are also required. Coded size is used as a hint for decoder selection (and to apply any generic UA constraints early) rather than treated as relevant to the decode and output process, since all supported video codecs include it in-band. We might consider making the `AudioDecoderConfig` fields optional (or removing them?) if no codec actually needs them.
- For `description`, I recommend using whatever the most common representation, even if that includes unnecessary metadata. If there is flexibility in the format, ideally we would support the variants. If the metadata is treated as in-band by containers, then it should not be extracted to `description` for WebCodecs (c.f. H.264 Annex B in WebCodecs).
- If the metadata is in-band, and UAs are likely to want it to apply constraints, then it should be required in the `AudioDecoderConfig`. If the metadata is in `description` then this is less important, but I would still recommend requiring it to reduce the need for UAs to violate layers. I'm not sure what sort of global UA constraints might apply to audio, but a maximum number of channels seems likely, and a maximum sample rate is at least possible.

vorbis_codec_registration.src.html
padenot: Yes, I went back and forth between "every packet is a key" or "this doesn't matter at all". I decided to simply ignore it, but I could certainly be convinced of the opposite behaviour.

padenot:
> Are vorbis packets bit-exact decodable in isolation? I.e., with MP3, Opus, and AAC we have the concept of preroll such that while every packet can be decoded independently the resulting waveform will vary based on order. I don't think you need to change this text or anything, I'm just curious.

Having the right order is necessary to get the correct output, because Vorbis uses a lapped transform (MDCT) as part of decoding, so there is some state in the decoder.

However, you can take any packet and decode it as long as the decoder has been initialized, and it will not error out, it will simply produce something unexpected, so it's what you describe.

padenot:
> I'm not familiar enough with Vorbis to have an opinion here, so just general guidance:
> 
>     * I'm not certain why the `AudioDecoderConfig` `sampleRate` and `numberOfChannels` are required, but they are similar to the `VideoDecoderConfig` `codedWidth` and `codedHeight`, which are also required. Coded size is used as a hint for decoder selection (and to apply any generic UA constraints early) rather than treated as relevant to the decode and output process, since all supported video codecs include it in-band. We might consider making the `AudioDecoderConfig` fields optional (or removing them?) if no codec actually needs them.

Filed #317 to discuss this, it's an annoying issue.

>     * For `description`, I recommend using the most common representation, even if that includes unnecessary metadata. If there is flexibility in the format, ideally we would support the variants. If the metadata is treated as in-band by containers, then it should not be extracted to `description` for WebCodecs (c.f. H.264 Annex B in WebCodecs).

Ok, that sounds good. In this case, this is part of the codec, not the container, it's specifically in the codec byte stream specification. There is no flexibility, it's always 3 header packets, and the spec is pretty clear about the number of packets, their type, and the order in which they must appear in the byte stream. Quoting the spec:

>> Before decoding can begin, a decoder must initialize using the bitstream headers matching the stream to be decoded. Vorbis uses three header packets; all are required, in-order, by this specification. Once set up, decode may begin at any audio packet belonging to the Vorbis stream. In Vorbis I, all packets after the three initial headers are audio packets. 

>     * If the metadata is in-band, and UAs are likely to want it to apply constraints, then it should be required in the `AudioDecoderConfig`. If the metadata is in `description` then this is less important, but I would still recommend requiring it to reduce the need for UAs to violate layers. I'm not sure what sort of global UA constraints might apply to audio, but a maximum number of channels seems likely, and a maximum sample rate is at least possible.

Here for Vorbis, in the metadata parsed in the codec byte stream, there is the sample-rate and the channel count. The sample-rate is `uint32`, the channel count `uint8`, so there's room for UA decisions (nobody is going to want to decode a 1Hz Vorbis file I assume). The UA can reject at this time, based on the description. We can then make those two attributes optional (or remove them altogether, only adding something in the registry for `WAV`, per #317).

I think in general I agree with all three of your points though.

sandersdan:
> In this case, this is part of the codec, not the container, it's specifically in the codec byte stream specification. There is no flexibility, it's always 3 header packets, and the spec is pretty clear about the number of packets, their type, and the order in which they must appear in the byte stream.

I read through what I could find about containerizing Vorbis, and there is some disagreement. I found:

- A draft MP4 proposal which doesn't extract a codec-specific `description` but instead defines a new structure for the purpose.
- MKV/WebM extract the headers to `Private Data`, but also prepends a few bytes of sizing information.
- Ogg, which has everything in-band.

On the demux side, I found:

- FFmpeg puts the headers in `extradata`, in the MKV `Private Data` format.
- Android MediaCodec expects the headers to be separated, ID and Setup are passed as CSD0 and CSD1.

It doesn't seem like there is a consistent approach, but the MKV style is probably the default approach on the web? There may be an argument for supporting Ogg-style in-band also, but requiring Ogg demuxers to pull it out doesn't seem terrible either.

padenot:
> * A draft MP4 proposal which doesn't extract a codec-specific `description` but instead defines a new structure for the purpose.

I had a look with the muxers I have at hand, and they seem to put the 3 header packets verbatim, in succession, in an `'esds` box. What draft did you find? I did not find anything seemingly authoritative either.

> * MKV/WebM extract the headers to `Private Data`, but also prepends a few bytes of sizing information.

This is because of the containerization, yes. It's the EBML id (`0x63A2`) and the box size, which is a Matroska variable integer. Then the three header packets are present in succession, verbatim, as a single unit.

Considering this, I think it would be appropriate to ask authors to provide the three header packets as the description, and then `decode()` would only ever take audio packets (packets of type `0`) as its input (i.e. what you say last in your comment).

sandersdan:
> I had a look with the muxers I have at hand, and they seem to put the 3 header packets verbatim, in succession, in an `'esds` box. What draft did you find? I did not find anything seemingly authoritative either.

> This is because of the containerization, yes. It's the EBML id (`0x63A2`) and the box size, which is a Matroska variable integer. Then the three header packets are present in succession, verbatim, as a single unit.

FFmpeg's ogg vorbis parser seems to construct new length headers for `extradata` here: https://source.chromium.org/chromium/chromium/src/+/main:third_party/ffmpeg/libavformat/oggparsevorbis.c;l=240;drc=d0b89fe9c3d6ecc1ec4ffaf0baeb6b36c34ba7fd.

I am proposing that these bytes are also included in the WebCodecs definition. I don't see them mentioned in VORBIS 4.2.

> Yes, I went back and forth between "every packet is a key" or "this doesn't matter at all". I decided to simply ignore it, but I could certainly be convinced of the opposite behaviour.

FWIW, I strongly prefer marking these as "key". I think that's an accurate description of how audio codecs have worked in the past, but I expect future audio codec formats to rely on delta frames.

It also makes sense given that we already require "key" frames after `configure()` and `flush()`.

padenot:
> FFmpeg's ogg vorbis parser seems to construct new length headers for `extradata` here: https://source.chromium.org/chromium/chromium/src/+/main:third_party/ffmpeg/libavformat/oggparsevorbis.c;l=240;drc=d0b89fe9c3d6ecc1ec4ffaf0baeb6b36c34ba7fd.

Looks like I looked at this too quickly indeed, it's verbatim _after a header_. This is the standard Xiph extradata format for communicating between demuxer and decoder, so it looks like a perfect fit for Web Codecs. Gecko (but also lots of others) does this as well: https://searchfox.org/mozilla-central/source/dom/media/XiphExtradata.h. It's quite simple (and the code linked is easy to read):
1. Write the number of packets to a buffer as u8, that MUST be < 255
2. For each packet, write its size as lacing values in the buffer
3. For each packet, copy it to the buffer

This is described [here](https://xiph.org/vorbis/doc/framing.html).

The count in 1. is called the `page_segments` and 2. is called the `segment_table`, lacing values are described in the section **Packet segmentation**.

Gecko also uses this for passing extradata to the Theora decoder, and [here](https://www.matroska.org/technical/codec_specs.html#a_vorbis) is the specification for WebM (in the matroska spec), that normatively requires this format, so it's quite convenient.

> FWIW, I strongly prefer marking these as "key". I think that's an accurate description of how audio codecs have worked in the past, but I expect future audio codec formats to rely on delta frames.

Ok. Thanks for the review, I'll update this patch.

padenot:
> and [here](https://www.matroska.org/technical/codec_specs.html#a_vorbis) is the specification for WebM (in the matroska spec), that normatively requires this format, so it's quite convenient.

Note that this is in contradiction with https://wiki.xiph.org/Oggless, but the latter is clearly marked as non-normative.

dalecurtis:
Defer to Dan on this one. lgtm though

dalecurtis (approved):

sandersdan (approved):
LGTM, thanks!

PR #318, opened 2021-08-03
Generalize the Makefile

padenot:
This uses the same targets (no target for local development and remote for the CI), but won't
need any further modification to change the list of files to process.

Additionally, it checks an environment variable called
WEBCODECS_IGNORE_WARNINGS, so that folks that add new registry entries
can still have their new document generated.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,78 +1,41 @@
-SHELL=/bin/bash
+SHELL := /bin/bash
 
-local: local-index local-codec-registry local-avc-codec-registration vorbis_codec_registration
+DST := $(patsubst %.src.html,%.html,$(wildcard *.src.html))
+REMOTE := $(filter remote,$(MAKECMDGOALS))
 
-local-index: index.src.html
-	bikeshed --die-on=warning spec index.src.html index.html
-
-local-codec-registry: codec_registry.src.html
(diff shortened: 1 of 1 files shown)


dalecurtis:
lg2m, but don't know this build system so forwarded to @tidoust

dalecurtis:
Since @tidoust seems OOO, I'm going to approve this since it looks syntactically correct to me % any platform voodoo that may exist.

dalecurtis (approved):

tidoust:
No comment on my side. The Makefile is now only useful for editors to build specs locally, so what it does is totally up to them!

The build mechanism for publication on GitHub and /TR rather uses the [auto-publish](https://github.com/w3c/webcodecs/blob/main/.github/workflows/auto-publish.yml) GitHub workflow, which got updated through #315 and will need to be updated whenever new entries get added to the registry (which pending registry PR seem to be taking care of).

FWIW, publication to /TR of registry documents won't work until after these documents have been published as First Public Working Drafts. That is normal.

(FWIW2, automatic publication to /TR of WebCodecs currently fails because the recent switch to the new process document and patent policy following WG rechartering has not yet been reflected in the Status of This Document section. I'll fix that)

padenot:
Thanks ! I've rebased this, and will rebase all the others after merging.

PR #319, opened 2021-08-03
Add Opus to the registry

padenot:
This has the `Makefile` changes so it builds, but I'll rebase that.

In contrast to Vorbis, the rate and channel count are required ahead of time here. The rate must be in [8, 12, 16, 24, 48] kHz, and the channel count is either 1 or 2, this is a user request, the decoder outputs what is requested as long as it's a valid Opus bytestream.

### DIFF
--- a/opus_codec_registration.src.html
+++ b/opus_codec_registration.src.html
@@ -76,6 +76,13 @@ An {{EncodedAudioChunk}} containing Opus can be in two different formats,
 depending on if {{AudioDecoderConfig.description}} has been set during
 initialization.
 
+
+AudioDecoderConfig description {#audiodecoderconfig-description}
+================================================================
+
+{{AudioDecoderConfig.description}} can optionally set to an Identification
+Header, described in section 5.1 of [[OPUS-IN-OGG]].
+
 If an {{AudioDecoderConfig.description}} has been set, the
 {{EncodedAudioChunk}}s passed to the {{AudioDecoder.decode}}} method have to be
 in audio data packets, as described in [section
@@ -85,13 +92,6 @@ If an {{AudioDecoderConfig.description}} has not been set, then the
 {{EncodedAudioChunk}}s passed to the {{AudioDecoder.decode}} method have to be
 Opus packets, as described in [section
 3](https://datatracker.ietf.org/doc/html/rfc6716#section-3) of [[OPUS]].
-
-
-AudioDecoderConfig description {#audiodecoderconfig-description}
-================================================================
-
-{{AudioDecoderConfig.description}} can optionally set to an Identification
-Header, described in section 5.1 of [[OPUS-IN-OGG]].
 
 EncodedAudioChunk type {#encodedaudiochunk-type}
 ================================================
--- a/opus_codec_registration.src.html
+++ b/opus_codec_registration.src.html
@@ -1,0 +1,92 @@
+<pre class='metadata'>
+Title: Opus WebCodecs Registration
+Repository: w3c/webcodecs
+Status: NOTE-ED
+Shortname: webcodecs-opus-codec-registration
+Level: none
+Group: mediawg
+ED: https://w3c.github.io/webcodecs/opus_codec_registration.html
+TR: https://www.w3.org/TR/webcodecs-opus-codec-registration/
+Editor: Chris Cunningham, w3cid 114832, Google Inc. https://www.google.com/
+Editor: Paul Adenot, w3cid 62410, Mozilla https://www.mozilla.org/
+Editor: Bernard Aboba, w3cid 65611, Microsoft Corporation https://www.microsoft.com/
+
+Abstract: This registration is entered into the [[webcodecs-codec-registry]].
+    It describes, for Opus, the (1) fully qualified codec strings, (2)
+    the {{AudioDecoderConfig.description}} bytes, (3) the codec-specific
+    {{EncodedAudioChunk}} [=EncodedAudioChunk/[[internal data]]=] bytes, and (4)
+    the values of {{EncodedAudioChunk}} [=EncodedAudioChunk/[[type]]=].
+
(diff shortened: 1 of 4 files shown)


opus_codec_registration.src.html
```diff
@@ -0,0 +1,92 @@
...
+{{EncodedAudioChunk}} [=EncodedAudioChunk/[[internal data]]=] is expected to be
+a "packet", as described in the section 3 of the [[OPUS]] specification.
+
+NOTE: The self-delimiting framing described in Appendix B of [[OPUS]] is not
+supported.
+
+AudioDecoderConfig description {#audiodecoderconfig-description}
+================================================================
+
+{{AudioDecoderConfig.description}} is not used when decoding Opus.
```
dalecurtis, line 76: I don't think this is true? It is used to set things like the pre skip, gain, and channel map. See https://source.chromium.org/chromium/chromium/src/+/main:third_party/ffmpeg/libavcodec/libopusdec.c;l=67

sandersdan: Looks like the format there is the same as https://opus-codec.org/docs/opus_in_isobmff.html#4.3.2, which seems to be specific to the MP4 embedding.

It's extremely similar to OpusHead though (https://datatracker.ietf.org/doc/html/rfc7845#section-5.1) which is used for WebM.

AndroidMediaCodec takes OpusHead in CSD 0, but also pre-skip and pre-roll in CSD 1 and 2. (MP4 stores pre-roll separately in AudioRollRecoveryEntry.)

> the channel count is either 1 or 2

This requirement isn't mentioned in the spec text, but it seems overly restrictive. It does seem to match what FFmpeg supports when there is no `extradata`.

padenot: Aren't those demuxer concerns ? The pre-skip, gain and channel map as linked here are in encapsulation definition (respectively in OGG, matroska and mp4 in @sandersdan's links).

> This requirement isn't mentioned in the spec text

https://datatracker.ietf.org/doc/html/rfc6716#section-2.1.2

I guess the real question is at which abstraction level Web Codecs sits at. What is in this PR is strictly at the RFC 6716 level (in terms of implementation that would be ~`libopus` without the `multistream` functions), but [Gecko](https://searchfox.org/mozilla-central/source/dom/media/platforms/agnostic/OpusDecoder.cpp) and `ffmpeg` both have a layer like what @dalecurtis's links with some use for the extradata. It's possible for authors to do this layer (it's no more complicated than any other codec: a gain, channel re-mapping, and dropping a few packets, with some packet mangling if in Appendix B framing), or it could be decided that Web Codecs handles it. What is probably the most useful for developers is to handle both, and it's not really more complicated in terms of implementation.

We can spec it so that:

- if no `description` is passed in, then there is 1 or 2 channels, and it's decoded at the rate specified in the `AudioDecoderConfig` (this is appropriate for a single Opus stream). It fails if channels > 2 or if the rate is not in the output rate list. This allows developers to decode raw Opus packets with minimal setup, probably good enough for lots of use-cases.
- if a `description` is passed in and is valid, then pre-skip, gain and channel mapping are used, the format being what is defined in rfc7845. The channel count and mapping found in the `desciption` override the `AudioDecoderConfig` attributes. The implementation might want to validate that the channel count and mapping are consistent at this time (like Gecko and ffmpeg do). This means someone that demuxes an mp4 will have to reconstruct an OpusHead structure, but it's not very hard. `elst` are used for preskip though, per the spec, so I'm not sure what to do there.

If we decide to do this, then Web Codecs must support the framing described in Appendix B.

All this is borderline in terms of demuxer/decoder boundary, but it's probably for the best.

Thoughts ?

dalecurtis: I think they're decoder concerns since they are fed directly into the libopus configuration. If those fields were something that an implementation must manually apply to decoded outputs I'd agree it's a demuxer concern, but that's not the case here.

We could split them out into actual AudioDecoderConfig attributes, but my expectation is that `description` is effectively codec specific configuration data that is passed around without high level users needing to understand it.

Your proposal to support both sounds good to me. I think generally almost every codec that could use a `description` will have to include text on what to do when it's present and not present.

dalecurtis (approved):
Thanks Paul!

opus_codec_registration.src.html
```diff
@@ -0,0 +1,110 @@
...
+Fully qualified codec strings {#fully-qualified-codec-strings}
+==============================================================
+
+The codec string is `"opus"`.
+
+EncodedAudioChunk data {#encodedaudiochunk-data}
+================================================
+
+An {{EncodedAudioChunk}} containing Opus can be in two different formats,
+depending on if {{AudioDecoderConfig.description}} has been set during
```
sandersdan, line 76: Nit: I think it's better to say that chunks can be in one of two formats, and leave the association with `description` for the next section. In this case we don't need to also map an enum value so I don't think that will make it as indirect as the AVC registration.

sandersdan (approved):

opus_codec_registration.src.html
```diff
@@ -0,0 +1,110 @@
...
+Level: none
+Group: mediawg
+ED: https://w3c.github.io/webcodecs/opus_codec_registration.html
+TR: https://www.w3.org/TR/webcodecs-opus-codec-registration/
+Editor: Chris Cunningham, w3cid 114832, Google Inc. https://www.google.com/
+Editor: Paul Adenot, w3cid 62410, Mozilla https://www.mozilla.org/
+Editor: Bernard Aboba, w3cid 65611, Microsoft Corporation https://www.microsoft.com/
+
+Abstract: This registration is entered into the [[webcodecs-codec-registry]].
+    It describes, for Opus, the (1) fully qualified codec strings, (2)
```
chrisn, line 15: Minor nit: the (1) -> (1) the (for consistency with other items in the list)

Issue #317, opened 2021-07-30
Are `AudioDecoderConfig.{sampleRate,channelCount}` always required, sometimes ignored, moved to the registry ?


padenot:
Some audio codec will need them, some won't use them.

For WAV, it's just a stream of numbers after having parsed the RIFF header. Web Codecs will only deal with the numbers (it's going to be either a `memcpy` or a trivial inflation from 24bits to 32bits really), and those numbers are meaningless without having a sample-rate and channel count, so it's kind of required.

But it's useless for the others, where those numbers are not in the container, just in the codec byte stream, and we don't want to have authors parse the header of the codec to get the correct values, because that would break the container/codec abstraction. 

It's annoying to have something optional in the IDL but required for a particular codec.

dalecurtis:
EC: Lets keep it required, codec registry entries with optional fields can state that it's ignored.

PR #323, opened 2021-08-05
Add mp3 to the registry

padenot:
This is using the latest revision of the text (from [1998](https://www.iso.org/standard/26797.html)), so it's in 2.4.2.2, but generally the same as in MSE ("an mp3 frame").

Again this has the Makefile changes so that it builds, please ignore.

### DIFF
--- a/mp3_codec_registration.src.html
+++ b/mp3_codec_registration.src.html
@@ -1,0 +1,96 @@
+<pre class='metadata'>
+Title: MP3 WebCodecs Registration
+Repository: w3c/webcodecs
+Status: NOTE-ED
+Shortname: webcodecs-mp3-codec-registration
+Level: none
+Group: mediawg
+ED: https://w3c.github.io/webcodecs/mp3_codec_registration.html
+TR: https://www.w3.org/TR/webcodecs-mp3-codec-registration/
+Editor: Chris Cunningham, w3cid 114832, Google Inc. https://www.google.com/
+Editor: Paul Adenot, w3cid 62410, Mozilla https://www.mozilla.org/
+Editor: Bernard Aboba, w3cid 65611, Microsoft Corporation https://www.microsoft.com/
+
+Abstract: This registration is entered into the [[webcodecs-codec-registry]].
+    It describes, for MP3, the (1) fully qualified codec strings, (2)
+    the {{AudioDecoderConfig.description}} bytes, (3) the codec-specific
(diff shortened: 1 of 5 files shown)


dalecurtis (approved):
Thanks Paul!

mp3_codec_registration.src.html
```diff
@@ -0,0 +1,94 @@
...
+{{EncodedAudioChunk}} [=EncodedAudioChunk/[[internal data]]=] is expected to be
+a "frame", as described in the section 2.4.2.2 of the [[MP3]] specification.
+
+AudioDecoderConfig description {#audiodecoderconfig-description}
+================================================================
+
+{{AudioDecoderConfig.description}} is not used for this codec.
+
+The {{AudioDecoderConfig.sampleRate}} and {{AudioDecoderConfig.channelCount}}
+members are not used.
```
sandersdan, line 78: What does not used mean exactly? I do expect that UAs can still reject configurations based on their values.

mp3_codec_registration.src.html
```diff
@@ -0,0 +1,94 @@
...
+The {{AudioDecoderConfig.sampleRate}} and {{AudioDecoderConfig.channelCount}}
+members are not used. 
+
+
+EncodedAudioChunk type {#encodedaudiochunk-type}
+================================================
+
+The [=EncodedAudioChunk/[[type]]=] for an {{EncodedAudioChunk}} containing
+mp3 is always "[=EncodedAudioChunkType/key=]".
+NOTE: Once the initialization has succeeded, any mp3 packet can be decoded at
```
sandersdan, line 86: Does this render correctly without the double space and second line indent?

sandersdan (approved):

mp3_codec_registration.src.html
padenot: This was discussed last week with @dalecurtis, but we can discuss it further: https://github.com/w3c/webcodecs/issues/317#issuecomment-892801222, which is not amazing, but was considered better than making the field optional in IDL but sometimes required from the registry, that is non-normative. But maybe there's something more elegant, that we could do.

In the case of an mp3 sequence, I don't expect authors to look at the byte stream to determine the rate and channel count ahead of decoding. I expect an author to just give the packets to the decoder, having passed any value to this sample-rate and channel count, and get the rate and channel count on the `AudioData`.

PR #325, opened 2021-08-06
Add AV1 to the codec registry.

dalecurtis:
As titled. This adds AV1 to the registry following the model used by recent
audio codec registry additions like FLAC.


dalecurtis:
Hmm, I was expecting this would be relative to the FLAC one, not include it. But oh well, I'll rebase after the others are submitted.

sandersdan (approved):

dalecurtis:
@wantehchang in case they have any thoughts (or want to route to someone who does) on the text here.

padenot (approved):

PR #330, opened 2021-08-09
Fix accidental duplication in workflow for codec registry.

dalecurtis:
There were some issues with the merges for the vp8, vp9 additions. Fix the duplication.


chcunningham (approved):
LGTM

PR #335, opened 2021-08-11
Add AAC to the registry

padenot:
This is for now only speccing that the is ADTS frames like in MSE. There's probably some value in accepting raw AAC packets`raw_aac()` + the (at least) two bytes of extradata (`DecoderSpecificInfo`) so that authors don't have to remux.

### DIFF
--- a/aac_codec_registration.src.html
+++ b/aac_codec_registration.src.html
@@ -62,14 +62,14 @@ spec: WEBCODECS; urlPrefix: https://w3c.github.io/webcodecs/#
 Fully qualified codec strings {#fully-qualified-codec-strings}
 ==============================================================
 
-This codec has multiple possible codec string:
+This codec has multiple possible codec strings:
 
-- `"mp4a.40.2"`
-- `"mp4a.40.02"`
-- `"mp4a.40.5"`
-- `"mp4a.40.05"`
-- `"mp4a.67"`
-- `"mp4a.40.29"`
+- `"mp4a.40.2"` — MPEG-4 AAC LC
+- `"mp4a.40.02"` — MPEG-4 AAC LC, leading 0 for Aud-OTI compatibility
+- `"mp4a.40.5"` — MPEG-4 HE-AAC v1 (AAC LC + SBR)
+- `"mp4a.40.05"` — MPEG-4 HE-AAC v1 (AAC LC + SBR), leading 0 for Aud-OTI compatibility
+- `"mp4a.40.29"` — MPEG-4 HE-AAC v2 (AAC LC + SBR + PS)
+- `"mp4a.67"` — MPEG-2 AAC LC
 
 AudioDecoderConfig description {#audiodecoderconfig-description}
 ================================================================
@@ -78,7 +78,7 @@ If {{AudioDecoderConfig.description}} is present, it is assumed to a
 `AudioSpecificConfig` as defined in [[iso14496-3]] section 1.6.2.1, Table 1.15,
 and the bitstream is assumed to be in {{AacBitstreamFormat/aac}}.
 
-If the {{AudioDecoderConfig.description}} is not present, ht ebitstream is
+If the {{AudioDecoderConfig.description}} is not present, the bitstream is
 assumed to be in {{AacBitstreamFormat/adts}} format.
 
 The {{AudioDecoderConfig.sampleRate}} and {{AudioDecoderConfig.channelCount}}
--- a/aac_codec_registration.src.html
+++ b/aac_codec_registration.src.html
@@ -1,0 +1,178 @@
+<pre class='metadata'>
+Title: AAC WebCodecs Registration
+Repository: w3c/webcodecs
+Status: NOTE-ED
+Shortname: webcodecs-aac-codec-registration
+Level: none
+Group: mediawg
+ED: https://w3c.github.io/webcodecs/aac_codec_registration.html
+TR: https://www.w3.org/TR/webcodecs-aac-codec-registration/
+Editor: Chris Cunningham, w3cid 114832, Google Inc. https://www.google.com/
+Editor: Paul Adenot, w3cid 62410, Mozilla https://www.mozilla.org/
+Editor: Bernard Aboba, w3cid 65611, Microsoft Corporation https://www.microsoft.com/
+
+Abstract: This registration is entered into the [[webcodecs-codec-registry]].
+    It describes, for AAC, (1) the fully qualified codec strings, (2)
+    the {{AudioDecoderConfig.description}} bytes, (3) the codec-specific
+    {{EncodedAudioChunk}} [=EncodedAudioChunk/[[internal data]]=] bytes, and (4)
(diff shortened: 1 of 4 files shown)


chcunningham:
Editors Call:
Generally looks good. @chcunningham wants to double check on ADTS framing (i.e. is that what ffmpeg expects). Also discussed adding codec strings for HE-AAC v1 and v2.

chcunningham:
I synched w/ @dalecurtis on the ADTS question. The outcome is we want to support both ADTS and raw AAC. The story here is very similar to what was done with H.264, where we registered support for both "annex-b" and "avc" formats. 
- Users may have (or desire to create) content in either format. 
- Decoders/Encoders may only support one format or the other. For example, example Android's MediaCodec and some Chromecast devices only support ADTS for LC AAC. For sophisticated UAs, we should always be able to support either one (by just doing internal conversion). 
- Newer profiles like XHE-AAC actually aren't compatible with ADTS (background https://chromium-review.googlesource.com/c/chromium/src/+/1992006)

In light of the above, and in alignment with what we did for H264/AVC, I propose:
- Define `enum AacBitstreamFormat { "aac", "adts"}` and `dictionary AacEncoderConfig { AacBitstreamFormat format = "aac"; }`
- Define "aac" and "adts" in style of  [AvcBitstreamFormat](https://www.w3.org/TR/webcodecs-avc-codec-registration/#avc-bitstream-format)
- Extend `partial dictionary AudioEncoderConfig { AacEncoderConfig aac; }`
- Revise AudioDecoderConfig description section to say that providing a description signals "aac" format, while omitting signals "adts" format (similar to [AVC section](https://www.w3.org/TR/webcodecs-avc-codec-registration/#videodecoderconfig-description))

padenot:
We probably should.

aac_codec_registration.src.html
```diff
@@ -0,0 +1,178 @@
...
+    "publisher": "ISO",
+    "date": "2009-09"
+  }
+}
+</pre>
+
+Fully qualified codec strings {#fully-qualified-codec-strings}
+==============================================================
+
+This codec has multiple possible codec string:
```
chcunningham, line 65: nit: codec string(s)

aac_codec_registration.src.html
```diff
@@ -0,0 +1,178 @@
...
+
+- `"mp4a.40.2"`
```
chcunningham, line 67: Can you offer some description of the AAC profiles described by the different strings (and which strings are simply duplicates for historical reasons). E.g. mp4a.40.2
```
      //   mp4a.40.2   - MPEG-4 AAC LC
      //   mp4a.40.02  - MPEG-4 AAC LC (leading 0 in aud-oti for
      //                 compatibility)
      //   mp4a.40.5   - MPEG-4 HE-AAC v1 (AAC LC + SBR)
      //   mp4a.40.05  - MPEG-4 HE-AAC v1 (AAC LC + SBR) (leading 0
      //                 in aud-oti for compatibility)
      //   mp4a.40.29  - MPEG-4 HE-AAC v2 (AAC LC + SBR + PS)
```

Chrome also recognizes these MPEG2 strings.
```
      //   mp4a.66     - MPEG-2 AAC MAIN
      //   mp4a.67     - MPEG-2 AAC LC
      //   mp4a.68     - MPEG-2 AAC SSR
```

I'm not an expert on these, but Wikipedia suggests that the LC of mpeg-4 may also include "Perceptual Noise Substitution (PNS)", not found in that of mpeg-2. Maybe that's a meaningful distinct from mpeg4 LC above (particularly if you have some ancient mpeg-2 codec). Our decode pipeline just sends it all to ffmpeg.

aac_codec_registration.src.html
```diff
@@ -0,0 +1,178 @@
...
+- `"mp4a.40.29"`
+
+AudioDecoderConfig description {#audiodecoderconfig-description}
+================================================================
+
+If {{AudioDecoderConfig.description}} is present, it is assumed to a
+`AudioSpecificConfig` as defined in [[iso14496-3]] section 1.6.2.1, Table 1.15,
+and the bitstream is assumed to be in {{AacBitstreamFormat/aac}}.
+
+If the {{AudioDecoderConfig.description}} is not present, ht ebitstream is
```
chcunningham, line 81: typo: ht ebitstrea

aac_codec_registration.src.html
```diff
@@ -0,0 +1,178 @@
...
+section 4.4.2.1 of [[iso14496-3]].
+
+
+EncodedAudioChunk type {#encodedaudiochunk-type}
+================================================
+
+The [=EncodedAudioChunk/[[type]]=] for an {{EncodedAudioChunk}} containing
+AAC is always "[=EncodedAudioChunkType/key=]".
+
+NOTE: Once the initialization has succeeded, any AAC packet can be decoded at
```
chcunningham, line 169: What's the history of this ominous comment? I saw it on some other registrations too.

chcunningham:
> I wonder if the similar thing Opus should be made explicit like this. It's somewhat comparable. Or, conversely, if passing in extradata here is enough to make it accept raw packets. In any case, both should be supported.

Sounds good. I filed #342

chcunningham (approved):
LGTM % small comments.

aac_codec_registration.src.html
padenot: Here's Gecko list: https://searchfox.org/mozilla-central/source/dom/media/VideoUtils.cpp#664-673, it only has `"mp4a.67"` from this second list. AFAIK (I just searched our bugzilla), we haven't had reports of this being a problem, did Chromium hear anything about those?

Not that I'm opposed about adding more though.

aac_codec_registration.src.html
padenot: It's just to warn authors that no error will be reported, ever, if the packet are shuffled around, but the audio will be incorrect, since the decoders are stateful, unlike with some video codecs/implementation of those codecs, that have various types of frames  that need to be fed in order.

aac_codec_registration.src.html
chcunningham: I'm not sure on the history of the MPEG-2 strings. We don't have good metrics to say what current usage is like.  @dalecurtis, any reason we shouldn't register these? Do we mind their continued usage?

@padenot does Gecko generally decode AAC w/ FFmpeg? If so, you may actually support these even without having advertised it. It could be that this is mostly used on sites that just do <audio src=> without bothering to canPlayType().

padenot: > @padenot does Gecko generally decode AAC w/ FFmpeg? If so, you may actually support these even without having advertised it. It could be that this is mostly used on sites that just do <audio src=> without bothering to canPlayType().

Only on Linux, if `ffmpeg` has been installed on the host. On macOS, Windows and Android, the OS decoders are used.

chcunningham:
EC: @aboba defers on this one

aac_codec_registration.src.html
dalecurtis: These all come from here: https://mp4ra.org/#/object_types -- sticking with only the mp4a.40.xx ones seems reasonable to me in this instance, but whatever is easiest.

PR #337, opened 2021-08-12
Make VideoColorSpace constructor argument optional

tidoust:
The constructor takes a `VideoColorSpaceInit` dictionary as input that does not have any required parameter. Quoting from the Web IDL spec, the input "must be specified as optional and have a default value provided. This is to encourage API designs that do not require authors to pass an empty dictionary value when they wish only to use the dictionary’s default values."
https://heycam.github.io/webidl/#idl-operations

***
<a href="https://pr-preview.s3.amazonaws.com/tidoust/webcodecs/pull/337.html" title="Last updated on Aug 12, 2021, 7:33 AM UTC (81ddaac)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/337/2ef514e...tidoust:81ddaac.html" title="Last updated on Aug 12, 2021, 7:33 AM UTC (81ddaac)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -3982,7 +3982,7 @@ Video Color Space Interface {#video-color-space-interface}
 <xmp class='idl'>
 [Exposed=(Window,DedicatedWorker)]
 interface VideoColorSpace {
-  constructor(VideoColorSpaceInit init);
+  constructor(optional VideoColorSpaceInit init = {});
 
   readonly attribute VideoColorPrimaries? primaries;
   readonly attribute VideoTransferCharacteristics? transfer;


sandersdan:
Hmm, not so sure about this one. The current shape comes from a previous iteration where `fullRange` was not optional, and I don't expect to add new required fields in the future, but it doesn't really make sense to construct a default `VideoColorSpace`.

We could do something like make `primaries` required, the concern is that this needlessly complicates cases where apps are translating existing incomplete metadata.

There is some tension here where `colorSpace` is always nullable/optional in our IDL which is treated the same as a default `VideoColorSpace`. There was at least some question about making it required (with a default); if we had done that then it would be useful to construct a default `VideoColorSpace`.

sandersdan (approved):

tidoust:
I see that there is an open discussion to "consider a pattern for dictionary arguments which must have at least one member" in Web IDL, see https://github.com/heycam/webidl/issues/903

Without that pattern, the constraint needs to be specified in prose, e.g. by throwing a `TypeError` in the constructor if none of the members were set as done in [`getUserMedia()`](https://w3c.github.io/mediacapture-main/#dom-mediadevices-getusermedia).

sandersdan:
I don't think that quite fits our case. After pondering for a while, I think we should just merge this (and also reconsider the question of nullability vs defaults).

PR #338, opened 2021-08-12
Accept multiple codec strings for mp3 and h264

### LINKED ISSUE #333: Allow duplicate codec strings
With earlier media apis, codec strings historically have some duplication. 

E.g. for MP3
* mp3
* mp4a.69     - MPEG-2 extension to MPEG-1 (MP3)
* mp4a.6B     - MPEG-1 audio (MP3)
* mp4a.69     - MPEG-2 extension to MPEG-1 (MP3)

For AAC LC
* mp4a.67       - MPEG-2 AAC LC
* mp4a.40.2    - MPEG-4 AAC LC
* mp4a.40.02  - MPEG-4 AAC LC (leading 0 in aud-oti for compatibility)

For H.264
* avc1.*****
* avc3.****

In my understanding, from the WC codec pov, these duplicates mean the same thing as far as identifying their respective codecs (and, where applicable, profiles).

Presently WC registry for [AVC prescribes that duplicates are disallowed](https://www.w3.org/TR/webcodecs-avc-codec-registration/#fully-qualified-codec-strings). The registry for MP3 simply states that the string is "mp3". In short, our direction has been to clean this up a bit. 

But I've had a change of heart. As I pondered actually implementing the restriction in Chrome (e.g. only allowing "mp3" for MP3), it just seems like a headache for implementers and authors with no clear benefit. I think we should continue to have a simple registry table that just uses a single string for a given codec, but we should update the registrations like mp3 and avc to say that the duplicate strings are also allowed.

NOTE: I'm definitely not saying we should allow *ambiguous* strings. IE, you can't just say "avc1", as you can currently with canPlayType(). We should continue to require that strings be fully qualified for WC.

padenot:
This fixes #333.

### DIFF
--- a/mp3_codec_registration.src.html
+++ b/mp3_codec_registration.src.html
@@ -61,7 +61,13 @@ spec: WEBCODECS; urlPrefix: https://w3c.github.io/webcodecs/#
 Fully qualified codec strings {#fully-qualified-codec-strings}
 ==============================================================
 
-The codec string is `"mp3"`.
+This codec has multiple equivalent codec strings:
+
+- `"mp3"
+- `"mp4a.69"`
+- `"mp4a.6B"`
+- `"mp4a.69"`
+
 
 EncodedAudioChunk data {#encodedaudiochunk-data}
 ================================================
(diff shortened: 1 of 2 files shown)


mp3_codec_registration.src.html
```diff
@@ -61,7 +61,13 @@
 Fully qualified codec strings {#fully-qualified-codec-strings}
 ==============================================================
 
-The codec string is `"mp3"`.
+This codec has multiple possible codec strings:
```
chcunningham, line 64: maybe s/possible/equivalent/ to clarify that these are interchangeable.

chcunningham (approved):
LGTM % comment

PR #347, opened 2021-08-21
Add visibleRect and displayWidth|Height to VideoFrameInit

### LINKED ISSUE #281: in VideoFrame-from-VideoFrame constructor, allow setting visibleRect and displayWidth/displayHeight
It would be useful to be able to modify the visibleRect, displayWidth, and displayHeight attributes on a VideoFrame in order to represent a subset of the original frame or a different scale. Since VideoFrame is immutable, the best approach seems to be to add these parameters to the "copy constructor."

Example:
```
const subset = new VideoFrame(origVideoFrame, {
  visibleRect: {
    left: origVideoFrame.visibleRect.left + 100,
    top: origVideoFrame.visibleRect.top + 100,
    width: origVideoFrame.visibleRect.width - 200,
    height: origVideoFrame.visibleRect.height - 200
  },
  displayWidth: origVideoFrame.displayWidth,
  displayHeight: origVideoFrame.displayHeight,
});
```

The new parameters would behave identically to those in VideoFramePlaneInit. (However, @sandersdan seemed unsure if that would be the correct behavior if visibleRect is specified and displayWidth/displayHeight is not specified.)

For simplicity, [_Initialize Frame With Resource and Size_](https://w3c.github.io/webcodecs/#videoframe-initialize-frame-with-resource-and-size) should throw TypeError if these new parameters are specified.

This API would be very useful in combination with [MediaStreamTrack Insertable Media Processing using Streams](https://w3c.github.io/mediacapture-transform/) in order to produce a cropped or scaled version of an existing video track.

The alternative to this API for visibleRect would be to use createImageBitmap with cropping parameters, then create a new VideoFrame from the image bitmap; however, this would require conversions and copies, so the result would be lower-quality and slower. The alternative for displayWidth/displayHeight would be to draw the video frame to a canvas, which has the same drawbacks.

I expect this API to be very easy to implement.

chcunningham:
Allows metadata override for efficient crop/resize.

Fixes #281.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/pull/347.html" title="Last updated on Sep 7, 2021, 11:44 PM UTC (526d28f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/347/55d1848...526d28f.html" title="Last updated on Sep 7, 2021, 11:44 PM UTC (526d28f)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -195,6 +195,11 @@ Definitions {#definitions}
     "matrix" → {{VideoMatrixCoefficients/bt709}},
     "fullRange" → `false`]».
 
+: <dfn>Half Away From Zero</dfn>
+:: A rounding method where a number is rounded to the nearest integer and
+    a tie (e.g. 1.5) is broken by rounding up (away from zero) to the next
+    largest integer.
+
 <dfn>Codec Processing Model</dfn> {#codec-processing-model-section}
 ===================================================================
 
@@ -2904,7 +2909,7 @@ dictionary VideoFrameInit {
   // aspect ratio unless an explicit displayWidth and displayHeight are given.
   DOMRectInit visibleRect;
 
-  // Default matches image unless visilbeRect is provided.
+  // Default matches image unless visibleRect is provided.
   [EnforceRange] unsigned long displayWidth;
   [EnforceRange] unsigned long displayHeight;
 };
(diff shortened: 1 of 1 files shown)


chcunningham:
@sandersdan can you take a first pass

index.src.html
```diff
@@ -2898,6 +2898,15 @@
   unsigned long long duration;  // microseconds
   long long timestamp;          // microseconds
   AlphaOption alpha = "keep";
+
+  // Default matches image. May be used to efficiently crop. Will trigger
+  // new computation of displayWidth and displayHeight using image's pixel
+  // aspect ratio unless an explicit displayWidth and displayHeight are given.
+  DOMRectInit visibleRect;
+
+  // Default matches image unless visilbeRect is provided.
```
sandersdan, line 2907: `visibleRect`

index.src.html
```diff
@@ -3384,53 +3408,92 @@
...
+    4. If {{VideoFrameInit/displayWidth}} and {{VideoFrameInit/displayHeight}}
+        [=map/exist=] in |init|, assign them to {{VideoFrame/[[display width]]}}
+        and {{VideoFrame/[[display height]]}} respectively.
+    5. Otherwise:
+        1. Let |widthScale| be the result of dividing |defaultDisplayWidth| by
+            |defaultVisibleRect|.{{DOMRect/width}}.
+        2. Let |heightScale| be the result of dividing |defaultDisplayHeight| by
+            |defaultVisibleRect|.{{DOMRect/height}}.
+        3. Multiply |frame|'s {{VideoFrame/[[visible width]]}} by |widthScale|
+            and round the result (away from zero). Assign the rounded result to
```
sandersdan, line 3492: I'm not sure about "(away from zero)", could be confusing. Can we avoid specifying?

sandersdan (approved):

index.src.html
padenot: It's always positive, this can just say "up", but "away from zero" is quite a common phrasing (e.g. `man 3 round`).

sandersdan: I don't think that interpretation is what was intended here though, I was expecting round-half-up.

chcunningham: I've added "half" to phrase and added dfn for "half away from zero" that aims to clarify.

chcunningham: I've added "half" to phrase and added dfn for "half away from zero" that aims to clarify.

index.src.html
```diff
@@ -3474,6 +3552,24 @@
         {{VideoFrame/[[format]]}}, and |optLayout|.
     9. Return |combinedLayout|.
 
+: <dfn for=VideoFrame>Verify Rect Sample Alignment</dfn> (with |format| and |rect|)
+:: 1. If |format| is `null`, return `true`.
```
chcunningham, line 3556: @sandersdan, seem right?

index.src.html
```diff
@@ -195,6 +195,11 @@
     "matrix" → {{VideoMatrixCoefficients/bt709}},
     "fullRange" → `false`]».
 
+: <dfn>Half Away From Zero</dfn>
```
chcunningham, line 198: Editors Call: No need for this, just say "round". This is what we all understand "round" to mean by default.

chcunningham (commented):
Editors Call: Generally the proposed path re:odd sizes seems ok, taking a closer look.

index.src.html
chcunningham: Yeah, this seemed like the lesser evil vs checking for `null` prior to the call (multiple callers). You lost me a bit on error messages; we aren't reporting an error here.

chcunningham (commented):
Feedback addressed

chcunningham:
@tidoust, can you help me determine why IPR check failed?

PR #349, opened 2021-08-25
Add format to AudioDataCopyToOptions to convert

### LINKED ISSUE #232: Converting AudioSampleFormats for rendering
Following discussion in https://github.com/w3c/webcodecs/issues/179#issuecomment-828856379 we resolved to expose more audio sample formats. This has the benefit of potentially using less memory (vs always float32) and does less throw away conversion. 

But, for decoding -> rendering scenarios, this will sometimes mean that apps need to take an additional step to convert, as WebAudio's AudioWorklet still only supports float32 planar. 

I propose we offer helper method. Something like
`AudioData.getAsPlanarFloat32() -> returns an ArrayBuffer`

@padenot WDYT?

chcunningham:
Fixes #232.

Also, I've added validation of copyTo options to both allocationSize() and
copyTo() (previously only the latter), such that any call to
allocationSize() that succeeds should also succeed with copyTo().

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/pull/349.html" title="Last updated on Sep 7, 2021, 11:50 PM UTC (771b5d5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/349/f25d932...771b5d5.html" title="Last updated on Sep 7, 2021, 11:50 PM UTC (771b5d5)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -2596,7 +2596,7 @@ dictionary AudioDataInit {
         equal to {{AudioData/[[number of channels]]}}, throw a
         {{RangeError}}.
     5. If {{AudioData/[[format]]}} does not equal |destFormat| and the User
-        Agent does not support requested {{AudioSampleFormat}} conversion,
+        Agent does not support the requested {{AudioSampleFormat}} conversion,
         throw a {{NotSupportedError}} {{DOMException}}.
         Conversion to {{f32-planar}} <em class="rfc2119">MUST</em> always be
         supported.
--- a/index.src.html
+++ b/index.src.html
@@ -2532,8 +2532,11 @@ dictionary AudioDataInit {
         {{InvalidStateError}} {{DOMException}}.
     2. Let |copyElementCount| be the result of running the
         [=Compute Copy Element Count=] algorithm with |options|.
-    3. Let |bytesPerSample| be the number of bytes per sample, as defined by
-        the {{AudioData/[[format]]}}.
+    3. Let |destFormat| be the value of {{AudioData/[[format]]}}.
+    4. If |options|.{{AudioDataCopyToOptions/format}} [=map/exists=], assign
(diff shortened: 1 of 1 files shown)


chcunningham:
@tguilbert-google @dalecurtis FYI

chcunningham:
@padenot please give priority to reviewing the API shape / behavior, as we'd like to launch with this in Chrome v1. 

In a follow up, I imagine we'll want to say more about how conversion takes place. Looking at Chrome's impl for just the float32 conversion (e.g. here's [planaru8 to float32](https://source.chromium.org/chromium/chromium/src/+/main:media/base/audio_buffer.cc;drc=54bae635414d532eb776749cdacfb59ba0e40d8b;l=384)), this seems like a potentially dense matrix of info. Lets chat about to spec this consicely.

chcunningham:
Editors Call: Paul to take a closer look, generally looks right. Follow up PR for conversion can simply use min/max/zero from AudioSampleFormat paramaterized steps.

chcunningham (commented):
Thanks!

Re: support  mandate, Chrome's support is currently limited to just conversion to f32 planar. Basically a detail of the longstanding audio pipeline in chrome, which does all mixing in f32. I agree we could implement more. Do you think conversions to other formats beyond f32 planar will useful?

chcunningham:
> Thanks!
> 
> Re: support mandate, Chrome's support is currently limited to just conversion to f32 planar. Basically a detail of the longstanding audio pipeline in chrome, which does all mixing in f32. I agree we could implement more. Do you think conversions to other formats beyond f32 planar will useful?

Lets go ahead and land this and follow up in https://github.com/w3c/webcodecs/issues/352

PR #355, opened 2021-09-10
Update webm-writer2.js

matthiasfostel:
Fix to read frame data in Chromium 95.


chcunningham (approved):
Thanks @matthiasfostel!

PR #361, opened 2021-09-22
Fix MP3 codec string list formatting

chrisn:
This PR fixes a minor formatting issue in the codec string list in the MP3 registry entry.


padenot:
Thanks!

Issue #359, opened 2021-09-21
Update TR published codec registrations (fix broken registry links)


chcunningham:
Several of the links in the https://www.w3.org/TR/webcodecs-codec-registry/ don't work. 

busted: https://www.w3.org/TR/webcodecs-aac-codec-registration/ 
working: https://w3c.github.io/webcodecs/aac_codec_registration.html

@tidoust, can you advise on how to get the codec registrations published to TR? IIRC, something about generating echidna tokens?

tidoust:
To generate Echidna tokens, specs needs to have been published as First Public Working Drafts (FPWD) to /TR.

To publish FPWD of these specs, I need approval from the group.

In other words, that all starts with issuing a call for consensus to publish the list of WebCodecs registration specs as FPWD, noting that these specs are intended to be published as WG Notes. @chrisn, @jernoble?

chrisn:
Yes, we can send out a CfC. I noticed that the [registry](https://w3c.github.io/webcodecs/codec_registry.html) contains a couple of inline issues. Can we resolve these before FPWD?

tidoust:
I note that the registry itself and the AVC (H.264) WebCodecs Registration have already been published as FPWD (and /TR version gets updated whenever a push gets made to the main branch on these specs). The call for consensus would be for the other WebCodecs registrations, namely the ones for FLAC, MP3, AAC, OPUS, VORBIS I, AV1, VP8, and VP9.

chcunningham:
Thanks guys. I've sent #367 and #368 to prepare for FPWD. When those merge we should be good to run Cfc.

chcunningham:
The blocking issues above are now resolved. @chrisn, please go ahead with CFC to publish the new registrations. 

For reference, the new registrations to publish are:
- https://w3c.github.io/webcodecs/aac_codec_registration.html
- https://w3c.github.io/webcodecs/alaw_codec_registration.html
- https://w3c.github.io/webcodecs/av1_codec_registration.html
- https://w3c.github.io/webcodecs/flac_codec_registration.html
- https://w3c.github.io/webcodecs/mp3_codec_registration.html
- https://w3c.github.io/webcodecs/opus_codec_registration.html
- https://w3c.github.io/webcodecs/pcm_codec_registration.html
- https://w3c.github.io/webcodecs/vorbis_codec_registration.html
- https://w3c.github.io/webcodecs/vp8_codec_registration.html
[...]

PR #367, opened 2021-09-24
Register alaw, ulaw, and linear pcm. Remove stale inline issues.

chcunningham:
Our earlier thinking was that alaw and ulaw did not needing registration documents, as they have simple codec strings and no need for AudioDecoderConfig.description bytes. But we've since used the registry to define the details of EncodedAudioChunk's internal data, so registration documents are added to provide those details. 

With this, the The TODOs are now all addressed, so  inline ISSUEs can be removed.

### DIFF
--- a/codec_registry.src.html
+++ b/codec_registry.src.html
@@ -81,9 +81,6 @@ Registration Entry Requirements {#registration-entry-requirements}
 Audio Codec Registry {#audio-codec-registry}
 ============================================
 
-ISSUE: Several registrations contain TODOs which should be replaced with links
-    to public specifications covering the mentioned definitions.
-
 <table class='data'>
   <tr>
     <td>**codec string**</td>
@@ -137,9 +134,6 @@ ISSUE: Several registrations contain TODOs which should be replaced with links
 
 Video Codec Registry {#video-codec-registry}
 ============================================
-
-ISSUE: Several registrations contain TODOs which should be replaced with links
-    to public specifications covering the mentioned definitions.
 
 <table class='data'>
   <tr>
--- a/codec_registry.src.html
+++ b/codec_registry.src.html
@@ -122,13 +122,17 @@ Audio Codec Registry {#audio-codec-registry}
   </tr>
   <tr>
     <td>ulaw</td>
-    <td>PCM u-law</td>
-    <td>N/A</td>
+    <td>u-law PCM</td>
+    <td>[u-law PCM WebCodecs
+      Registration](https://www.w3.org/TR/webcodecs-ulaw-codec-registration/)
+    [[WEBCODECS-ULAW-CODEC-REGISTRATION]]</td>
   </tr>
   <tr>
     <td>alaw</td>
-    <td>PCM a-law</td>
-    <td>N/A</td>
+    <td>A-law PCM</td>
+    <td>[A-law PCM WebCodecs
+      Registration](https://www.w3.org/TR/webcodecs-alaw-codec-registration/)
+    [[WEBCODECS-ALAW-CODEC-REGISTRATION]]</td>
   </tr>
 </table>
 
--- a/codec_registry.src.html
+++ b/codec_registry.src.html
@@ -134,6 +134,13 @@ Audio Codec Registry {#audio-codec-registry}
       Registration](https://www.w3.org/TR/webcodecs-alaw-codec-registration/)
     [[WEBCODECS-ALAW-CODEC-REGISTRATION]]</td>
   </tr>
+  <tr>
+    <td>pcm-*</td>
+    <td>Linear PCM</td>
+    <td>[Linear PCM WebCodecs
+      Registration](https://www.w3.org/TR/webcodecs-pcm-codec-registration/)
+    [[WEBCODECS-ALAW-CODEC-REGISTRATION]]</td>
+  </tr>
 </table>
 
 Video Codec Registry {#video-codec-registry}
--- a/pcm_codec_registration.src.html
+++ b/pcm_codec_registration.src.html
@@ -1,0 +1,117 @@
+<pre class='metadata'>
+Title: Linear PCM WebCodecs Registration
+Repository: w3c/webcodecs
+Status: NOTE-ED
+Shortname: webcodecs-pcm-codec-registration
+Level: none
+Group: mediawg
+ED: https://w3c.github.io/webcodecs/pcm_codec_registration.html
+TR: https://www.w3.org/TR/webcodecs-pcm-codec-registration/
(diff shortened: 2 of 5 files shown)


chrisn:
Perfect, thanks Chris. When this is merged I'll go ahead with the CfC per #359.

padenot:
What about linear PCM? It's common enough, and it seems that forcing authors to write a different path altogether to handle it for no real reason isn't great. Adding it also means that authors will be able to use the sample format conversion facilities that Web Codecs will provide.

chcunningham:
I'm probably confused, but I thought linear PCM was already our native AudioData format, such that "decoding" alaw/ulaw (or any audio codec) would produce linear PCM.

padenot:
Yes, it's unnecessary technically, but authors probably don't want to have a special code path to handle it, they just want to use `AudioDecoder` regardless of the audio bit stream.

In practice, it would just output the same data as you said.

codec_registry.src.html
```diff
@@ -125,22 +122,23 @@
   </tr>
   <tr>
     <td>ulaw</td>
-    <td>PCM u-law</td>
-    <td>N/A</td>
+    <td>u-law PCM</td>
```
chrisn, line 125: A minor thing, should we refer to this as μ-law, to be more precise (here and in the registration entry)?

chcunningham:
> Yes, it's unnecessary technically, but authors probably don't want to have a special code path to handle it, they just want to use `AudioDecoder` regardless of the audio bit stream.
> 
> In practice, it would just output the same data as you said.

I'm supportive. Chrome does exactly what you describe when given a wav file w/ linear PCM. 

Unfortunately, we have a bit of a yak to shave. In FFmpeg and probably other libraries, linear PCM has several codec identifiers to distinguish sample formats and endianness. Chrome presently supports (since 2015)

* AV_CODEC_ID_PCM_U8
* AV_CODEC_ID_PCM_S16LE
* AV_CODEC_ID_PCM_S24LE
* AV_CODEC_ID_PCM_S32LE
* AV_CODEC_ID_PCM_F32LE

My first thought would be mint new codec strings for each of these similar to the above. For ex
AV_CODEC_ID_PCM_U8 -> pcm-u8
AV_CODEC_ID_PCM_S16LE -> pcm-s16-le (or maybe just pcm-s16, since no plans to support BE formats)

I'd then register these all together under the string pcm-*

Sound right?

codec_registry.src.html
chcunningham: I'd like to keep the actual string used in code as 'ulaw' (using unicode chars in code seems burdensome), so I'm worried that using the μ anywhere may lead to confusion and copy/paste

I agree μ-law is more precise. FWIW, I do find some precedent for ulaw
https://developer.apple.com/library/archive/documentation/QuickTime/QTFF/QTFFChap3/qtff3.html
http://mp4ra.org/#/codecs

chcunningham:
EC: Proposal sounds good

> AV_CODEC_ID_PCM_S16LE -> pcm-s16-le (or maybe just pcm-s16, since no plans to support BE formats)

Drop -le ... and just document that its semantics are little endian. Big endian systems (rare) would need to shift ints to little endian upstream of decoder.

chcunningham:
Latest commit adds Linear PCM to the registry. Here's a [preview of the html output](https://storage.googleapis.com/chcunningham/pcm_codec_registration.html).

A few things to point out:
- I minted a `pcm-s24` codec string, but did not add an `s24` sample format. This follows ffmpeg's model (e.g. [AV_CODEC_ID_PCM_S24LE](https://source.chromium.org/chromium/chromium/src/+/main:third_party/ffmpeg/libavcodec/codec_id.h;drc=e1712fe5e4cb5af066118968f3d8cb2fc13cb0d3;l=325), but no [AV_SAMPLE_FMT_S24](https://source.chromium.org/chromium/chromium/src/+/main:third_party/ffmpeg/libavutil/samplefmt.h;drc=e1712fe5e4cb5af066118968f3d8cb2fc13cb0d3;l=58)). 
- Earlier, I thought I would need to formally define s24, but then I realized we don't actually have need for this in the code. Folks can make EncodedAudioChunks and containing 24-bit PCM and decode them without ever referring to a 24-bit sample format. The only thing we really need is to describe how 24-bit samples work inside an EncodedAudioChunk, and I do that now inline as part of the `pcm-s24` codec string dfn.
- I did not end up referencing [Multimedia Programming Interface and Data Specifications 1.0](http://www-mmsp.ece.mcgill.ca/Documents/AudioFormats/WAVE/Docs/riffmci.pdf). While page 58 does a decent job of describing samples and interleaving, I think the WebCodecs spec is actually already more thorough (kudos @padenot) and I reference that already with dnfs for "sample", "interleaved" etc.

pcm_codec_registration.src.html
```diff
@@ -0,0 +1,117 @@
...
+==============================================================
+
+This codec's string begins with the prefix "pcm-", followed by a suffix that
+indicates the sample format. The complete list of strings and associated
+formats is as follows.
+
+* pcm-u8, using format {{u8}}
+* pcm-s16, using format {{s16}}
+* pcm-s24, using three [[WEBIDL#idl-octet|byte]] (24-bit) [=samples=] with
+    [=interleaved=] [[webcodecs#audio-buffer-arrangement|channel arrangement]].
```
padenot, line 78: All of those are interleaved, what makes s24 different? The way it's packed, 3 bytes by 3 bytes ?

pcm_codec_registration.src.html
```diff
@@ -0,0 +1,117 @@
...
+================================================
+
+Linear pulse code modulation (linear PCM) describes a format where the audio
+values are sampled at a regular interval, and where the quantization levels
+between two successive values are linearly uniform.
+
+{{EncodedAudioChunk}} [=EncodedAudioChunk/[[internal data]]=] is expected to be
+a sequence of bytes of arbitrary length, with a [=sample=] occuring every N
+bits, where N is defined by the codec string. For multichannel PCM, [=samples=]
+from different channels are [=interleaved=].
```
padenot, line 97: The fact that it's interleaved is well described here.

pcm_codec_registration.src.html
chcunningham: No difference, all interleaved, but I'm metnioning it here just because we don't have s24, so this is part of the inline dfn for that format.

pcm_codec_registration.src.html
chcunningham: Agree

chcunningham (commented):
EC: LGTM

PR #376, opened 2021-10-06
Make VideoFrame.timestamp non nullable, less fussy

chcunningham:
It was always intended that timestamp be required when constructing a
VideoFrame. Previously, the spec steps threw a TypeError for frames
constructed with a CanvasImageSource that didn't otherwise have a
timestamp. The only way to actually construct a frame w/ a null
timestamp was to decode an image (ImageDecoder) that was not animated.
But animated images always start at time `0` and increment by the frame
duration. Hence, `0` seems seems like a reasonable default for non-animated.

Also, for VideoFrame constructed images, rather than strictly
requiring folks to supply a timestamp when constructing with things like
Canvas, we can be less fussy and just let it default to `0`, similar to
image decoded frames. Full disclosure: This is behavior is (by mistake)
already implemented in Chrome.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/pull/376.html" title="Last updated on Oct 6, 2021, 6:59 AM UTC (4c72501)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/376/d0a2078...4c72501.html" title="Last updated on Oct 6, 2021, 6:59 AM UTC (4c72501)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -2931,7 +2931,7 @@ interface VideoFrame {
   readonly attribute unsigned long displayWidth;
   readonly attribute unsigned long displayHeight;
   readonly attribute unsigned long long? duration;  // microseconds
-  readonly attribute long long? timestamp;          // microseconds
+  readonly attribute long long timestamp;           // microseconds
   readonly attribute VideoColorSpace colorSpace;
 
   unsigned long allocationSize(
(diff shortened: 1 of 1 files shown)


chcunningham:
EC: @padenot mentioned concern about learnability for encoding from canvas w/ default timestamp of zero (potentially doesn't go well).

chcunningham:
@padenot, checking in on [your concern](https://github.com/w3c/webcodecs/pull/376#issuecomment-936639849) from the EC. How would you like to proceed?

chcunningham:
Abandoning this. We resolved instead to always require a timestamp and Chrome deprecated the earlier behavior that defaulted to zero.

We still want VideoFrame.timestamp to be non-nullable, but I'll do that in a follow up for clarity.

PR #387, opened 2021-10-25
Move MP4 demuxe/decode demo to workers

tguilbert-google:
This PR moves the MP4 demux+decode demo to fetch, demux and decode in a worker, posting frames back to the main thread.

This change is done in order to align our demos with the recommended best practice of offloading webcodecs from the main window to workers.

### DIFF
--- a/samples/mp4-decode/mp4_demuxer.js
+++ b/samples/mp4-decode/mp4_demuxer.js
@@ -111,7 +111,7 @@ class Writer {
   }
 }
 
-export class MP4Demuxer {
+class MP4Demuxer {
   constructor(uri) {
     this.source = new MP4Source(uri);
   }
@@ -140,7 +140,6 @@ export class MP4Demuxer {
     for (i = 0; i < avccBox.SPS.length; i++) {
       writer.writeUint16(avccBox.SPS[i].length);
       writer.writeUint8Array(avccBox.SPS[i].nalu);
-      window.temp = avccBox.SPS[i].nalu;
     }
 
     writer.writeUint8(avccBox.nb_PPS_nalus);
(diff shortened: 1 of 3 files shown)


chcunningham:
IIUC this is still doing sync w/ main thread at every output to draw to the canvas. I think we instead want to use https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvas#asynchronous_display_of_frames_produced_by_an_offscreencanvas to avoid that sync.

tguilbert-google:
I purposely did not go that route, because IIUC, that functionality is [not available on Safari or Firefox](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/transferControlToOffscreen#browser_compatibility).

Granted, WC is also only available on Chromium for now. I can still update the demo regardless.

chcunningham:
Yeah, I'd go ahead and make the change so folks can see the full picture of how to be entirely off main thread. FWIW, I expect Firefox and Safari will ship OffscreenCanvas before they ship WebCodecs.

tguilbert-google:
I'm updated the demo to draw from an offscreen canvas. I also added frame stats to the video frames themselves.

samples/mp4-decode/mp4_demuxer.js
```diff
@@ -111,7 +111,7 @@ class Writer {
   }
 }
 
-export class MP4Demuxer {
```
chcunningham, line 114: why is this no longer needed?

chcunningham (approved):
LGTM. Stats to canvas is clever :)

samples/mp4-decode/mp4_demuxer.js
tguilbert-google: There's probably a way to keep this file as a module, and to still use MP4Box.js as a script. I had tried Dan's suggestion of making the worker a module, but I still ran into issues. Making everything a script was the simplest way that I could get the demo to work.

PR #400, opened 2021-11-10
Update publish & ignore file w/ new registrations

chcunningham:
Updates publish workflow and gitignore to include new pcm, alaw, and ulaw registrations.


chcunningham:
FYI @padenot @aboba - minor infra update, will go ahead and merge.

PR #404, opened 2021-11-16
Fix the type of the Promise returned by VideoEncoder.isTypeSupported.

### LINKED ISSUE #403: VideoEncoder.isConfigSupported returns boolean?
I'm writing a polyfill for WebCodecs and found an inconsistency: in the definition for `VideoEncoder`, the spec says that `VideoEncoder.isConfigSupported` returns `Promise<boolean>`, which is inconsistent with all other `isConfigSupported`s, and also the actual description of `VideoEncoder.isConfigSupported` describes returning a `Promise<VideoEncoderSupport>`, which is consistent. I'm assuming this was just a typo, or perhaps an old version that wasn't updated.

padenot:
This fixes #403.

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -1167,7 +1167,7 @@ interface VideoEncoder {
   undefined reset();
   undefined close();
 
-  static Promise<boolean> isConfigSupported(VideoEncoderConfig config);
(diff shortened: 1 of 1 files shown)


chcunningham (approved):
LGTM

PR #412, opened 2021-11-25
Define importing VideoFrame into WebGPU

kainino0x:
The WebGPU community group is so far unable to reach consensus on adding support for WebCodecs interop, because WebCodecs is not agreed upon by the community group members. It has been tentatively proposed to define WebGPU-WebCodecs interop in WebCodecs instead.

Draft for discussion.

Issue: gpuweb/gpuweb#2124

***
<a href="https://pr-preview.s3.amazonaws.com/kainino0x/webcodecs/pull/412.html" title="Last updated on Dec 9, 2021, 3:16 AM UTC (690d53b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/412/ff5738b...kainino0x:690d53b.html" title="Last updated on Dec 9, 2021, 3:16 AM UTC (690d53b)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -3788,7 +3788,7 @@ objects, which can be read from WebGPU shaders.
 The definition of the {{GPUExternalTextureDescriptor/source}} member of
 {{GPUExternalTextureDescriptor}} is modified to allow {{VideoFrame}}:
 
-<xmp>
+<xmp class='idl'>
 partial dictionary GPUExternalTextureDescriptor {
     required (HTMLVideoElement or VideoFrame) source;
 };
@@ -3798,7 +3798,7 @@ The lifetime of such a {{GPUExternalTexture}} object is implicitly tied to the
 lifetime of the source {{VideoFrame}} object.
 When the {{VideoFrame}} is [=Close VideoFrame|closed=]
 (e.g. via {{VideoFrame/close()}} or
-[[videoframe-transfer-serialization#|transferring the VideoFrame]]),
+[[#videoframe-transfer-serialization|transferring the VideoFrame]]),
 the {{GPUExternalTexture}} is *destroyed*, so no further operations
 using it can be issued.
 
--- a/index.src.html
+++ b/index.src.html
@@ -101,6 +101,14 @@ spec: webrtc-svc; urlPrefix: https://w3c.github.io/webrtc-svc/
 
 spec: visibility-state; urlPrefix: https://www.w3.org/TR/page-visibility/#
     type: enum-value; text: hidden; url: dom-visibilitystate-hidden
+
+spec: webgpu; urlPrefix: https://www.w3.org/TR/webgpu/#
+    type: interface; text: GPUExternalTexture; url: gpuexternaltexture
+    type: dictionary; text: GPUExternalTextureDescriptor; url: dictdef-gpuexternaltexturedescriptor
+    for: GPUExternalTextureDescriptor;
+        type: dict-member; text: source; url: dom-gpuexternaltexturedescriptor-source
+    for: GPUDevice;
+        type: method; text: importExternalTexture(); url: dom-gpudevice-importexternaltexture
 </pre>
 
 <pre class='biblio'>
@@ -3771,6 +3779,26 @@ conversion is explicitly disabled.
 Color space conversion during {{ImageBitmap}} construction is controlled by
 {{ImageBitmapOptions}} {{ImageBitmapOptions/colorSpaceConversion}}. Setting this
 value to "none" disables color space conversion.
+
+### WebGPU ### {#videoframe-webgpu}
+
+WebCodecs extends [[WebGPU]] to allow importing {{VideoFrame}} objects via
+{{GPUDevice/importExternalTexture()}}, producing opaque {{GPUExternalTexture}}
+objects, which can be read from WebGPU shaders.
+The definition of the {{GPUExternalTextureDescriptor/source}} member of
+{{GPUExternalTextureDescriptor}} is modified to allow {{VideoFrame}}:
+
+<xmp>
+partial dictionary GPUExternalTextureDescriptor {
+    required (HTMLVideoElement or VideoFrame) source;
+};
+</xmp>
+
+The lifetime of such a {{GPUExternalTexture}} object is implicitly tied to the
+lifetime of the source {{VideoFrame}} object.
+When the {{VideoFrame}} is [=Close VideoFrame|closed=] (implicitly or
+explicitly), the {{GPUExternalTexture}} is *destroyed*, so no further operations
+using it can be issued.
 
 VideoFrame CopyTo() Options {#videoframe-copyto-options}
 ------------------------------------------------------------
--- a/index.src.html
+++ b/index.src.html
@@ -3796,8 +3796,10 @@ partial dictionary GPUExternalTextureDescriptor {
 
 The lifetime of such a {{GPUExternalTexture}} object is implicitly tied to the
 lifetime of the source {{VideoFrame}} object.
-When the {{VideoFrame}} is [=Close VideoFrame|closed=] (implicitly or
-explicitly), the {{GPUExternalTexture}} is *destroyed*, so no further operations
+When the {{VideoFrame}} is [=Close VideoFrame|closed=]
+(e.g. via {{VideoFrame/close()}} or
+[[videoframe-transfer-serialization#|transferring the VideoFrame]]),
+the {{GPUExternalTexture}} is *destroyed*, so no further operations
 using it can be issued.
 
 VideoFrame CopyTo() Options {#videoframe-copyto-options}


kainino0x:
Tangentially related, an open question came up a while ago when trying to decide how applications would actually use VideoFrame with WebGPU:
https://github.com/gpuweb/gpuweb/issues/1380#issuecomment-861027636

kainino0x:
It's also likely we would want to later be able to import individual VideoFrame planes into WebGPU. This doesn't tackle that yet. See gpuweb/gpuweb#1380.

index.src.html
```diff
@@ -3772,6 +3780,26 @@
...
+{{GPUExternalTextureDescriptor}} is modified to allow {{VideoFrame}}:
+
+<xmp>
+partial dictionary GPUExternalTextureDescriptor {
+    required (HTMLVideoElement or VideoFrame) source;
+};
+</xmp>
+
+The lifetime of such a {{GPUExternalTexture}} object is implicitly tied to the
+lifetime of the source {{VideoFrame}} object.
```
sandersdan, line 3798: The WebGPU spec currently says that `GPUExternalTexture` objects are destroyed after a turn of the microtask queue, in which case having them keep the underlying resource alive should be fine.

Is it expected that this approach will be changed? Are the sampler configurations heavy enough that it would be useful to retain them for longer? If they are retained longer, is their own lifetime insufficient for some reason (eg. being GC-based)?

kainino0x: My proposal is to have GPUExternalTexture tied to different implicit lifetimes depending on how it's created. If from a VideoFrame, it uses the lifetime of the VideoFrame. If from an HTMLVideoElement, I'm proposing a change to use the lifetime of the currently presented frame, rather than an implicit microtask: https://github.com/gpuweb/gpuweb/pull/2302
(The microtask thing was our initial "most cautious" approach but we're finding that it's not the most practical solution for a number of reasons.)

sandersdan: Thanks. What sort of synchronization do you expect apps to need to do to correctly complete rendering before they can call `VideoFrame.close()`?

kainino0x: Synchronization in WebGPU is implicit. Apps just need to submit() any work that uses the external texture before close()ing. So an implementation might need to flush commands on close(), or hold the VideoFrame(Handle) alive internally until flush, and use some internal synchronization to make sure the different subsystems are synchronized.

I'm pretty sure this wouldn't be much different from using VideoFrame as a texImage2D() source in WebGL, when boiled down.

index.src.html
```diff
@@ -3772,6 +3780,26 @@
...
+When the {{VideoFrame}} is [=Close VideoFrame|closed=] (implicitly or
+explicitly), the {{GPUExternalTexture}} is *destroyed*, so no further operations
```
sandersdan, line 3800: > implicitly or explicitly

This should probably be in terms of the effect of close(), which is the clearing of the resource reference. We don't mention implicit close() elsewhere in the spec.

Any suggestions for how best to do that from the spec editors?

sandersdan (approved):
I'm satisfied that this design is compatible with WebCodec's lifetime/ownership model.

index.src.html
kainino0x: "Close VideoFrame" is also invoked by transfers. I can explicitly call that out, or just not go into details.

Strictly speaking, there probably should be some step inside "Close VideoFrame" that does this. But I wasn't sure if it would be appropriate to invade on other parts of the spec for this.

kainino0x:
I've taken this PR out of Draft since the WebGPU group has agreed on hosting this interaction in WebCodecs. Members reiterated that trying to define these interactions in the WebGPU spec was unpalatable because (unlike the rest of WebGPU) there would be no alternate implementations available to validate the design. (Lifetime design in particular - note [specific concerns](https://github.com/gpuweb/gpuweb/issues/2124#issuecomment-977559473) about wired memory on Apple platform(s).)

kainino0x:
(This is ready to review and land whenever WebCodecs editors are happy with it!)

chcunningham:
Sorry for the delay. I'll put it on the agenda for tomorrow's call.

index.src.html
chcunningham: Text you settled on (mentioning transfer and close()) works for me. I'm also happy to support adding steps to "Close Video Frame" if you'd prefer. IIUC, this would be something like: for each texture referencing this frame, set textures [[detached]] to true.

index.src.html
chcunningham: Text you settled on (mentioning transfer and close()) works for me. I'm also happy to support adding steps to "Close Video Frame" if you'd prefer. IIUC, this would be something like: for each texture referencing this frame, set textures [[detached]] to true.

chcunningham: Text you settled on (mentioning transfer and close()) works for me. I'm also happy to support adding steps to "Close Video Frame" if you'd prefer. IIUC, this would be something like: for each texture referencing this frame, set textures [[detached]] to true.

index.src.html
```diff
@@ -3772,6 +3780,26 @@
...
 
+### WebGPU ### {#videoframe-webgpu}
+
+WebCodecs extends [[WebGPU]] to allow importing {{VideoFrame}} objects via
+{{GPUDevice/importExternalTexture()}}, producing opaque {{GPUExternalTexture}}
+objects, which can be read from WebGPU shaders.
+The definition of the {{GPUExternalTextureDescriptor/source}} member of
+{{GPUExternalTextureDescriptor}} is modified to allow {{VideoFrame}}:
+
+<xmp>
```
chcunningham, line 3791: nit: add  class='idl' for idl styling

index.src.html
```diff
@@ -3772,6 +3780,28 @@
...
+partial dictionary GPUExternalTextureDescriptor {
+    required (HTMLVideoElement or VideoFrame) source;
+};
+</xmp>
+
+The lifetime of such a {{GPUExternalTexture}} object is implicitly tied to the
+lifetime of the source {{VideoFrame}} object.
+When the {{VideoFrame}} is [=Close VideoFrame|closed=]
+(e.g. via {{VideoFrame/close()}} or
+[[videoframe-transfer-serialization#|transferring the VideoFrame]]),
```
chcunningham, line 3801: nit: bikeshed linking fails for this dfn.

chcunningham (commented):
Generally looks good to me.

>  (Lifetime design in particular - note specific concerns about wired memory on Apple platform(s).)

Do any concerns remain after considering Dale's points in https://github.com/gpuweb/gpuweb/issues/2124#issuecomment-984906806

kainino0x:
Any tips for building this locally? I'm using bikeshed 3.3.2.

```
LINK ERROR: Multiple possible 'hidden' idl refs for '['Document']'.
Arbitrarily chose https://html.spec.whatwg.org/multipage/interaction.html#dom-document-hidden
To auto-select one of the following refs, insert one of these lines into a <pre class=link-defaults> block:
spec:html; type:attribute; text:hidden
spec:page-visibility-2; type:attribute; text:hidden
{{Document/hidden}}
 ✘  Did not generate, due to fatal errors
```

index.src.html
kainino0x: I'm OK with keeping this localized to this section if you are.

kainino0x:
Never mind, of course as soon as I posted that I realized I could just manually invoke bikeshed (rather than using the makefile) without `--die-on=warning`.

kainino0x:
... hm, that didn't catch my `videoframe-transfer-serialization` error though. How did you catch it?

kainino0x:
FYI I verified the fixes above by building with #419 included and checking the generated HTML output. Marked comments as addressed (and I think this should be OK to land).

> Do any concerns remain after considering Dale's points in gpuweb/gpuweb#2124 (comment)

I don't think so. WebGPU doesn't add any more ways to hold onto video frames than WebCodecs does by itself.

index.src.html
chcunningham: Works for me.

chcunningham (approved):

chcunningham:
Fixes LGTM. @padenot, @aboba did you want additional time to review this? 

> ... hm, that didn't catch my `videoframe-transfer-serialization` error though. How did you catch it?

I think just manually by looking through the generated preview.

Kangz:
All, is there anything blocking forward progress on this PR? Are we waiting for a review from @padenot and @aboba?

padenot:
Conceptually this probably looks alright (from a Web Codecs perspective, after reading the various discussions in WebGPU issues), but it isn't clear to me why this change is in the Web Codecs specification. Monkey patching specifications is a bad practice, and we shouldn't do it. This PR changes the semantics of a dict member and a method, both defined in WebGPU, and doesn't change _anything_ about Web Codecs, it merely uses an object that can be created using an API defined in the Web Codecs specification.

If the extension points needed for specifying the intended behavior of `VideoFrame`s when used in conjunction with `GPUDevice.importExternalTexture(...)` are lacking, then we can add them here for sure, but it looks fine: `[[Detached]]` can be checked in WebGPU to know whether the `VideoFrame` is still valid.

Kangz:
My understanding (and @kainino0x can expand) is that there are two reasons why this needs a change in the WebCodec specification:

 - It seems more natural that the destruction of the `GPUExternalTexture` is done immediately when the `VideoFrame` is closed, instead of polling in the WebGPU specification the status of `[[Detached]]` when the `GPUExternalTexture` is used.
 - There has been concerns by some browser vendors (Mozilla, Apple) about adding references to WebCodec's `VideoFrame` in the WebGPU specification because they want to implement WebGPU but are less clear about WebCodec. The interactions between the two APIs needs to be defined somewhere, so this PR suggests to define it in WebCodec since it is very minimal.

padenot:
>     * It seems more natural that the destruction of the `GPUExternalTexture` is done immediately when the `VideoFrame` is closed, instead of polling in the WebGPU specification the status of `[[Detached]]` when the `GPUExternalTexture` is used.

I understand this, but monkey patching is not ok. Would it be OK to state that [`GPUExternalTexture.[[destroyed]]`](https://gpuweb.github.io/gpuweb/#dom-gpuexternaltexture-destroyed-slot)'s getter returns the value of the underlying `VideoFrame.[[Detached]]` (plus any other bits of behaviour of WebGPU of course, and only when it's backed by a `VideoFrame`'s underlying resource). Doing this, there's no polling, and WebGPU probably knows what to do with a `[[destroyed]]` `GPUExternalTexture` already.

The alternative is to have a way to signal to the `GPUExternalTexture` that the underlying resource has been destroyed, but it feels less elegant than the lazy approach described above.

>     * There has been concerns by some browser vendors (Mozilla, Apple) about adding references to WebCodec's `VideoFrame` in the WebGPU specification because they want to implement WebGPU but are less clear about WebCodec. The interactions between the two APIs needs to be defined somewhere, so this PR suggests to define it in WebCodec since it is very minimal.

It's fine not to implement a specification in its entirety, it's fairly common. That said, speaking for Mozilla, we're definitely implementing Web Codecs, so there's no concern from us here.

At the end of the day, this modifies the behaviour of the WebGPU API, it needs to go into the WebGPU spec.

aboba:
Agree with Paul that WebCodecs is probably not the best place to land this.  This seems related to the all-GPU pipeline discussion in Issue https://github.com/webmachinelearning/webnn/issues/226  

Note the concern in the issue about whether existing ML implementations can make use of GPUExternalTexture

kainino0x:
Thank you all, I'll bring this feedback back to WebGPU and see where it goes. https://github.com/gpuweb/gpuweb/issues/2124#issuecomment-1011314117

Issue #416, opened 2021-12-01
Prehistoric codecs/H.263+ entry


Yahweasel:
Short version: None of the video codecs in the registry are very approachable with a WebAssembly-compiled software-only encoder. Can we get something a bit more prehistoric, such as H.263+?

Long version:

Let me start with a not-hypothetical: If you're strange (bold?) enough to want to build a telephony app where the WebCodecs implementation _might_ be a WebAssembly-compiled polyfill, what codecs should you use? For audio, it's simple: Opus. Opus is so blazingly fast there's essentially no reason to use anything _but_ a polyfill. For video, the answer is a bit murky.

libvpx, bless its heart, can encode in real time, even when compiled to WebAssembly. But, its real-time mode is essentially "do as much as you can in the time budget, then give up". So, if the time budget is very limiting, the encoding looks awful. Real RealMedia vibes. H.264-era and later codecs are just too sophisticated to expect good quality with real-time encoding in WebAssembly, at least for the time being.

Would it be reasonable to add an earlier video codec to the registry? My proposal/suggestion is H.263v2/H.263+. Admittedly, the selection is purely self-serving—I can easily encode and decode H.263+ in real time in WebAssembly—but I think it's also a reasonable choice.

Some considerations:

- There is no standard codec tag in MIME types for H.263+. Something would have to be invented, presumably "h263v2", "h263p", or "h263+".
- Why not MPEG-4 Part 2? Patents. Besides, if real-time is the concern, then telephony is probably the application, and MPEG-4 Part 2's enhancements over H.263+ are mostly geared towards mastered video, not telephony applications. MPEG-4 Part 2 is backwards-compatible, so existing decoders can be used off-the-shelf for H.263+. Probably several of them are tweakable to emit H.263+ as well, but I'm insufficiently familiar to say this with any confidence.
- Why not plain H.263? Its very limited selection of video frame sizes.
- Why not even earlier? H.261, MPEG-1 video, H.262 (MPEG-2), and H.263 are all so similar to each other, practical software implementations I'm aware of all perform very similarly, so I just can't find a compelling reason for them.
- Chrome presumably won't support H.263+. Well, that's what polyfills are for :)

dalecurtis:
There's no limitation on what codecs can be put in the registry since all codecs are optional to implementers. Please feel free to contribute a registry entry for H.263.

Yahweasel:
[...]

PR #340, opened 2021-08-13
Replace uses of rfc2119 keywords in non-normative sections

padenot:
This is split in multiple commits for ease of review, but the second commit has most of the actual changes.

There is a single occurrence that I thought was good to make normative (third commit). It's a MAY, so it's not a requirement per se.

Also, three section (background, and recommendation for developers) are non-normative, marked as such, and their prose has been fixed.

I'm not sure yet if we want to capitalize the words, maybe we do. This is easily done in any case, because they only appear normatively now.

This fixes #301.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/pull/340.html" title="Last updated on Aug 20, 2021, 1:34 PM UTC (3d41cca)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/340/55d1848...3d41cca.html" title="Last updated on Aug 20, 2021, 1:34 PM UTC (3d41cca)">Diff</a>

PR #417, opened 2021-12-03
Adding a registration for H.263

Yahweasel:
This PR addresses my own #416, and adds a registration for H.263. Included in the registration is all revisions of H.263 by ITU-T, and for the latest revision (v3, H.263++), the profiles and levels specified in Annex X. I believe it covers all realistic use cases of H.263 excluding MPEG-4 Part 2, which should be covered by a separate registration if at all.

### DIFF
--- a/h263_codec_registration.src.html
+++ b/h263_codec_registration.src.html
@@ -75,21 +75,21 @@ The codec string begins with the prefix `"h263."`, with a suffix of 1 character
 describing the version of the specification: `"h263.1"` for the first revision
 ([[H.263v1]]), `"h263.2"` for the second revision ([[H.263v2]]), or `"h263.3"`
 for the third revision ([[H.263v3]]). As the specifications are backwards
-compatible, an encoder may encode using an earlier revision than specified, and
-a decoder may decode using a later revision than specified.
+compatible, an encoder can encode using an earlier revision than specified, and
+a decoder can decode using a later revision than specified.
 
-`"h263.3"`, but no prior revision, may further be suffixed with a `"."` and a
+`"h263.3"`, but no prior revision, can further be suffixed with a `"."` and a
 one-character profile number, as specified in Annex X.2 of [[H.263v3]]. If a
-profile is not specified, then an encoder may use any optional features and a
-decoder must support all optional features. This may further be suffixed by a
-`"."` and a two-character level, as described in Table X.2 of [[H.263v3]]. If a
-level is not specified, then an encoder may support any level or may support
-formats other than those in any specified level, and a decoder must support any
-level and may support formats other than those in any specified level.
+profile is not specified, then an encoder can use any optional features and a
+decoder needs to support all optional features. This can further be suffixed by
+a `"."` and a two-character level, as described in Table X.2 of [[H.263v3]]. If
+a level is not specified, then an encoder can support any level or can support
+formats other than those in any specified level, and a decoder needs to support
+any level and can support formats other than those in any specified level.
 
 As revisions of H.263 prior to H.263v3 do not have profiles or levels, if
-`"h263.1"` or `"h263.2"` is specified, then an encoder may support any optional
-features in the specified revision, and a decoder must support all optional
+`"h263.1"` or `"h263.2"` is specified, then an encoder can support any optional
+features in the specified revision, and a decoder needs to support all optional
 features in the specified revision. The non-normative profiles and levels
 specified in Appendix II are not used.
 
@@ -142,7 +142,7 @@ H.263v2 is also known as H.263+ or H.263 1998. H.263v3 is also known as H.263++
 or H.263 2005.
 
 MPEG-4 Part 2 specifies a video encoding format which is backwards compatible
-with H.263 (H.263 streams may be decoded as MPEG-4 Part 2), but it is not
+with H.263 (H.263 streams can be decoded as MPEG-4 Part 2), but it is not
 forwards compatible (MPEG-4 Part 2 streams do not decode as H.263), and it is
 not covered by this registration.
 
--- a/h263_codec_registration.src.html
+++ b/h263_codec_registration.src.html
@@ -1,0 +1,154 @@
+<pre class='metadata'>
+Title: H.263 WebCodecs Registration
+Repository: w3c/webcodecs
+Status: NOTE-ED
+Shortname: webcodecs-h263-codec-registration
+Level: none
+Group: mediawg
+ED: https://w3c.github.io/webcodecs/h263_codec_registration.html
+TR: https://www.w3.org/TR/webcodecs-h263-codec-registration/
+Editor: Yahweasel, Wereweasel https://wereweasel.com/
+
+Abstract: This registration is entered into the [[webcodecs-codec-registry]].
+    It describes, for H.263, (1) the fully qualified codec strings, (2) the
+    codec-specific {{EncodedVideoChunk}} [=EncodedVideoChunk/[[internal
+    data]]=] bytes, (3) the {{VideoDecoderConfig.description}} bytes, (4) the
+    values of {{EncodedVideoChunk}} [=EncodedVideoChunk/[[type]]=], and (5)
+    the naming convention used in this document.
+
+    The registration is not intended to include any information on whether a
+    codec format is encumbered by intellectual property claims. Implementers and
+    authors are advised to seek appropriate legal counsel in this matter if they
+    intend to implement or use a specific codec format. Implementers of
+    WebCodecs are not required to support the H.263 codec.
+
+    This registration is non-normative.
+
+Markup Shorthands:css no, markdown yes, dfn yes
+!Participate: <a href="https://github.com/w3c/webcodecs">Git Repository.</a>
+!Participate: <a href="https://github.com/w3c/webcodecs/issues/new">File an issue.</a>
+!Version History: <a href="https://github.com/w3c/webcodecs/commits">https://github.com/w3c/webcodecs/commits</a>
+</pre>
+
+<pre class='anchors'>
+spec: WEBCODECS; urlPrefix: https://w3c.github.io/webcodecs/#
+    type: attribute
+        text: EncodedVideoChunkMetadata.decoderConfig; url: dom-encodedvideochunkmetadata-decoderconfig
+        for: EncodedVideoChunkType; text: key; url: dom-encodedvideochunktype-key
+        text: VideoDecoderConfig.description; url: dom-videodecoderconfig-description
+    type: dfn
+        for: EncodedVideoChunk; text: [[internal data]]; url: dom-encodedvideochunk-internal-data-slot
+        for: EncodedVideoChunk; text: [[type]]; url: dom-encodedvideochunk-type-slot
+        for: VideoEncoder; text: [[output callback]]; url: dom-videoencoder-output-callback-slot
+    type: interface
+        text: EncodedVideoChunk; url: encodedvideochunk
+        text: VideoEncoder; url: videoencoder
+    type: dictionary
+        text: VideoEncoderConfig; url: dictdef-videoencoderconfig
+        text: VideoDecoderConfig; url: dictdef-videodecoderconfig
+</pre>
+
+<pre class='biblio'>
+{
+  "H.263v1": {
+    "href": "https://www.itu.int/rec/T-REC-H.263-199603-S/en",
+    "title": "H.263: Video coding for low bit rate communication (03/96)",
+    "publisher": "ITU-T"
+  },
+  "H.263v2": {
+    "href": "https://www.itu.int/rec/T-REC-H.263-199802-S/en",
+    "title": "H.263: Video coding for low bit rate communication (02/98)",
+    "publisher": "ITU-T"
+  },
+  "H.263v3": {
+    "href": "https://www.itu.int/rec/T-REC-H.263-200501-I/en",
+    "title": "H.263: Video coding for low bit rate communication (01/2005)",
+    "publisher": "ITU-T"
+  }
+}
+</pre>
+
+Fully qualified codec strings {#fully-qualified-codec-strings}
+==============================================================
+
+The codec string begins with the prefix `"h263."`, with a suffix of 1 character
+describing the version of the specification: `"h263.1"` for the first revision
+([[H.263v1]]), `"h263.2"` for the second revision ([[H.263v2]]), or `"h263.3"`
+for the third revision ([[H.263v3]]). As the specifications are backwards
+compatible, an encoder may encode using an earlier revision than specified, and
+a decoder may decode using a later revision than specified.
+
+`"h263.3"`, but no prior revision, may further be suffixed with a `"."` and a
+one-character profile number, as specified in Annex X.2 of [[H.263v3]]. If a
+profile is not specified, then an encoder may use any optional features and a
+decoder must support all optional features. This may further be suffixed by a
+`"."` and a two-character level, as described in Table X.2 of [[H.263v3]]. If a
+level is not specified, then an encoder may support any level or may support
+formats other than those in any specified level, and a decoder must support any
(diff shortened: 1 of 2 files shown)


Yahweasel:
Just to confirm for the above workflow step, this contribution affects only the non-normative codec registry.

h263_codec_registration.src.html
```diff
@@ -0,0 +1,154 @@
...
+([[H.263v1]]), `"h263.2"` for the second revision ([[H.263v2]]), or `"h263.3"`
+for the third revision ([[H.263v3]]). As the specifications are backwards
+compatible, an encoder may encode using an earlier revision than specified, and
+a decoder may decode using a later revision than specified.
+
+`"h263.3"`, but no prior revision, may further be suffixed with a `"."` and a
+one-character profile number, as specified in Annex X.2 of [[H.263v3]]. If a
+profile is not specified, then an encoder may use any optional features and a
+decoder must support all optional features. This may further be suffixed by a
+`"."` and a two-character level, as described in Table X.2 of [[H.263v3]]. If a
```
sandersdan, line 85: Are levels in H.263 inter-compatible, as in, can an encoder output a lower level than requested like we have defined for H.264?

Yahweasel: With the exception of level 45, each level is strictly looser than all previous levels, so an encoder outputting at, e.g., level 10 is correctly outputting at level 20 or 30 or 40. 45 is an exceptional case, and while all streams in level 45 are also valid bitstreams in all higher levels, it is not true that all valid bitstreams at lower levels are valid as level 45. Because 45 is a weird exception, but all other levels are inter-compatible, I think it's easier to leave that as implicit, rather than repeating the document's rather long-winded explanation of their intercompatibility that I've just detailed here :)

tidoust:
There are two parts in this pull request:
1. It adds a registry entry to the registry. The [requirements](https://www.w3.org/TR/webcodecs-codec-registry/#registration-entry-requirements) are described in the registry and all seems good to me here.
2. It creates a WebCodecs registration document for H.263. Adding that document to the repo means that you're asking the Media Working Group to take responsibility for the document's content, publication and maintenance. We need to make sure that the Media Working Group is ok with that before we proceed. Cc @chrisn @jernoble

Yahweasel:
>     2. It creates a WebCodecs registration document for H.263. Adding that document to the repo means that you're asking the Media Working Group to take responsibility for the document's content, publication and maintenance. We need to make sure that the Media Working Group is ok with that before we proceed. Cc @chrisn @jernoble

My assumption is that this is how it's supposed/intended to work, if for no other reason than that every other codec registration is also maintained in this repository. If it's more reasonable to put the codec registration document elsewhere and simply link it from the codec registry, I'm happy to do that; perhaps it would even be good to be the first one, as it would be the example of how to register a codec externally.

That being said, H.263 is set in stone, so notionally any maintenance is just to do with changes to the information provided by registration documents. Hopefully it's not much of a burden.

chrisn:
> We need to make sure that the Media Working Group is ok with that before we proceed. 

Yes, we need consensus from the Working Group to add this registration. To help us make that decision, can you describe or share a link to your WebCodecs polyfill? Is it open source?

Yahweasel:
> > We need to make sure that the Media Working Group is ok with that before we proceed.
> 
> Yes, we need consensus from the Working Group to add this registration. To help us make that decision, can you describe or share a link to your WebCodecs polyfill? Is it open source?

Of course it's open source!

https://github.com/ennuicastr/libavjs-webcodecs-polyfill

LibAVJS-WebCodecs-Polyfill uses (libav.js)[https://github.com/Yahweasel/libav.js/], which is a port of ffmpeg (via emscripten) I've been maintaining for various projects for a couple years. Since the underlying software is ffmpeg, most of this is automatically on solid footing; I've got the vast majority of the WebCodecs API implemented on the first day I started polyfilling it.

h263_codec_registration.src.html
```diff
@@ -0,0 +1,154 @@
...
+Section 5.1.3 of [[H.263v1]].
+
+H.263v2, H.263v3: If `PTYPE`'s bits 6-8 are not set to 111, then `PTYPE`'s bit
+9 (Picture Coding Type) is 0 (`INTRA`), as defined in Section 5.1.3. If
+`PTYPE`'s bits 6-8 are set to 111, as defined in Section 5.1.3, then
+`PLUSPTYPE`'s `MPPTYPE`'s bits 1-3 (Picture Type Code) are 000 (`INTRA`), as
+defined in Section 5.1.4.3. Section numbers refer to [[H.263v2]] if the second
+revision is used, or [[H.263v3]] if the third revision is used.
+
+Notes on codec naming {#notes-on-codec-naming}
```
chcunningham, line 126: I typically see codec strings break down by profiles rather than annexes. I appreciate that profiles came along pretty late here, but I wonder if they could still serve this purpose and thereby simplify the codec string scheme. At a glance, it looks like the profiles refer to combinations of features from earlier annexes. Would it work to simply have a string like h263.[profile]?

h263_codec_registration.src.html
```diff
@@ -0,0 +1,154 @@
...
+    "title": "H.263: Video coding for low bit rate communication (01/2005)",
+    "publisher": "ITU-T"
+  }
+}
+</pre>
+
+Fully qualified codec strings {#fully-qualified-codec-strings}
+==============================================================
+
+The codec string begins with the prefix `"h263."`, with a suffix of 1 character
```
chcunningham, line 74: I found one mention of "s263" in https://cconcolato.github.io/media-mime-support/. @cconcolato can you shed light on where that comes from and any thoughts on the scheme proposed herein?

h263_codec_registration.src.html
```diff
@@ -0,0 +1,154 @@
...
+describing the version of the specification: `"h263.1"` for the first revision
+([[H.263v1]]), `"h263.2"` for the second revision ([[H.263v2]]), or `"h263.3"`
+for the third revision ([[H.263v3]]). As the specifications are backwards
+compatible, an encoder may encode using an earlier revision than specified, and
```
chcunningham, line 78: nit: this doc has a lot of "may" and "must", which are unfortunately reserved for normative documents (telling implementers what they must/may do). We have to find alternative phrasing (ex: https://github.com/w3c/webcodecs/pull/340)

chcunningham (commented):
> My assumption is that this is how it's supposed/intended to work, if for no other reason than that every other codec registration is also maintained in this repository.

I think you're assumption is correct and @tidoust is just describing the _next_ step in this process. IMO we don't have to block landing your PR on working group review. It could remain an "editors draft" until WG votes to formally add the registration. @tidoust lmk if I have that right.

h263_codec_registration.src.html
```diff
@@ -0,0 +1,154 @@
+<pre class='metadata'>
+Title: H.263 WebCodecs Registration
+Repository: w3c/webcodecs
+Status: NOTE-ED
+Shortname: webcodecs-h263-codec-registration
+Level: none
+Group: mediawg
+ED: https://w3c.github.io/webcodecs/h263_codec_registration.html
+TR: https://www.w3.org/TR/webcodecs-h263-codec-registration/
+Editor: Yahweasel, Wereweasel https://wereweasel.com/
```
chcunningham, line 10: @Yahweasel just to be clear, is it your intention to be on the hook for future maintenance (PR reviews) for this doc?

@tidoust, any issue with that? For my part, happy to have the help, but I don't know relevant W3C rules offhand.

tidoust:
> I think you're assumption is correct and @tidoust is just describing the next step in this process.

Absolutely. Sorry if these process steps feel a bit impersonal. By definition, they are ;)
I personally agree that collecting all registrations in one place under the umbrella of the Media Working Group is a good thing.

> IMO we don't have to block landing your PR on working group review. It could remain an "editors draft" until WG votes to formally add the registration. @tidoust lmk if I have that right.

Different groups have different perspectives. In some groups, participants feel strongly that whatever lands in the repo needs to have been vetted by the group. Other groups take a more casual approach, sometimes flagging drafts as "Unofficial" ([status](https://tabatkins.github.io/bikeshed/#metadata-status) `UD` in Bikeshed) to distinguish them from Editor's Drafts. All such drafts typically originate from group participants, which is not the case here, so I would recommend running the call for consensus before landing the PR.

h263_codec_registration.src.html
tidoust: There is an issue in that, [per process](https://www.w3.org/2021/Process-20211102/#publication), "an editor must be a participant". I'm confident that we can find a suitable workaround though.

@Yahweasel could perhaps be granted an [invited expert status](https://www.w3.org/2021/Process-20211102/#invited-expert-wg), although note that has some licensing implications (the Patent Policy does not apply to this document, but the invited expert status is linked to the group as a whole, and the group develops normative specs), or simply switch to another official spec editor.

chrisn:
I would prefer to see WG consensus before merging. We have a WG call next week, so can include in the agenda.

h263_codec_registration.src.html
GregorR: The problem is that the profiles and levels don't have a clean relationship with the revisions, but encoders and decoders tend to be revision-specific. Only profile 0 is compatible with the original H.263v1, but it doesn't use all the features H.263v1 is allowed to use. Profiles 1–3 are compatible with H.263v2, but leave out features it's allowed to use. Any practical implementation is going to use an off-the-shelf en/decoder, and ffmpeg for instance has h263 (v1) and h263p (v2) as their own codecs.

Yahweasel: The problem is that the profiles don't have a clean relationship with the revisions, but practical codecs are by revision. Profile 0 is only decodable with H.263v1, but doesn't use all of H.263v1's features. Profiles 1–3 are decodable with H.263v2, but don't use all of _its_ features. Profiles 4–8 all require H.263v3. Any practical implementation is going to use off-the-shelf codecs, and in practice, the codecs are by revision. For instance, ffmpeg has "h263" (v1) and "h263p" (v2).

h263_codec_registration.src.html
Yahweasel: Looks like s263 is used for 3GPP, which is one of many "H.263 but..." codec specifications, as is MPEG-4 Part 2. I must admit, I'm wholly unaware of what the "but" part is; it may just be a particular restriction of features.

h263_codec_registration.src.html
Yahweasel: I didn't cap and <em> them because I assumed _that_ was reserved for normative documents. Are "should" and "ought" reserved?

padenot: The list is herre: https://datatracker.ietf.org/doc/html/rfc2119. `should` is reserverd, `ought` is not.

h263_codec_registration.src.html
Yahweasel: I would warn you that I'm _not_, in particular, an expert. I guess I'm an expert in the way that anyone is an expert (I fell into the digital media field and learned while falling), but even as such, my expertise is far more in audio than video. This whole rabbit hole is scratching a particular itch for another project. I intend to maintain that project, of course, and to that end, this is part of that. I don't particularly mind being on the hook for this document, as I expect that (a) there's not much maintenance involved for a codification of a codec that hasn't changed in 15 years, and (b) no one else cares about H.263 :) .

Yahweasel:
I really don't mind one way or the other how things go. I decided I should ask about H.263 in the registry because I intend to use H.263 with my own polyfill, and I'd rather that _not_ be a special exception, so that notionally a future implementation could be compatible. If this is all too messy, then I'll just have it be a special exception in my polyfill.

h263_codec_registration.src.html
Yahweasel: "Can" and "needs to" it is ;)

Yahweasel:
I should also say: I don't even intend to implement H.263v3, as FFmpeg doesn't. H.263v3 is very uncommonly used, because it essentially competes with MPEG-4 Part 2, which isn't quite compatible. I included H.263v3 and its profiles and levels because it felt like it would be really bizarre to codify H.263 but _not_ codify its latest version. This is also why the profiles and levels are quite marginal; practical implementations don't support them, because practical implementations are of H.263v2 or v1.

chrisn:
Discussion from today's working group meeting: https://www.w3.org/2021/12/14-mediawg-minutes.html#t02

h263_codec_registration.src.html
chrisn: [3GPP TS 26.244](https://portal.3gpp.org/desktopmodules/Specifications/SpecificationDetails.aspx?specificationId=1441) defines the `s263.profile.level` codecs string format for H.263++. I wonder if we can reference this for H.263++?

Yahweasel: I would be happy to add the s263.profile.level strings as alternatives, but as mentioned in one of the threads here, the tricky thing with the H.263(+(+)) profiles is that they don't have any clean correspondence to the document revisions, but actual codecs implement document revisions, so _not_ having a way to specify H.263 vs H.263+ vs H.263++ isn't really an option.

chrisn:
Sorry for the delay. The call for consensus passed with no objection, so next step is to invite some more review feedback.

https://lists.w3.org/Archives/Public/public-media-wg/2022Mar/0002.html

chrisn:
We have received some review feedback, which said:

* [RFC 4629](https://datatracker.ietf.org/doc/html/rfc4629) contains media types for video/H263-1998 and video/H263-2000. The proposed WebCodecs registration refers to the 1998 version of H.263 as H.263v2 and the 2005 version as H.263v3, but does not seem to consider the 2000 version explicitly, even though that version is a specific focus in RFC 4629 (and [RFC 2429](https://datatracker.ietf.org/doc/html/rfc2429) before it). That seems a little odd, especially in regard to the status of Annexes U and V. Maybe there should be v3 and v4 names to distinguish between those, and maybe suffixing names by years would be more clear than versions. Year suffixes were used in RFC 4629.

* There is some commentary about assuming decoders support all optional features. This seems rather optimistic and generally impractical. Some of the options may require special system support, for example.

* The latest annex, Annex X, defines profiles and levels of H.263. Specifying packages of capabilities in terms of those profiles and levels would probably be a better idea, as those were agreed by an approved consensus as part of H.263 itself.

chcunningham:
@chrisn, just checking in. What are next steps?

chrisn:
Thanks Chris. Is there any update we can make to the registration based on the latest replies above (particularly the discussion of Annex X profiles and levels)?

If we need further guidance, Gary Sullivan suggested that we could bring this to [ITU-T SG-16](https://www.itu.int/en/ITU-T/studygroups/2022-2024/16/Pages/default.aspx). To do that, I would suggest that we pull together a summary and a set of specific questions. We can then either approach Gary again directly or send as a liaison to ITU-T. If we do that, it may be good to combine with questions about the proposed H.265 registration, to cover everything in one go.

chcunningham:
_If we can_ use profiles, that has my vote. It matches what's done for other codecs and it pushes the decision about how to group features to ITU folks who are certainly the experts. I understand the idea to combine profiles with versions, but that seems too complex. Let me poke a bit more to see if we can reconcile profiles w/ implementations. 

> Popular open source en/decoders of H.263, in particular ffmpeg, don't use them (because open source en/decoders only implement H.263+, not H.263++), and they're not designed to correlate well to H.263 releases. Only profiles 0–3 are restricted to H.263v2 features, but there are H.263v2 features that are only “activated” in profiles 6 and higher, along with H.263v3 features.

I propose that WebCodecs implementers act as a bridge between the versioning schemes. For example, if you're underlying codec supports H.263v2, return supported=true for profiles 0-3. If the underlying codec supports a subset of required features for some higher level profiles (e.g. H.263v2 vs profile 6), I propose that we return supported=false for such profiles. This may mean that we don't advertise the full list of supported features (choosing instead to break at the granularity of profiles), but I imagine that still strikes the right balance between power/simplicity.

PR #437, opened 2022-02-09
Make VideoFrame.timestamp non-nullable.

chcunningham:
Current spec always defines a timestamp for user constructed
VideoFrames. Setting a real timestamp is especially important for frames
that sent to VideoEncoder, as the timestamp is used for rate control.

With no way for users to generate a null timestamp, it makes sense to
make the type non-nullable. This also aligns with AudioData.timestamp.

This patch also addresses the one lingering opporutnity for null
timestamps: ImageDecoder used null as a default for output VideoFrames.
This patch updates that default to be zero. While it could be argued
that null is technically more correct (we don't often present images on
timeline), this is outweighed by the practical impact of using zero
which allows us to define a clear contract: VideoFrame always has a
timestamp.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/pull/437.html" title="Last updated on Sep 19, 2022, 11:20 PM UTC (15183a6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/437/938df12...15183a6.html" title="Last updated on Sep 19, 2022, 11:20 PM UTC (15183a6)">Diff</a>


aboba:
Agree that `VideoFrame.timestamp` should not be nullable.  Would also be good to clarify the `timestamp` requirements for APIs taking `VideoFrame` as an input.  For example, I believe that a non-null (and non-zero?) `VideoFrame.timestamp` required for a stream of VideoFrames provided to `MediaStreamTrackGenerator`.

aboba (approved):

chcunningham:
EC: Will file an issue on MSTG to clarify their requirements. For VideoEncoder, lets add a note encouraging real timestamps, warning of broken rate control.

chcunningham:
EC: Add note explaining intent around animated images (provide real timestamp when feasible, default to zero)

chcunningham:
MSTG issue filed (https://github.com/w3c/mediacapture-transform/issues/80). Clarifying NOTE:s added. 

@padenot LMK if you have any other feedback

index.src.html
```diff
@@ -4686,11 +4693,21 @@
...
-            duration, assign `null` to the corresponding variable.
-        3. Assign {{ImageDecodeResult/image}} with the result of running the
+        2. Let |timestamp| be the presentation timestamp for |output| as
+            described by |encodedFrame|. If |encodedFrame| does not describe a
+            timestamp, assign `0` to |timestamp|.
+
+            NOTE: For animated images, the timestamp of a given frame will
+                generally equal the sum of durations from preceeding frames.
+                However, implementers are not expected to decode earlier frames
+                to compute this timestamp. If it is infeasible to dermine a
```
padenot, line 4703: `s/dermine/determine/`

index.src.html
```diff
@@ -4686,11 +4693,21 @@
...
+                imestamp by simply inspecing |encodedFrame|, timestamp can
+                instead be `0`.
```
padenot, line 4705: This should in fact be normative, not a note, which are non-normative. Probably `SHOULD`: the implementation is supposed to provide a timestamp, but can decide not to do it for good reasons (here, because it's a waste of resources for some codecs/image formats). This also means that an implementation that puts in 0 timestamp is conformant.

index.src.html
```diff
@@ -4743,15 +4760,25 @@
...
+    15. Let |timestamp| be the presentation timestamp for |output| as
+        described by |encodedFrame|. If |encodedFrame| does not describe a
+        timestamp, assign `0` to |timestamp|.
+
+        NOTE: For animated images, the timestamp of a given frame will
+            generally equal the sum of durations from preceeding frames.
+            However, implementers are not expected to decode earlier frames
+            to compute this timestamp. If it is infeasible to dermine a
+            imestamp by simply inspecing |encodedFrame|, timestamp can
+            instead be `0`.
```
padenot, line 4772: same comments apply here as well.

dalecurtis:
@padenot I've addressed all your comments. How does the new PR look?

padenot (approved):
Looks good, but needs a rebase.

dalecurtis:
Thanks! It merged cleanly, so maybe you were looking at an earlier version.

sandersdan:
This doesn't seem to update the Close VideoFrame algorithm, which still sets the timestamp to `null`.

dalecurtis:
Thanks for pointing that out @sandersdan. Given we leave duration unchanged, we should probably leave timestamp unchanged as well. Alternatively, we should set duration to nullptr and timestamp to zero as we do with other fields. @padenot @aboba -- WDYT?

aboba:
Mild preference for leaving timestamp unchanged, but it's not critical either way in the Close algorithm.

padenot:
I prefer leaving the timestamps unchanged.

dalecurtis:
One correction above, we do null the duration today. The one minor downside of leaving timestamp, duration unchanged is that the close() algorithm is intended to release a handle to some internal frame object. Likely the timestamp, duration will be attached to that object, so close() then needs its own local copies of the values. This is easy enough to produce during close(), but it's worth mentioning as a bit of a wart relative to null.

Dan proposed an alternate suggestion where we don't allow null timestamps outside of close(); e.g., there's no way to end up with a null timestamp, duration aside from close(). This seems reasonable to me as a clear indicator that a frame is closed. WDYT?

dalecurtis:
EC: Bernard feels its a bit ugly, but no memory issues so doesn't change opinion. @padenot I'll assume your opinion is unchanged if I don't hear back this week.

PR #438, opened 2022-02-09
Add MIME "type" attribute to ImageDecoder

### LINKED ISSUE #435: Spec ImageDecoder.type
Chrome's ImageDecoder interface shipped with a `type` attribute ([IDL](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/webcodecs/image_decoder.idl;drc=b07499a3508421cacd736ad41be72d8c6dc08499;l=38)).

 [It was originally envisioned](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/webcodecs/image_decoder.idl;drc=b07499a3508421cacd736ad41be72d8c6dc08499;l=38) as the place to return the sniffed type, but now it just returns the type passed in at construction. It's not terribly useful, but it also isn't really a noticeable wart. Considering the tradeoff of breaking sites vs adding a tiny attribute, I vote to just go ahead and add it to the spec.

chcunningham:
Fixes #435.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/pull/438.html" title="Last updated on Feb 9, 2022, 9:00 PM UTC (d75d11c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/438/14cf28f...d75d11c.html" title="Last updated on Feb 9, 2022, 9:00 PM UTC (d75d11c)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -4265,6 +4265,7 @@ enum VideoMatrixCoefficients {
   </dd>
 </dl>
 
+
 Image Decoding {#image-decoding}
 ====================================
 
@@ -4294,6 +4295,7 @@ ImageDecoder Interface {#imagedecoder-interface}
 interface ImageDecoder {
   constructor(ImageDecoderInit init);
 
+  readonly attribute DOMString type;
   readonly attribute boolean complete;
   readonly attribute Promise<undefined> completed;
   readonly attribute ImageTrackList tracks;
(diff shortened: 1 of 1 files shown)


aboba (commented):
Makes sense to add a `type` attribute. Might be worth noting that the value will always be `keyframe`, right?

PR #475, opened 2022-03-21
Describe that WG consensus is needed to add registry entries

chrisn:
This is to match expected practice in the WG, and the process we're following for reviewing the proposed H.263 registration.


chrisn:
I see some validation errors, but they're unrelated to the change in this PR.

tidoust:
> I see some validation errors, but they're unrelated to the change in this PR.

There seems to have been an update in how Bikeshed interprets Notes that appear within `<dl>` tags. Bikeshed used to automatically tie notes with the previous `<dd>` but no longer fixes that kind of constructs, resulting in invalid HTML markup. I filed #476 and #477 to fix these errors.

chcunningham (approved):
LGTM. I'll leave it to you / @tidoust to merge in whatever order you like (considering your other outstanding PR).

PR #476, opened 2022-03-21
[OPUS registration] Fix typo

tidoust:
Note the typo now also triggers HTML markup validation errors (Bikeshed used to fix things automatically but no longer does)


tidoust:
Note the CI failure on the main spec is expected. #477 would fix it.

chrisn:
Thanks! I'll merge, as this is just a typo fix.

PR #480, opened 2022-04-08
Update capture-to-file sample to use Workers

chcunningham:
Also deletes the webcam-in-worker sample, as that is now redundant.


samples/capture-to-file/encode-worker.js
```diff
@@ -0,0 +1,86 @@
...
+
+  frameReader = frameStream.getReader();
+
+  const init = {
+    output: (chunk) => {
+      webmWriter.addFrame(chunk);
+    },
+    error: (e) => {
+      console.log(e.message);
+      vtr.stop();
```
tguilbert-google, line 26: Remove `vtr.stop()` :)

Is there extra cleanup needed for the `frameStream` or `frameReader`?

samples/capture-to-file/encode-worker.js
```diff
@@ -0,0 +1,86 @@
...
+  let support = await VideoEncoder.isConfigSupported(config);
+  console.assert(support.supported);
+  encoder.configure(config);
+
+  frameReader.read().then(async function processFrame({done, value}) {
+    let frame = value;
+
+    if(done) {
+      await encoder.flush();
+      encoder.close();
```
tguilbert-google, line 47: Does this also need a call to `frameReader.releaseLock()` or `frameReader.cancel()`?

samples/capture-to-file/capture-to-file.html
```diff
@@ -1,133 +1,96 @@
...
+<script>
+    let video = document.getElementById('src');
+    let button = document.getElementById('record');
+    let encodeWorker = null;
+    let stream = null;
+    let videoTrack = null;
+
+    async function startRecording() {
+      console.assert(button.innerText == 'Record');
+      button.disabled = true;
```
tguilbert-google, line 30: I think (but I'm not sure) that disabling the button at the top of the function, and re-enabling it at the bottom is unnecessary, since the style won't be recomputed during that time (e.g. you can do/undo style changes as many times as you want between window.rAF calls, they only will be visible after rAF).

tguilbert-google (approved):

samples/capture-to-file/encode-worker.js
chcunningham: I think no. The trigger for us entering this branch should be that stopRecording() was called higher in the stack, which calls frameReader.cancel(), which sets done = true.

For this demo I think cancel() (vs releaseLock()) is the right call because we have no need to use the readableStream (cancel it) after recording stops. Starting a new recording makes a new worker w/ a new stream.

samples/capture-to-file/capture-to-file.html
chcunningham: In this case the await on l.32 gives the style a chance to be computed. I added this because I found the save-file-picker didn't always get focus when it opened and I was able to press "start" multiple times, putting the code into a weird state.

chcunningham (commented):
Thanks Thomas

chcunningham:
@aboba @padenot going ahead w/ merge since this is just a code-sample refactor. Feedback welcome though.

PR #489, opened 2022-05-07
Fix the problem that getAvccBox cannot get the value when the first track of the video is not AVC

Vinlic:


chcunningham (approved):
Seems ok. It's not super robust for all types of MP4 files (e.g. will break if you have missing boxes), but it's more robust than what we had. LGTM

chcunningham:
@aboba @padenot I'll go ahead and push since this is a 1-liner for the mp4 sample.

chcunningham:
FYI, had to revert since this produced the following error

`Error handling response: TypeError: Cannot read properties of undefined (reading 'filter')`

Feel free to resubmit with a fix.

chcunningham:
Ignore me! The sample wasn't broken. I misread the error. The commit is now re-landed.

PR #493, opened 2022-05-16
Revise processing model, fix queue size attributes

chcunningham:
A few changes
* Updates process model to process control message queue entirely on the main thread. This aligns spec with Chrome, fixing #494.
* Moving ^that processing to the main thread means we need to be more explicit about when the loop is blocked/exited. For example, an implementation may block processing a decode() call when it finds the decoder is saturated by active decodes. Previously we elided description of this blocking behavior because the blocked loop lived on another thread, which was unobservable (beyond the queue size attributes). But with message processing happening entirely on the main (control) thread, we're motivated to be more explicit (we can't block scripts). 
* Named "codec" and "control" threads are removed to be more idiomatic with other W3C specs. WebAudio has named threads, but I've learned that isn't typical. This isn't strictly necessary, but it seems like a smart time to make the change (I'm re-writing all references to these threads anyway).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/pull/493.html" title="Last updated on Jul 2, 2022, 7:53 AM UTC (1012fcd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/493/f0d7a73...1012fcd.html" title="Last updated on Jul 2, 2022, 7:53 AM UTC (1012fcd)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -290,14 +290,6 @@ Each [=codec=] instance has a unique <dfn>codec task source</dfn>. Tasks
 [=queue a task|queued=] from the [=[[codec work queue]]=] to the [=/event loop=]
 will use the [=codec task source=].
 
-DELETE ME {#delete-me}
-----------------------
-These dfns should be removed by the final commit in this PR. I've placed them
-here while I iterate to avoid breaking their references.
-
-<dfn>control thread</dfn>,
-<dfn>codec thread</dfn>
-
 
 AudioDecoder Interface {#audiodecoder-interface}
 ================================================
@@ -4532,6 +4524,19 @@ interface ImageDecoder {
 
 ### Internal Slots ### {#imagedecoder-internal-slots}
 
+: <dfn attribute for=ImageDecoder>[[control message queue]]</dfn>
+:: A [=queue=] of [=control messages=] to be performed upon this [=codec=]
+    instance. See [=[[control message queue]]=].
+
+: <dfn attribute for=ImageDecoder>[[message queue blocked]]</dfn>
+:: A boolean indicating when processing the
+    {{ImageDecoder/[[control message queue]]}} is blocked by a pending
+    [=control message=]. See [=[[message queue blocked]]=].
+
+: <dfn attribute for=ImageDecoder>[[codec work queue]]</dfn>
+:: A [=parallel queue=] used for running parallel steps that reference the
+    {{ImageDecoder/[[codec implementation]]}}. See [=[[codec work queue]]=].
+
 : <dfn attribute for=ImageDecoder>\[[ImageTrackList]]</dfn>
 :: An {{ImageTrackList}} describing the tracks found in
     {{ImageDecoder/[[encoded data]]}}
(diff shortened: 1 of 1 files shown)


chcunningham:
@padenot @aboba - review guide:
* start with #494 
* then read PR comment above

This first commit revises the processing model and updates AudioDecoder.configure() and AudioDecoder.decode() to use the new way. Will go further after getting some initial feedback.

chcunningham:
@sandersdan FYI

aboba:
A few questions: 

1. In associated text (though not in the PR itself), there is a reference to the "main thread".  Is it true that the "control thread" is the thread on which decoder or encoder was constructed?  For example, in a worker, the control thread would be the worker, not the main thread, correct? 

2. Is there an attribute that indicates the maximum encodeQueueSize/decodeQueueSize, after which encode/decode operations would block?

3. Are there situations other than going over the maximum queue sizes (such as postMessaging of VideoFrames) where encode/decode methods would block?  For example there are reports of [encode operations taking 30 ms](https://bugs.chromium.org/p/chromium/issues/detail?id=1329036) which so far I haven't been able to reproduce.

chcunningham:
@aboba @padenot, PTAL

aboba (approved):

chcunningham:
@padenot has gone OOO for a bit, so planning to TBR for his review (he's already blessed the first round of change for AudioDecoder). 

@dalecurtis, could you take a look at the ImageDecoder changes here. Should be no change at all to Chrome's behavior - just editorial refactoring to align spec:chrome.  Happy to go through it on VC if you'd like.

index.src.html
```diff
@@ -4344,6 +4525,19 @@
 
 ### Internal Slots ### {#imagedecoder-internal-slots}
 
+: <dfn attribute for=ImageDecoder>[[control message queue]]</dfn>
+:: A [=queue=] of [=control messages=] to be performed upon this [=codec=]
+    instance. See [=[[control message queue]]=].
+
+: <dfn attribute for=ImageDecoder>[[message queue blocked]]</dfn>
+:: A boolean indicating when processing the
```
dalecurtis, line 4533: Currently there are no operations which can block the work queue for ImageDecoder. The only things close are reset() and track changes, both of which will abort the queue.

index.src.html
```diff
@@ -4870,7 +5073,7 @@
 6. Return `true`.
 
 A <dfn>valid image MIME type</dfn> is a string that is a [=valid MIME type
-string=] and for which the `type`, per Section 3.1.1.1 of [[RFC7231]], is
+string=] and for which the `type`, per Section 3.1.1.1 of [[RFC9110]], is
```
dalecurtis, line 5076: Was this a typo before?

dalecurtis (approved):
Seems to make things more complicated than necessary for ImageDecoder, but it's probably better to be consistent throughout the document on the processing model.

index.src.html
chcunningham: The constructor has an async step to configure the decoder, similar to how configure() works for the other interfaces. This gets queued as a control message and blocks any subsequent decode() until configure resolves. It may be that our impl is actually simpler than this, and I don't know that the blockage is really observable anyway, but it seemed nice to be consistent w/ the other configure() control messages in this way.

index.src.html
chcunningham: Bikeshed let me know that the original RFC was obsoleted by this new one.

chcunningham (commented):
Thanks @dalecurtis.

I definitely agree w/ your sentiment that this is more complicated than I'd make it were ImageDecoder a standalone thing.

chcunningham:
Thanks everyone. @padenot, I'll go ahead and merge as TBR while you're OOO. Late feedback welcome.

PR #534, opened 2022-08-18
Remove unused property from example code

chrisguttandin:
This PR just removes a property from the `AudioSink` in the example code. I believe it's not in use anywhere.


chcunningham (approved):
Thanks!

PR #535, opened 2022-08-19
Codec registry: escape "*" in codec strings

tidoust:
As the spec uses Markdown, some of the `*` characters were interpreted by Bikeshed as a request to turn the string in italics, whereas we actually want to render the asterisk characters. For instance, `hev1.*, hvc1.*` was rendered as `hev1. <em>, hvc1.<em>`.

This update escapes the asterisk character in codec strings.


chcunningham (approved):
LGTM

PR #548, opened 2022-09-03
Add new pull demuxer interface. Inject demuxers.

chcunningham:
Rather than have the renderer create a specific demuxer, let's inject a
demuxer that uses a common interface. This allows us to re-use the same
renderer with the a variety of demuxers (including the forthcoming
FFmpeg based demuxer).


samples/audio-video-player/mp4_pull_demuxer.js
```diff
@@ -1,3 +1,5 @@
+import {PullDemuxerBase, AUDIO_STREAM_TYPE, VIDEO_STREAM_TYPE} from '../library/pull_demuxer_base.js'
```
sandersdan, line 1: Does this not run in a worker?

sandersdan (approved):

samples/audio-video-player/mp4_pull_demuxer.js
chcunningham: It does run in a worker (samples/audio-video-player/media_worker.js). Am I missing something?

chcunningham: Ah, I guess you were referring to this comment
https://github.com/w3c/webcodecs/pull/549#discussion_r968914714

I'll answer there.

PR #551, opened 2022-09-05
Add frameDuration attribute to OpusEncoderConfig

### LINKED ISSUE #371: AudioEncoderConfig.latencyMode (or similar)
VideoEncoderConfig has latencyMode (https://github.com/w3c/webcodecs/issues/269).

AudioEncoder has similar concerns around latency (realtime) vs quality. For example, Chrome's opus encoder currently uses 60ms buffers ([framesize](https://wiki.xiph.org/Opus_Recommended_Settings#Framesize_Tweaking)) to maximize quality. This is in keeping with our default philosophy of "don't constrain features/quality unless the user asks for it". But we should offer realtime users a way to choose a lower-latency value. 

An alternative would be to specify a more granular knob for just framesize. Or do both.

bdrtc:
This fixes #371

***
<a href="https://pr-preview.s3.amazonaws.com/bdrtc/webcodecs/pull/551.html" title="Last updated on Sep 26, 2022, 2:15 AM UTC (67cc867)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/551/1db4c21...bdrtc:67cc867.html" title="Last updated on Sep 26, 2022, 2:15 AM UTC (67cc867)">Diff</a>

### DIFF
--- a/opus_codec_registration.src.html
+++ b/opus_codec_registration.src.html
@@ -131,6 +131,7 @@ OpusEncoderConfig {#opus-encoder-config}
 <xmp>
 dictionary OpusEncoderConfig {
   OpusBitstreamFormat format = "opus";
+  [EnforceRange] unsigned long frameDuration;
 };
 </xmp>
 </pre>
@@ -140,6 +141,10 @@ dictionary OpusEncoderConfig {
   <dd>
     Configures the format of output {{EncodedAudioChunk}}s. See
     {{OpusBitstreamFormat}}.
+  </dd>
+  <dt><dfn dict-member for=OpusEncoderConfig>frameDuration</dfn></dt>
+  <dd>
+    Configures the frame duration, in milliseconds, of output {{EncodedAudioChunk}}s.
   </dd>
 </dl>
 
--- a/opus_codec_registration.src.html
+++ b/opus_codec_registration.src.html
@@ -105,7 +105,7 @@ Opus is always "[=EncodedAudioChunkType/key=]".
 NOTE: Once the initialization has succeeded, any packet can be decoded at any
 time without error, but this might not result in the expected audio output.
 
-AudiEncoderConfig extensions {#audioencoderconfig-extensions}
+AudioEncoderConfig extensions {#audioencoderconfig-extensions}
 =============================================================
 
 <pre class='idl'>
@@ -131,10 +131,16 @@ OpusEncoderConfig {#opus-encoder-config}
 <xmp>
 dictionary OpusEncoderConfig {
   OpusBitstreamFormat format = "opus";
-  [EnforceRange] unsigned long long frameDuration;
+  [EnforceRange] unsigned long frameDuration;
 };
 </xmp>
 </pre>
+
+To check if an {{OpusEncoderConfig}} is a valid OpusEncoderConfig,
+run these steps:
+1. If {{OpusEncoderConfig/frameDuration}} is not a valid ptime value,
+    which is described in section 6.1 of [[RFC6381]], return `false`.
+2. Return `true`.
 
 <dl>
   <dt><dfn dict-member for=OpusEncoderConfig>format</dfn></dt>
@@ -144,7 +150,7 @@ dictionary OpusEncoderConfig {
   </dd>
   <dt><dfn dict-member for=OpusEncoderConfig>frameDuration</dfn></dt>
   <dd>
-    Configures the frame duration, in microseconds, of output {{EncodedAudioChunk}}s.
+    Configures the frame duration, in milliseconds, of output {{EncodedAudioChunk}}s.
   </dd>
 </dl>
 
--- a/opus_codec_registration.src.html
+++ b/opus_codec_registration.src.html
@@ -137,7 +137,7 @@ dictionary OpusEncoderConfig {
 </pre>
 
 To check if an {{OpusEncoderConfig}} is valid, run these steps:
-1. If {{OpusEncoderConfig/frameDuration}} is not a valid ptime value,
+1. If {{OpusEncoderConfig/frameDuration}} is not a valid `ptime` value,
     which is described in Section 6.1 of [[RFC7587]], return `false`.
 2. Return `true`.
 
--- a/opus_codec_registration.src.html
+++ b/opus_codec_registration.src.html
@@ -136,10 +136,9 @@ dictionary OpusEncoderConfig {
 </xmp>
 </pre>
 
-To check if an {{OpusEncoderConfig}} is a valid OpusEncoderConfig,
-run these steps:
+To check if an {{OpusEncoderConfig}} is valid, run these steps:
 1. If {{OpusEncoderConfig/frameDuration}} is not a valid ptime value,
-    which is described in section 6.1 of [[RFC6381]], return `false`.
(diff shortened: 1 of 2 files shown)


opus_codec_registration.src.html
```diff
@@ -131,6 +131,7 @@
 <xmp>
 dictionary OpusEncoderConfig {
   OpusBitstreamFormat format = "opus";
+  [EnforceRange] unsigned long frameDuration;
```
dalecurtis, line 134: Should probably be "long long" and in microseconds like all other time fields in WebCodecs.

bdrtc: we do find duration in audioData is in microseconds,  
as described in #371 , 
`frameSize = numberOfChannels*sampleRate * frameDuration / 1000`
most audio encoder use millseconds as frame duration(opus ptime for example),  we can change it to micoseconds, then ,the frameSize Calculation will be:
`frameSize = numberOfChannels*sampleRate * frameDuration / 1000/1000` 

Thanks.

bdrtc: @dalecurtis Do you see the new commit according to your comments? This is my first PR that need some modifications, I'm afraid of missing an important action.

tguilbert-google: We can see the new commits. This issue will be discussed at [TPAC](https://www.w3.org/2022/09/TPAC/) next week, so we might hold off on merging this PR until those discussions happen.

dalecurtis: I defer to Thomas for the final decision, I was just driving by.

tguilbert-google: @aboba, @padenot: any thoughts on microseconds vs milliseconds?

IIUC, the valid `frameDurations` in milliseconds are {2.5, 5, 10, 20, 40, 60} according to [the spec](https://www.rfc-editor.org/rfc/rfc6716#section-2.1.4). Leaving it in microseconds might give the impression of finer grain support.

tguilbert-google:
@aboba, @padenot : does this need editor's approval? If so, please take a look.

padenot:
We probably want to have something to do the error checking/validation/rounding to nearest here. Only a very small set of numbers is valid. What happens if you pass something that Opus can't work with?

bdrtc:
if we want add error checking, it should be run in [audioencoder configure method](https://w3c.github.io/webcodecs/#audioencoder-methods), by checking the [validation of AudioEncodeConfig](https://w3c.github.io/webcodecs/#valid-audioencoderconfig), Add one more check step like this: 
2. if codec is opus and  frameDuration exist, frameDuration is not in {2.5, 5, 10, 20, 40, 60}, return false.  
 what's more, frameDuration type should  be changed to double.  
@tguilbert-google @padenot any more sugguestion about this ?  thanks.

aboba:
I think the allowable values are a bit broader. From RFC 7587 Section 6.1: 

"   ptime:  the preferred duration of media represented by a packet
      (according to Section 6 of [RFC4566]) that a decoder wants to
      receive, in milliseconds rounded up to the next full integer
      value.  Possible values are 3, 5, 10, 20, 40, 60, or an arbitrary
      multiple of an Opus frame size rounded up to the next full integer
      value, up to a maximum value of 120, as defined in Section 4.  If
      no value is specified, the default is 20."

tguilbert-google:
> 2. if codec is opus and frameDuration exist, frameDuration is not in {2.5, 5, 10, 20, 40, 60}, return false.

I think we are trying to avoid codec-specific steps in the main spec. We could instead add a generic validation step referring to the registries, along the lines of "If the current `codec`'s registry defines a `Codec-specific additional configuration validation steps`, run the algorithm defined in the registry. If the return value is ...", and add a section validating `OpusEncoderConfig`.

There is probably better wording though.

padenot:
I agree with @tguilbert-google.

helaozhao:
We add a new commit according to the disscussion above, please take a review. @tguilbert-google

bdrtc:
We add a new commit according to the disscussion above, please take a review. @tguilbert-google

opus_codec_registration.src.html
```diff
@@ -131,16 +131,27 @@
...
 };
 </xmp>
 </pre>
 
+To check if an {{OpusEncoderConfig}} is a valid OpusEncoderConfig,
```
tguilbert-google, line 139: NIT: `<dfn>valid OpusEncoderConfig</dfn>`

tguilbert-google (commented):
Latest commit looks good % comments.

FYI, I will be generally unavailable in the next week and a half.

opus_codec_registration.src.html
```diff
@@ -131,16 +131,27 @@
...
+run these steps:
+1. If {{OpusEncoderConfig/frameDuration}} is not a valid ptime value,
+    which is described in section 6.1 of [[RFC6381]], return `false`.
```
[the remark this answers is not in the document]
helaozhao: Yes, It's a wrong reference. It should be rfc7587.

opus_codec_registration.src.html
bdrtc: Cause there is not a more detail dfn about "valid OpusEncoderConfig", maybe a clever desc is:
`To check if an {{OpusEncoderConfig}} is valid,`
......

opus_codec_registration.src.html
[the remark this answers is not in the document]
bdrtc: I will fix it.

helaozhao:
another commit, please review it.

opus_codec_registration.src.html
```diff
@@ -131,16 +131,26 @@
 <xmp>
 dictionary OpusEncoderConfig {
   OpusBitstreamFormat format = "opus";
+  [EnforceRange] unsigned long frameDuration;
 };
 </xmp>
 </pre>
 
+To check if an {{OpusEncoderConfig}} is valid, run these steps:
+1. If {{OpusEncoderConfig/frameDuration}} is not a valid ptime value,
```
tguilbert-google, line 140: NIT: should ptime be `ptime`? (wrapped with backticks?)

tguilbert-google:
Looks good to me!

I'm leaving editors @aboba and @padenot to do the final approval and merge.

Thanks for the patch!

opus_codec_registration.src.html
bdrtc: Yes, it should be.

bdrtc:
@aboba @padenot any thoughts about new commit?

opus_codec_registration.src.html
aboba: Not sure microseconds are necessary.  ptimes when expressed in ms are rounded up.

opus_codec_registration.src.html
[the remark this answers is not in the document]
aboba: Yes, RFC 7587 Section 6.1 makes more sense.  We typically link to the RFC and include the Section # but not link to it.

aboba (approved):

tguilbert-google:
Assuming no objections from @padenot, I am merging this change. 

Chromium implementation work will be tracked in [crbug.com/1372152](https://crbug.com/1372152)

tguilbert-google:
Thanks for the PR!

PR #559, opened 2022-09-14
[WebCodecs VideoFrame metadata registry] Introduce VideoFrame metadata

youennf:
***
<a href="https://pr-preview.s3.amazonaws.com/youennf/webcodecs/pull/559.html" title="Last updated on Oct 20, 2022, 8:20 AM UTC (8b59018)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/559/938df12...youennf:8b59018.html" title="Last updated on Oct 20, 2022, 8:20 AM UTC (8b59018)">Diff</a>

### DIFF
--- a/.github/workflows/auto-publish.yml
+++ b/.github/workflows/auto-publish.yml
@@ -99,8 +99,8 @@ jobs:
             build_override: |
               status: NOTE-WD
           - source: video_frame_metadata_registry.src.html
-            destination: video_frame_metadata_registry..html
-            echidna_token: ECHIDNA_TOKEN_HEVC_REGISTRATION
+            destination: video_frame_metadata_registry.html
+            echidna_token: ECHIDNA_TOKEN_VIDEOFRAMEMETADATA_REGISTRY
             build_override: |
               status: NOTE-WD
     steps:
--- a/.github/workflows/auto-publish.yml
+++ b/.github/workflows/auto-publish.yml
@@ -102,7 +102,7 @@ jobs:
             destination: video_frame_metadata_registry.html
             echidna_token: ECHIDNA_TOKEN_VIDEOFRAMEMETADATA_REGISTRY
             build_override: |
-              status: NOTE-WD
+              status: DRY
     steps:
       # See doc at https://github.com/actions/checkout#checkout-v2
       - name: Checkout repository
--- a/.github/workflows/auto-publish.yml
+++ b/.github/workflows/auto-publish.yml
@@ -98,6 +98,11 @@ jobs:
             echidna_token: ECHIDNA_TOKEN_HEVC_REGISTRATION
             build_override: |
               status: NOTE-WD
+          - source: video_frame_metadata_registry.src.html
+            destination: video_frame_metadata_registry..html
+            echidna_token: ECHIDNA_TOKEN_HEVC_REGISTRATION
+            build_override: |
+              status: NOTE-WD
     steps:
       # See doc at https://github.com/actions/checkout#checkout-v2
       - name: Checkout repository
--- a/index.src.html
+++ b/index.src.html
@@ -3855,7 +3855,7 @@ A <dfn>computed plane layout</dfn> is a [=struct=] that consists of:
         and {{VideoFrame/[[display height]]}}.
     5. Assign `null` to |frame|'s {{VideoFrame/[[duration]]}} and
         {{VideoFrame/[[timestamp]]}}.
-    6. Assign a new {{VideoFrameMetadata}} to |frame|'s {{VideoFrame/[[metadata]]}}.
+    6. Assign a new {{VideoFrameMetadata}} to |frame|.{{VideoFrame/[[metadata]]}}.
 
 : <dfn for=VideoFrame>Parse VideoFrameCopyToOptions</dfn> (with |options|)
 ::  1. Let |defaultRect| be the result of performing the getter steps for
--- a/index.src.html
+++ b/index.src.html
@@ -3185,7 +3185,7 @@ interface VideoFrame {
   readonly attribute long long? timestamp;          // microseconds
   readonly attribute VideoColorSpace colorSpace;
 
-  VideoFrameMetadata getMetadata();
+  VideoFrameMetadata metadata();
 
   unsigned long allocationSize(
       optional VideoFrameCopyToOptions options = {});
@@ -3291,6 +3291,7 @@ dictionary VideoFrameMetadata {
 
 : <dfn attribute for=VideoFrame>\[[metadata]]</dfn>
 :: The {{VideoFrameMetadata}} associated with this frame.
+
 <div class='note'>
 {{VideoFrameMetadata}} is designed to be extended by other specifications.
 Specifications that do extend {{VideoFrameMetadata}} must make sure that
@@ -3646,7 +3647,7 @@ A <dfn>computed plane layout</dfn> is a [=struct=] that consists of:
 
     When invoked, run the [=Close VideoFrame=] algorithm with [=this=].
 
-: <dfn method for=VideoFrame>getMetadata()</dfn>
+: <dfn method for=VideoFrame>metadata()</dfn>
 :: Gets the {{VideoFrameMetadata}} associated with this frame.
 
     When invoked, return the result of calling [=Copy VideoFrame metadata=]
--- a/index.src.html
+++ b/index.src.html
@@ -100,9 +100,8 @@ spec: css-images-3; urlPrefix: https://www.w3.org/TR/css-images-3/
 spec: webrtc-svc; urlPrefix: https://w3c.github.io/webrtc-svc/
     type: dfn; text: scalability mode identifier; url:#scalabilitymodes*
 
-spec: webcodecs-videoframemetadata-registry; urlPrefix: https://w3c.github.io/webcodecs/videoframemetadata_registry.html
-    type: dictionary
-        text: VideoFrameMetadata; url: dictdef-videoframemetadata
+spec: webcodecs-video-frame-metadata-registry; urlPrefix: https://w3c.github.io/webcodecs/video-frame-metadata-registry.html
+    type: dictionary; text: VideoFrameMetadata; url: dictdef-videoframemetadata
 
 spec: ECMASCRIPT; urlPrefix: https://tc39.es/ecma262/
     type: dfn; text: the current Realm; url: #current-realm
@@ -3292,11 +3291,8 @@ dictionary VideoFrameBufferInit {
 
 : <dfn attribute for=VideoFrame>\[[metadata]]</dfn>
 :: The {{VideoFrameMetadata}} associated with this frame.
-
-<div class='note'>
-{{VideoFrameMetadata}} is defined in [[webcodecs-videoframemetadata-registry]].
-By design, all {{VideoFrameMetadata}} properties must be serializable.
-</div>
+     {{VideoFrameMetadata}} is defined in [[webcodecs-video-frame-metadata-registry]].
+     By design, all {{VideoFrameMetadata}} properties are serializable.
 
 ### Constructors ###{#videoframe-constructors}
 
@@ -3336,7 +3332,7 @@ By design, all {{VideoFrameMetadata}} properties must be serializable.
             {{InvalidStateError}} {{DOMException}}.
         2. Let |currentPlaybackFrame| be the {{VideoFrame}} at the [=current
             playback position=].
-        3. If {{VideoFrameInit/metadata}} does not [=map/exists=] in |init|,
+        3. If {{VideoFrameInit/metadata}} does not [=map/exist=] in |init|,
             assign |currentPlaybackFrame|.{{VideoFrame/[[metadata]]}} to it.
         4. Run the [=VideoFrame/Initialize Frame From Other Frame=] algorithm
             with |init|, |frame|, and |currentPlaybackFrame|.
--- a/index.src.html
+++ b/index.src.html
@@ -3293,7 +3293,7 @@ dictionary VideoFrameMetadata {
 :: The {{VideoFrameMetadata}} associated with this frame.
 <div class='note'>
 {{VideoFrameMetadata}} is designed to be extended by other specifications.
-Specifications that do extend {{VideoFrameMetadata}} need to make sure that
+Specifications that do extend {{VideoFrameMetadata}} must make sure that
 added properties are serializable.
 </div>
 
@@ -3335,7 +3335,9 @@ added properties are serializable.
             {{InvalidStateError}} {{DOMException}}.
         2. Let |currentPlaybackFrame| be the {{VideoFrame}} at the [=current
             playback position=].
-        3. Run the [=VideoFrame/Initialize Frame From Other Frame=] algorithm
+        3. If {{VideoFrameInit/metadata}} does not [=map/exists=] in |init|,
+            assign |currentPlaybackFrame|.{{VideoFrame/[[metadata]]}} to it.
+        4. Run the [=VideoFrame/Initialize Frame From Other Frame=] algorithm
             with |init|, |frame|, and |currentPlaybackFrame|.
 
     - {{HTMLCanvasElement}}
@@ -3445,7 +3447,7 @@ added properties are serializable.
         algorithm, with |colorSpace| and {{VideoFrame/[[format]]}}, to
         {{VideoFrame/[[color space]]}}.
     10. Assign the result of calling [=Copy VideoFrame metadata=]
-        with |init|'s {{VideoFrameInit/metadata}} to {{VideoFrame/[[metadata]]}}.
+        with |init|'s {{VideoFrameInit/metadata}} to |frame|.{{VideoFrame/[[metadata]]}}.
 17. Return |frame|.
 
 ### Attributes ###{#videoframe-attributes}
@@ -3774,7 +3776,7 @@ A <dfn>computed plane layout</dfn> is a [=struct=] that consists of:
         |frame|.{{VideoFrame/timestamp}}.
     11. Assign |format| to |frame|.{{VideoFrame/[[format]]}}.
     12. Assign the result of calling [=Copy VideoFrame metadata=]
-        with |init|'s {{VideoFrameInit/metadata}} to {{VideoFrame/metadata}}.
+        with |init|'s {{VideoFrameInit/metadata}} to |frame|.{{VideoFrame/[[metadata]]}}.
 
 : <dfn for=VideoFrame>Initialize Frame With Resource and Size</dfn> (with
     |init|,  |frame|, |resource|, |width| and |height|)
--- a/index.src.html
+++ b/index.src.html
@@ -100,6 +100,10 @@ spec: css-images-3; urlPrefix: https://www.w3.org/TR/css-images-3/
 spec: webrtc-svc; urlPrefix: https://w3c.github.io/webrtc-svc/
     type: dfn; text: scalability mode identifier; url:#scalabilitymodes*
 
+spec: webcodecs-videoframemetadata-registry; urlPrefix: https://w3c.github.io/webcodecs/videoframemetadata_registry.html
+    type: dictionary
+        text: VideoFrameMetadata; url: dictdef-videoframemetadata
+
 spec: ECMASCRIPT; urlPrefix: https://tc39.es/ecma262/
     type: dfn; text: the current Realm; url: #current-realm
 </pre>
@@ -3232,9 +3236,6 @@ dictionary VideoFrameBufferInit {
 
   VideoColorSpaceInit colorSpace;
 };
-
-dictionary VideoFrameMetadata {
-};
 </xmp>
 
 ### Internal Slots ###{#videoframe-internal-slots}
@@ -3293,9 +3294,8 @@ dictionary VideoFrameMetadata {
 :: The {{VideoFrameMetadata}} associated with this frame.
 
 <div class='note'>
-{{VideoFrameMetadata}} is designed to be extended by other specifications.
-Specifications that do extend {{VideoFrameMetadata}} must make sure that
-added properties are serializable.
+{{VideoFrameMetadata}} is defined in [[webcodecs-videoframemetadata-registry]].
+By design, all {{VideoFrameMetadata}} properties must be serializable.
 </div>
 
 ### Constructors ###{#videoframe-constructors}
--- a/index.src.html
+++ b/index.src.html
@@ -114,6 +114,15 @@ spec: ECMASCRIPT; urlPrefix: https://tc39.es/ecma262/
     "title": "Coding-independent code points for video signal type identification",
     "publisher": "ITU",
     "date": "December 2016"
+  },
+  "WEBCODECS-VIDEO-FRAME-METADATA-REGISTRY": {
+    "href": "https://w3c.github.io/webcodecs/video_frame_metadata_registry.html",
+    "title": "WebCodecs VideoFrame Metadata Registry",
+    "publisher": "W3C",
+    "authors": [
+      "Youenn Fablet"
+    ],
+    "status": "ED"
   }
 }
 </pre>
--- a/index.src.html
+++ b/index.src.html
@@ -99,6 +99,9 @@ spec: css-images-3; urlPrefix: https://www.w3.org/TR/css-images-3/
 
 spec: webrtc-svc; urlPrefix: https://w3c.github.io/webrtc-svc/
     type: dfn; text: scalability mode identifier; url:#scalabilitymodes*
+
+spec: ECMASCRIPT; urlPrefix: https://tc39.es/ecma262/
+    type: dfn; text: the current Realm; url: #current-realm
 </pre>
 
 <pre class='biblio'>
@@ -3182,6 +3185,8 @@ interface VideoFrame {
   readonly attribute long long? timestamp;          // microseconds
   readonly attribute VideoColorSpace colorSpace;
 
+  VideoFrameMetadata getMetadata();
+
   unsigned long allocationSize(
       optional VideoFrameCopyToOptions options = {});
   Promise<sequence<PlaneLayout>> copyTo(
@@ -3204,6 +3209,8 @@ dictionary VideoFrameInit {
   // Default matches image unless visibleRect is provided.
   [EnforceRange] unsigned long displayWidth;
   [EnforceRange] unsigned long displayHeight;
+
+  VideoFrameMetadata metadata;
 };
 
 dictionary VideoFrameBufferInit {
@@ -3224,6 +3231,9 @@ dictionary VideoFrameBufferInit {
   [EnforceRange] unsigned long displayHeight;
 
   VideoColorSpaceInit colorSpace;
+};
+
+dictionary VideoFrameMetadata {
 };
 </xmp>
 
@@ -3278,6 +3288,14 @@ dictionary VideoFrameBufferInit {
 
 : <dfn attribute for=VideoFrame>[[color space]]</dfn>
 :: The {{VideoColorSpace}} associated with this frame.
+
+: <dfn attribute for=VideoFrame>\[[metadata]]</dfn>
+:: The {{VideoFrameMetadata}} associated with this frame.
+<div class='note'>
+{{VideoFrameMetadata}} is designed to be extended by other specifications.
+Specifications that do extend {{VideoFrameMetadata}} need to make sure that
+added properties are serializable.
+</div>
 
 ### Constructors ###{#videoframe-constructors}
 
@@ -3426,6 +3444,8 @@ dictionary VideoFrameBufferInit {
     9. Assign the result of running the [=VideoFrame/Pick Color Space=]
         algorithm, with |colorSpace| and {{VideoFrame/[[format]]}}, to
         {{VideoFrame/[[color space]]}}.
+    10. Assign the result of calling [=Copy VideoFrame metadata=]
+        with |init|'s {{VideoFrameInit/metadata}} to {{VideoFrame/[[metadata]]}}.
 17. Return |frame|.
 
 ### Attributes ###{#videoframe-attributes}
@@ -3624,6 +3644,12 @@ A <dfn>computed plane layout</dfn> is a [=struct=] that consists of:
 
     When invoked, run the [=Close VideoFrame=] algorithm with [=this=].
 
+: <dfn method for=VideoFrame>getMetadata()</dfn>
+:: Gets the {{VideoFrameMetadata}} associated with this frame.
+
+    When invoked, return the result of calling [=Copy VideoFrame metadata=]
+    with [=this=] {{VideoFrame/[[metadata]]}}.
+
 ### Algorithms ###{#videoframe-algorithms}
 : <dfn>Create a VideoFrame</dfn> (with |output|, |timestamp|, |duration|, |displayAspectWidth|, |displayAspectHeight|, and |colorSpace|)
 :: 1. Let |frame| be a new {{VideoFrame}}, constructed as follows:
@@ -3747,6 +3773,8 @@ A <dfn>computed plane layout</dfn> is a [=struct=] that consists of:
         |otherFrame|.{{VideoFrame/timestamp}} to
         |frame|.{{VideoFrame/timestamp}}.
     11. Assign |format| to |frame|.{{VideoFrame/[[format]]}}.
+    12. Assign the result of calling [=Copy VideoFrame metadata=]
+        with |init|'s {{VideoFrameInit/metadata}} to {{VideoFrame/metadata}}.
 
 : <dfn for=VideoFrame>Initialize Frame With Resource and Size</dfn> (with
     |init|,  |frame|, |resource|, |width| and |height|)
@@ -3825,6 +3853,7 @@ A <dfn>computed plane layout</dfn> is a [=struct=] that consists of:
         and {{VideoFrame/[[display height]]}}.
     5. Assign `null` to |frame|'s {{VideoFrame/[[duration]]}} and
         {{VideoFrame/[[timestamp]]}}.
+    6. Assign a new {{VideoFrameMetadata}} to |frame|'s {{VideoFrame/[[metadata]]}}.
 
 : <dfn for=VideoFrame>Parse VideoFrameCopyToOptions</dfn> (with |options|)
 ::  1. Let |defaultRect| be the result of performing the getter steps for
@@ -3997,6 +4026,13 @@ A <dfn>computed plane layout</dfn> is a [=struct=] that consists of:
         2. Assign |minAllocationSize| to
             [=combined buffer layout/allocationSize=].
     9. Return |combinedLayout|.
+
+: <dfn for=VideoFrame>Copy VideoFrame metadata</dfn> (with |metadata|)
+:: 1. Let |metadataCopySerialized| be [$StructuredSerialize$](|metadata|).
+    2. Let |metadataCopy| be [$StructuredDeserialize$](|metadataCopySerialized|, [=the current Realm=]).
+    3. return |metadataCopy|.
+
+The goal of this algorithm is to ensure that metadata owned by a {{VideoFrame}} is immutable.
 
 ### Transfer and Serialization ###{#videoframe-transfer-serialization}
 
--- a/index.src.html
+++ b/index.src.html
@@ -3650,8 +3650,11 @@ A <dfn>computed plane layout</dfn> is a [=struct=] that consists of:
 : <dfn method for=VideoFrame>metadata()</dfn>
 :: Gets the {{VideoFrameMetadata}} associated with this frame.
 
-    When invoked, return the result of calling [=Copy VideoFrame metadata=]
-    with [=this=] {{VideoFrame/[[metadata]]}}.
+    When invoked, run these steps:
+    1. If {{platform object/[[Detached]]}} is <code>true</code>,
+        throw an {{InvalidStateError}} {{DOMException}}.
+    2. Return the result of calling [=Copy VideoFrame metadata=]
+        with {{VideoFrame/[[metadata]]}}.
 
 ### Algorithms ###{#videoframe-algorithms}
 : <dfn>Create a VideoFrame</dfn> (with |output|, |timestamp|, |duration|, |displayAspectWidth|, |displayAspectHeight|, and |colorSpace|)
--- a/video_frame_metadata_registry.src.html
+++ b/video_frame_metadata_registry.src.html
@@ -30,12 +30,12 @@ spec: WEBCODECS; urlPrefix: https://w3c.github.io/webcodecs/#
 Registration Entry Requirements {#registration-entry-requirements}
 ==================================================================
 
-A registration entry is a document describing one or several metadata entry.
-A registrating entry has the following requirements:
+A registration entry is a document describing one or several metadata entry,
+and has the following requirements:
 1. Each metadata entry is defined as a separate {{VideoFrameMetadata}}
     dictionary member.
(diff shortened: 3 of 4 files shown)


youennf:
@sandersdan FYI.

index.src.html
```diff
@@ -3182,6 +3186,8 @@
   readonly attribute long long? timestamp;          // microseconds
   readonly attribute VideoColorSpace colorSpace;
 
+  VideoFrameMetadata getMetadata();
```
sandersdan, line 3189: Should this perhaps be an attribute? A method may be safer since we can always return a new copy and then mutations won't be a risk, but an attribute seems more natural to me.

index.src.html
```diff
@@ -3624,6 +3646,11 @@
 
     When invoked, run the [=Close VideoFrame=] algorithm with [=this=].
 
+: <dfn method for=VideoFrame>getMetadata()</dfn>
+:: Gets the {{VideoFrameMetadata}} associated with this frame.
+
+    When invoked, return [=this=] {{VideoFrame/[[metadata]]}}.
```
sandersdan, line 3652: Per above, this may need to be a new copy each time, unless we deeply freeze the deserialized data.

index.src.html
```diff
@@ -3825,6 +3852,8 @@
         and {{VideoFrame/[[display height]]}}.
     5. Assign `null` to |frame|'s {{VideoFrame/[[duration]]}} and
         {{VideoFrame/[[timestamp]]}}.
+    6. Assign `null` to |frame|'s {{VideoFrame/[[metadata]]}}.
```
sandersdan, line 3855: Probably should be an empty dictionary, since it's being returned directly?

index.src.html
```diff
@@ -3998,19 +4027,29 @@
             [=combined buffer layout/allocationSize=].
     9. Return |combinedLayout|.
 
+
+: <dfn for=VideoFrame>Copy VideoFrame metadata</dfn> (with |metadata|)
+:: 1. Let |metadataCopySerialized| be [$StructuredSerialize$](|metadata|).
```
sandersdan, line 4032: I still think we want `forStorage` here so that the resulting serialized data can be held long-term in an opaque format (I'm imagining that Chrome's implementation would want to hold the data in the underlying resource, so that it can flow through the media engine). Do you happen to know which WG we can ask for advice on that?

index.src.html
```diff
@@ -3998,19 +4027,29 @@
...
+    2. Let |metadataCopy| be [$StructuredDeserialize$](|metadataCopySerialized|, [=the current Realm=]).
+    3. Let |userMetadataCopy| be |metadataCopy|'s {{VideoFrameMetadata/user}}.
+    4. Call [=SetIntegrityLevel=] with |userMetadataCopy| and frozen.
```
sandersdan, line 4035: As far as I can tell, this is a shallow freeze. Is there maybe way to get a deeply frozen deserialization?

index.src.html
dalecurtis: I assumed each frame would create an immutable copy, so this could be a SameObject readonly field. If that's true then an attribute sounds preferable to me for consistency of the rest of the fields.

youennf: VideoFrameMetadata is a dictionary right now.
I guess we could make it an interface to put it as an attribute.

index.src.html
youennf: The idea is to copy the data at constructor time and deep freeze it if we can.
Otherwise, agreed we need to make a copy.

index.src.html
youennf: I used `null` for that purpose but it is clearer to just set it to a new VideoFrameMetadata.

index.src.html
youennf: Is there a use case for actually storing VideoFrame in things like IDB?

I thought of a potential use case where you could assign some serialisable but not forStorage objects (like a CropTarget) to a VideoFrame metadata.

index.src.html
youennf: This is the best I found so far in terms of link to express the idea.
Agreed though that we should have a deep freeze if we can.
I plan to investigate it but if you already know this/have pointers, let me know.

sandersdan:
We just discussed this among the Chrome media team, and it turns out we're okay leaving out the `user` metadata. This would allow us to sidestep most of the questions about serialization.

The reasoning here is that if we're not storing metadata also on chunks, and we're not copying metadata between chunks and frames, then users will have to map their own metadata to frames separately anyway (presumably by timestamp). Therefore the `user` metadata is just a convenience rather than a critical feature in the short term.

index.src.html
sandersdan: Are both dictionaries and interfaces things that can have `partial` implementations?

youennf:
> it turns out we're okay leaving out the `user` metadata.

FWIW, one use case I can see is the AR/VR use case where metadata is retrieved by a worker through WebRTC encoded transform, then metadata is passing through decoder to MediaStreamTrack VideoFrames. In that case there might be two contexts so it seems nice to attach application metadata to VideoFrames directly.

That said, I am more than happy to leave this to another day/another PR.

index.src.html
youennf: Yes, they can both be `partial`.

index.src.html
youennf: FWIW, VideoFrame current algorithm does not allow serialisation for IDB (it throws when forStorage is true).

youennf:
I removed user metadata from this PR.
I kept dictionary/method for now.
There is a slight benefit for interface/attribute as we would return the same object every time, which might be handy if the metadata becomes a potentially big object.

index.src.html
youennf: I added a copy.

index.src.html
youennf: Done

index.src.html
youennf: Removed user metadata for now

index.src.html
```diff
@@ -3279,6 +3289,14 @@
 : <dfn attribute for=VideoFrame>[[color space]]</dfn>
 :: The {{VideoColorSpace}} associated with this frame.
 
+: <dfn attribute for=VideoFrame>\[[metadata]]</dfn>
```
sandersdan, line 3292: Is this `\` correct here?

youennf: Yes, otherwise, we do not get link. This is done for duration for instance.

index.src.html
```diff
@@ -3279,6 +3289,14 @@
...
+:: The {{VideoFrameMetadata}} associated with this frame.
+<div class='note'>
+{{VideoFrameMetadata}} is designed to be extended by other specifications.
+Specifications that do extend {{VideoFrameMetadata}} need to make sure that
```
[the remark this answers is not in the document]
youennf: OK

index.src.html
```diff
@@ -3426,6 +3444,8 @@
     9. Assign the result of running the [=VideoFrame/Pick Color Space=]
         algorithm, with |colorSpace| and {{VideoFrame/[[format]]}}, to
         {{VideoFrame/[[color space]]}}.
+    10. Assign the result of calling [=Copy VideoFrame metadata=]
```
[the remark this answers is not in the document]
youennf: Good idea, I'll add it.

youennf:
One potential downsides of using a VideoFrameMetadata interface is that we would probably anyway need a VideoFrameMetadataInit dictionary.
Anytime a spec would extend VideoFrameMetadata with a new readonly attribute, it would also have to extend VideoFrameMetadataInit.

youennf:
As a side note, VideoFrameMetadata name (even though dictionary names are not that important) conflicts with https://wicg.github.io/video-rvfc/#dictdef-videoframemetadata.
@tguilbert-google FYI.

tguilbert-google:
I'd be fine with renaming the rVFC `VideoFrameMetadata` to something else, to avoid collision. I think that `VideoFrameMetadata` makes more sense here.

Albeit verbose, rVFC `VideoFrameMetadata` could be renamed to `VideoFrameCallbackMetadata`?

index.src.html
sandersdan: I don't think it's going to be possible to deeply freeze deserialized data, because for example `ArrayBuffer` is always mutable. Given that, I have become supportive of staying with a method and deserializing a new copy each time it is called.

youennf: Agreed, let's mark this conversation as resolved.

youennf:
> rVFC VideoFrameMetadata could be renamed to VideoFrameCallbackMetadata?

Sounds good, dictionary names are not exposed to JS so verbosity is not a big issue anyway.

index.src.html
```diff
@@ -3182,6 +3185,8 @@
   readonly attribute long long? timestamp;          // microseconds
   readonly attribute VideoColorSpace colorSpace;
 
+  VideoFrameMetadata getMetadata();
```
sandersdan, line 3188: Nit: consider just `metadata()` for consistency with `allocationSize()`.

sandersdan (approved):
Approach LGTM. The editors are more qualified to review the text of the changes.

index.src.html
dalecurtis: Let's go with metadata() for consistency.

index.src.html
```diff
@@ -3279,6 +3289,14 @@
 : <dfn attribute for=VideoFrame>[[color space]]</dfn>
 :: The {{VideoColorSpace}} associated with this frame.
 
+: <dfn attribute for=VideoFrame>\[[metadata]]</dfn>
+:: The {{VideoFrameMetadata}} associated with this frame.
+<div class='note'>
```
dalecurtis, line 3294: Needs indent - see PR preview failures.

youennf: Done

index.src.html
youennf: Done. It is a bit strange we do not have consistency across specs here.

index.src.html
```diff
@@ -3624,6 +3647,12 @@
 
     When invoked, run the [=Close VideoFrame=] algorithm with [=this=].
 
+: <dfn method for=VideoFrame>metadata()</dfn>
+:: Gets the {{VideoFrameMetadata}} associated with this frame.
```
dalecurtis, line 3651: Is there a reason you went with this approach, versus the exception that all the other methods throw when closed? I think it's more error prone to return an empty/fresh structure versus throwing an exception.

I think you should instead throw an exception and assign null during close() algorithm unless there's a good reason otherwise.

index.src.html
```diff
@@ -3998,6 +4030,13 @@
             [=combined buffer layout/allocationSize=].
     9. Return |combinedLayout|.
 
+: <dfn for=VideoFrame>Copy VideoFrame metadata</dfn> (with |metadata|)
+:: 1. Let |metadataCopySerialized| be [$StructuredSerialize$](|metadata|).
```
dalecurtis, line 4034: StructuredSerialize/StructuredDeserialize don't seem to linkify in the preview. Do we need some other tagging somewhere to make that happen?

index.src.html
youennf: Right, now that we moved from an attribute to a method, it makes sense to throw for consistency.

youennf: Done

index.src.html
youennf: It linkifies locally for me when using https://api.csswg.org/bikeshed/.
I am not sure why it does not do this in the preview.

youennf: The links do work in [the diff preview link](https://pr-preview.s3.amazonaws.com/w3c/webcodecs/559/938df12...youennf:673a0fa.html).

youennf: Seems to work now, I am not sure why

dalecurtis:
Thanks, lgtm. @aboba @padenot any objections?

padenot (commented):
Generally looks good, but we need to figure out how we're going to use this. What's the first user here?

Monkey patching can be annoying. Maybe it's going to be less annoying if it's simply a dict, since it's somewhat less structured, but it would be easier to just ask all users to modify this text.

index.src.html
```diff
@@ -3279,6 +3289,15 @@
 : <dfn attribute for=VideoFrame>[[color space]]</dfn>
 :: The {{VideoColorSpace}} associated with this frame.
 
+: <dfn attribute for=VideoFrame>\[[metadata]]</dfn>
+:: The {{VideoFrameMetadata}} associated with this frame.
+
+<div class='note'>
+{{VideoFrameMetadata}} is designed to be extended by other specifications.
+Specifications that do extend {{VideoFrameMetadata}} must make sure that
+added properties are serializable.
```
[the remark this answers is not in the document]
youennf: It might be worth centralising these in the long run but partial dictionaries are pretty nice in terms of modularity.
First user might be FaceDetection related metadata, see https://github.com/w3c/mediacapture-extensions/pull/48.
I think it is worth keeping it in a separate spec until the whole proposal migrates to media capture-main, WebCodecs and HTML if there is a need for rvfc..

I could see additional metadata like the ones exposed by rvfc (rtpTimestamp, receiveTime, captureTime) be candidate as well. These ones could go to WebCodecs directly and we could try to have rvfc metadata dictionary extend WebCodecs VideoFrameMetadata to not duplicate the fields and descriptions.

Also, the fact that properties need to be serializable is exercised when calling metadata(). In theory, implementors would catch such issues if spec authors do not.

youennf:
> What's the first user here?

First user might be https://github.com/w3c/mediacapture-extensions/pull/48
rvfc exposed metadata might also be good candidate as well (say you are exposing VideoFrames coming from a peer connection track).

> Monkey patching can be annoying.

With partial dictionaries, I do not expect a lot of monkey patching.

index.src.html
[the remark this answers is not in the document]
aboba: I agree with Paul that this is problematic. There needs to be a central place where metadata is defined, or at least collected.  One possibility is to define a registry for metadata.  That registry specification would also define the requirements for additions.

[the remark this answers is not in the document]
youennf: I am ok with this registry approach, or actual users of metadata to write PRs that would add members to this dictionary.
What changes to this PR are you envisioning?
Something like: 
`By design, all {{VideoFrameMetadata}} properties must be serializable.
Specification editors need to ensure this requirement is met whenever extending {{VideoFrameMetadata}}`.

dalecurtis:
Editor's call:
- Prefer registry track approach for central registration and clear implications of effects on WebCodecs API. E.g., face detection metadata has been proposed to influence VideoEncoder. Registry allows review through Media WG. 
- PR looks okay otherwise. Just needs to change text from partial dictionary to registry. @padenot to follow up with registry document / or directions for adding one. Will require WG vote, so @chrisn as an FYI.

padenot:
This is an example of adding a registry document to the repo: https://github.com/w3c/webcodecs/pull/319, but I can also do it.

dalecurtis:
Since it'd be an entire new registry document: https://github.com/w3c/webcodecs/commit/a270f0f3adafcea640cc8b2834eadce4f671a54e is the commit where we added the codec registry for reference.

dalecurtis (approved):
This lgtm, I know @padenot is busy this week, but @aboba or @chrisn did you have any comments?

chrisn (changes requested):
Thanks @dalecurtis. The registry approach overall SGTM, and seems preferable to having to update the main WebCodecs spec for each new kind of metadata.

I can run a call for consensus as soon as this PR is ready. I invite review also from @tidoust.

video_frame_metadata_registry.src.html
```diff
@@ -0,0 +1,61 @@
...
+Registration Entry Requirements {#registration-entry-requirements}
+==================================================================
+
+A registration entry, called metadata, has the following requirements:
+1. Each metadata is defined as a separate {{VideoFrameMetadata}} dictionary member.
+2. Each metadata MUST be serializable.
+3. Each metadata should have clearly defined semantics.
+   In particular, its interactions with the media processing pipeline
+   (encoders, decoders, renderers...) should be well defined.
+
```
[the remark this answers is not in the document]
aboba: Yes. For example, it seems sensible to require a W3C specification, and probably one that has been adopted by a WG.  If there is some expectation of changed behavior in WebCodecs (e.g. an encoder making use of FaceDetection metadata), then the registration should probably be approved by the MEDIA WG.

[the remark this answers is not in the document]
chrisn: I agree, although your comment suggests that another WG could update the registry without needing approval from Media WG, if there weren't any expectations on WebCodecs. It would mean coordinating between multiple WGs, but would we not still want Media WG approval, as publishers of the registry?

.github/workflows/auto-publish.yml
```diff
@@ -98,6 +98,11 @@ jobs:
             echidna_token: ECHIDNA_TOKEN_HEVC_REGISTRATION
             build_override: |
               status: NOTE-WD
+          - source: video_frame_metadata_registry.src.html
+            destination: video_frame_metadata_registry..html
```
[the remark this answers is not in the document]
youennf: OK

video_frame_metadata_registry.src.html
```diff
@@ -0,0 +1,61 @@
...
+Registration Entry Requirements {#registration-entry-requirements}
+==================================================================
+
+A registration entry, called metadata, has the following requirements:
+1. Each metadata is defined as a separate {{VideoFrameMetadata}} dictionary member.
+2. Each metadata MUST be serializable.
+3. Each metadata should have clearly defined semantics.
+   In particular, its interactions with the media processing pipeline
+   (encoders, decoders, renderers...) should be well defined.
```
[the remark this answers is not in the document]
youennf: Fixed

video_frame_metadata_registry.src.html
```diff
@@ -0,0 +1,61 @@
...
+
+VideoFrameMetadata definition
```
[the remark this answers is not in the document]
youennf: Fixed

video_frame_metadata_registry.src.html
```diff
@@ -0,0 +1,61 @@
...
+Level: none
+Group: mediawg
+ED: https://w3c.github.io/webcodecs/videoframemetadata_registry.html
+TR: https://www.w3.org/TR/webcodecs-videoframemetadata-registry/
+Editor:
+
+Boilerplate: omit conformance
+
+Abstract: This registry is intended to enumerate the metadata fields that can be attached
+    to {{VideoFrame}} objects via the {{VideoFrameMetadata}} dictionary..
```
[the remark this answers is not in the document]
youennf: Fixed

index.src.html
```diff
@@ -3279,6 +3290,14 @@
 : <dfn attribute for=VideoFrame>[[color space]]</dfn>
 :: The {{VideoColorSpace}} associated with this frame.
 
+: <dfn attribute for=VideoFrame>\[[metadata]]</dfn>
+:: The {{VideoFrameMetadata}} associated with this frame.
+
+<div class='note'>
+{{VideoFrameMetadata}} is defined in [[webcodecs-videoframemetadata-registry]].
+By design, all {{VideoFrameMetadata}} properties must be serializable.
```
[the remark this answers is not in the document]
youennf: Fixed

index.src.html
```diff
@@ -3317,7 +3336,9 @@
             {{InvalidStateError}} {{DOMException}}.
         2. Let |currentPlaybackFrame| be the {{VideoFrame}} at the [=current
             playback position=].
-        3. Run the [=VideoFrame/Initialize Frame From Other Frame=] algorithm
+        3. If {{VideoFrameInit/metadata}} does not [=map/exists=] in |init|,
```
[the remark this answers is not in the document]
youennf: Fixed

video_frame_metadata_registry.src.html
```diff
@@ -0,0 +1,61 @@
+<pre class='metadata'>
+Title: WebCodecs VideoFrame Metadata Registry
+Repository: w3c/webcodecs
+Status: NOTE-ED
+Shortname: webcodecs-videoframemetadata-registry
+Level: none
+Group: mediawg
+ED: https://w3c.github.io/webcodecs/videoframemetadata_registry.html
+TR: https://www.w3.org/TR/webcodecs-videoframemetadata-registry/
+Editor:
```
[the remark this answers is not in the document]
youennf: I added my name for now.

video_frame_metadata_registry.src.html
```diff
@@ -0,0 +1,61 @@
+<pre class='metadata'>
+Title: WebCodecs VideoFrame Metadata Registry
+Repository: w3c/webcodecs
+Status: NOTE-ED
+Shortname: webcodecs-videoframemetadata-registry
+Level: none
+Group: mediawg
+ED: https://w3c.github.io/webcodecs/videoframemetadata_registry.html
```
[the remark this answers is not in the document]
youennf: I initially went with videoframemetadata since that is related to VideoFrameMetadata.
I changed it though.

video_frame_metadata_registry.src.html
```diff
@@ -0,0 +1,61 @@
+<pre class='metadata'>
+Title: WebCodecs VideoFrame Metadata Registry
+Repository: w3c/webcodecs
+Status: NOTE-ED
```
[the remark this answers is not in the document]
youennf: OK

video_frame_metadata_registry.src.html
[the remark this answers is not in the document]
youennf: I added the lines from the codec registry.

video_frame_metadata_registry.src.html
```diff
@@ -0,0 +1,61 @@
...
+    type: dictionary
+        text: VideoFrame; url: dictdef-videoframe
+</pre>
+
+
+Registration Entry Requirements {#registration-entry-requirements}
+==================================================================
+
+A registration entry, called metadata, has the following requirements:
+1. Each metadata is defined as a separate {{VideoFrameMetadata}} dictionary member.
```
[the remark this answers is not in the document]
youennf: I am not sure we have yet these rules.
For face detection, I would expect one VideoFrameMetadata member being a dictionary.
For some of WebRTC metadata appearing in rvfc, I would expect multiple single value entries.

index.src.html
```diff
@@ -99,6 +99,13 @@
 
 spec: webrtc-svc; urlPrefix: https://w3c.github.io/webrtc-svc/
     type: dfn; text: scalability mode identifier; url:#scalabilitymodes*
+
+spec: webcodecs-videoframemetadata-registry; urlPrefix: https://w3c.github.io/webcodecs/videoframemetadata_registry.html
```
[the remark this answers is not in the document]
youennf: ok

youennf:
I think there is some magic needed somewhere (@tidoust maybe?) to get a proper echidna name.
Hopefully, this would also make the [[webcodecs-video-frame-metadata-registry]] link work in this PR as well (and WebCodecs adding the metadata registry to its document reference list (normative since it defines VideoFrameMetadata?).

video_frame_metadata_registry.src.html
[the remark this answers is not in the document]
aboba: @chrisn By default, it would probably make sense to require MEDIA WG approval for all "top level" entries in the registry. If it chose to do so, the MEDIA WG could delegate authority for a particular "top level" entry (e.g. a dictionary) to another entity.  However, before a registry entry is reviewed by MEDIA WG, there probably needs to be some minimum requirements.  For example, that the registry entry refer to a specification that has achieved some level of consensus. I also think we need to talk about what the requirements are for those submitting registration requests. Do they have to be participants in the W3C, so as to pass the IPR check?

aboba (changes requested):
As it currently stands, the registration policy is what IANA would call "first come, first served". 

I do not think such a policy makes sense, because it would result in a thicket of registry entries that lack documentation, do not represent consensus, and potentially are IPR encumbered.

I'd like to see a registration policy closer to "MEDIA WG consensus required".  If it chose, MEDIA WG could delegate a top level entry to another WG, but even then, consensus of that WG should be required. Also, it seems important to me that the PR pass the IPR check.

video_frame_metadata_registry.src.html
```diff
@@ -0,0 +1,70 @@
...
+2. Each metadata entry MUST be serializable.
+3. Each metadata entry should have clearly defined semantics.
+    In particular, its interactions with the media processing pipeline
+    (encoders, decoders, renderers, etc.) should be well defined.
+4. A candidate registration entry must be announced by filing an issue in the
+    WebCodecs GitHub issue tracker so they can be discussed and evaluated for
+    compliance before being added to the registry. If the Media Working Group
+    reaches consensus to accept the candidate, a pull request should be drafted
+    (either by editors or by the party requesting the candidate registration)
+    to register the candidate.The registry editors will review and merge the
```
youennf, line 46:
```suggestion
    to register the candidate. The registry editors will review and merge the
```

youennf:
> I'd like to see a registration policy closer to "MEDIA WG consensus required".

Maybe it is not clear, but I would expect this to be the current policy. The current wording tells that the PR is merged based on MEDIA WG consensus.
The PR to the registry would update the VideoFrameMetadata WebIDL definition to include the metadata entry(ies).
And the PR would include a link to the document describing each entry in more details.
If more than that is needed, would you be able to provide some suggestions on how to update this PR?

By the way, this raises the question whether this document is a normative reference or not to the main spec.
I would tend to think so given it defines the VideoFrameMetadata WebIDL.

> Also, it seems important to me that the PR pass the IPR check.

We can add such check, I was thinning this was implied by the fact this document is a Media WG document.
Maybe this is not the case given the document is a registry?

.github/workflows/auto-publish.yml
```diff
@@ -98,6 +98,11 @@ jobs:
             echidna_token: ECHIDNA_TOKEN_HEVC_REGISTRATION
             build_override: |
               status: NOTE-WD
+          - source: video_frame_metadata_registry.src.html
+            destination: video_frame_metadata_registry.html
+            echidna_token: ECHIDNA_TOKEN_VIDEOFRAMEMETADATA_REGISTRY
+            build_override: |
+              status: NOTE-WD
```
tidoust, line 105: If we're going to publish this document on the Registry Track, published status will be "Draft Registry"

```suggestion
              status: DRY
```

Note publication to /TR won't work until after the document has been manually published once to /TR. That's normal.

tidoust (commented):
The Process Document defines requirements for the [registry definition](https://www.w3.org/2021/Process-20211102/#reg-def). One of them is:

> Define the policy for changes to existing entries, such as
> - whether entries can be deleted or deprecated
> - whether entries can be changed after being published, and what kinds of changes are allowed
> - whether previously-deleted unique identifiers can be re-used, or are reserved indefinitely 

That is not covered by the current text (and now that I look into it, the WebCodecs Codec Registry does not cover that either).

Other requirements seem to be adequately covered.

tidoust:
> > Also, it seems important to me that the PR pass the IPR check.
> 
> We can add such check, I was thinking this was implied by the fact this document is a Media WG document. Maybe this is not the case given the document is a registry?

As described in the [W3C Process Document](https://www.w3.org/2021/Process-20211102/#reg-pub), registry documents are like notes and are not subject to the W3C Patent Policy:

"A [registry report](https://www.w3.org/2021/Process-20211102/#registry-report) or [registry section](https://www.w3.org/2021/Process-20211102/#registry-section) is purely documentational, is not subject to the W3C Patent Policy, and must not contain any requirements on implementations. For the purposes of the Patent Policy [[PATENT-POLICY]](https://www.w3.org/2021/Process-20211102/#biblio-patent-policy), any [registry section](https://www.w3.org/2021/Process-20211102/#registry-section) in a [Recommendation track](https://www.w3.org/2021/Process-20211102/#w3c-recommendation-track) document is not a normative portion of that specification."

.github/workflows/auto-publish.yml
youennf: I reused the codec registry status in this file. I guess it should be changed for the codec registry then.

tidoust: The codec registry is not yet on the Registry track (first publication was before the introduction of the Registry track in the Process document). The group resolved to move it to the Registry track, so status will indeed soon change.

tidoust:
> I see, I am not sure we should go with a pure registry document like codecs then. My understanding was that VideoFrameMetadata WebIDL would be normative.

Perhaps that what you need is the same approach as for the codec registry: the registry itself is a table that maps an identifier to some spec that normatively defines the IDL.

Codec registrations are published on the Note track mostly because they touch on codecs and it wasn't clear that we would get appropriate IPR commitments. Ideally, they would rather be published as normative documents on the Recommendation track. If the situation is simpler for VideoFrameMetadata, the working group can publish the specs that define the WebIDL on the Recommendation track.

dalecurtis:
@youennf How come you want metadata to be normative? 

My expectation is that UAs can pick and choose from the metadata they want to support. I think requiring media WG consensus for VFM entries strikes a good balance between allowing everything and only allowing use cases where all UAs agree on the underlying feature (E.g., face detection).

Like codecs I can foresee some metadata having patent implications (e.g., specific HDR metadata types), so I think the metadata registry must be non-normative.

aboba:
Agree with @dalecurtis that the metadata registry should be non-normative.  However, if the registry is "specification required", it can provide a table that includes both the IDL for the metadata as well as a specification that normatively defines the IDL. 

This is how IANA tables work; the tables themselves are not normative, it is the specifications that the tables link to that provide the normative language.

youennf:
> @youennf How come you want metadata to be normative?

There is some normative language that needs to exist.
For instance, if a UA exposes a field X within the metadata dictionary, then the field value must have a value of the type given by the registry/metadata document.

> My expectation is that UAs can pick and choose from the metadata they want to support.

Each metadata entry is an optional API somehow. But if UA implements it, it needs to follow the registry and registry metadata documents, so they become normative.
By design, VideoFrameMetada will be designed this way as all its members will be optional.

> if the registry is "specification required", it can provide a table that includes both the IDL for the metadata as well as a link to the specification that normatively defines the IDL.

That works for me.

> 1. Will the specification always have been produced in the W3C?  For example, could the ITU-T, IETF or AoMedia make a request for a registry entry based on one of their specifications?

It makes sense that the registry can link (directly or indirectly) to other documents.
The normal way should be that the registry links to a Media WG metadata document which links to ITU-T, IETF, AOM documents.

> 2\. Could the MEDIA WG delegate authority for a top-level entry to another entity?

This makes sense to me as well.
The registry will link to a document that may define two members, one of which being a simple type and another one being a dictionary whose definition is owned by another WG.

youennf:
@dalecurtis , @aboba, maybe the registry type is the right way.
Somehow, the codec registry is normative in the sense that if 'vp8' is used, then the UA has to follow the VP8 spec.

tidoust:
> I've always assumed implementing whatever thing is listed in a registry was optional, but once a UA decides to implement, the text within the registry entry is normative. Is that not true? @tidoust @chrisn

I would define normative statements in W3C specs as conformance requirements on implementations. Such requirements are subject to the W3C Patent Policy. As raised in a previous comment, the [W3C Process](https://www.w3.org/2021/Process-20211102/#reg-pub) is explicit that a registry "is not subject to the W3C Patent Policy and MUST NOT contain any requirements on implementations".

Registries are more like lookup tables to redirect people to the right spec. As mentioned by @aboba, [IANA protocol registries ](https://www.iana.org/protocols) come to mind as typical examples: most IANA registries are tables where each entry maps an identifier to a name and/or description and to a normative RFC that defines requirements on implementations. Registry entries themselves do not contain normative content.

No document has been published under the Registry track so far (the track was added last year) so it's hard to reflect on experience within W3C. Having Web IDL content definitions directly in a registry document seems wrong to me since Web IDL content is normative in essence. I would rather expect registry entries to link to one or more specs that define the Web IDL and that are published on the Recommendation track.

youennf:
WebRTC-stats might be a good example we could follow:
1. WebRTC-stats defines WebIDL APIs
2. WebRTC stats does not define implementation conformance. See in particular https://w3c.github.io/webrtc-stats/#conformance: `This specification does not define what objects a conforming implementation should generate.`

By default, metadata would be optional to expose/support.
If, in the future, we want to document that some metadata are mandatory to support, the requirements could be added either in this metadata document or in other documents (WebCodecs main spec or others).
For instance, webrtc-pc (or media capture-transform maybe) could make it mandatory to expose any WebRTC specific metadata related to remote tracks.

dalecurtis:
I see. Thanks @tidoust -- maybe a registry of spec links is the best we can do with the registry model then. The WebRTC stats model also seems reasonable, but I have no experience with it so I don't know if it's well received. I'll defer to @aboba and @padenot on that.

I don't think we'd ever add mandatory metadata to VideoFrameMetadata. Mandatory fields should probably be on the VideoFrame object itself (e.g., timestamp, duration, etc). @sandersdan and I were just discussing this for something like `orientation`.

chrisn:
Having mandatory fields on `VideoFrame` itself makes sense to me, not everything needs to be done via registry entries. The registry is helpful when implementation is optional. In this case, I'd suggest adding language to the registry to prevent naming collisions between mandatory fields and fields defined via the registry.

Summarising where I think we've got to:

* The registry itself isn't normative. As @tidoust said, the registry "is not subject to the W3C Patent Policy and MUST NOT contain any requirements on implementations"

* The registry is a mapping from VideoFrameMetadata attributes to the specs that define those attributes. A single metadata spec can define multiple attributes, or a single attribute that contains a sub-object with further attributes, without restriction.

* Registry entries (i.e., individual metadata specs) are W3C specifications on the Recommendation track that place requirements on implementations. Any registry entry is optional to implement, but if implemented must follow the requirements in the metadata spec.

* Metadata specs would normally be published by Media WG, or may be published by another W3C WG (e.g., WebRTC WG). These can then refer to specs from other organisations as needed, such as ITU-T, IETF, AOM.

* Media WG consensus is needed to add a metadata spec to the registry (as a registry entry).

* Similar to https://github.com/w3c/webcodecs/pull/572 (which applies to the codec registry), we'll need to add guidance on deletion or deprecation of registry entries. An initial suggestion:

  *  **whether entries can be deleted or deprecated**: May be deprecated, if the metadata spec somehow ends up being [obsoleted](https://www.w3.org/Consortium/Process/Drafts/#deactivation)?
  *  **whether entries can be changed after being published, and what kinds of changes are allowed**:  Shouldn't be changed after being published?
  *  **whether previously-deleted unique identifiers can be re-used, or are reserved indefinitely**: Reserved indefinitely?

aboba:
If we think about the registry entries as just a member of a table that points to a specification, then it is the specifications that define how registry entries evolve.

For example, a specification that created registry entries might be superceded by a new specification which could include new registry entries, could clarify the meaning of existing entries, or might not use one or more registry entries defined in the former specification.  

In the IETF model, RFCs never disappear once published, so if the new specification does not mention a registry entry created in a previous specification, the entry continues to link to the old specification, which might be deprecated by the new specification.  In this case, the registry entries are not deleted or deprecated; it is the specification that the entries link to that changes status. 

Within the IETF model, a new specification can update the allocation policy of a registry.  If a registry entry is used in existing implementations, typically new specifications will recognize that usage and will not reallocate the entry for another use so as to avoid interoperability problems.  

This suggests that if a metadata spec is [obsoleted](https://www.w3.org/Consortium/Process/Drafts/#deactivation), the successor specification should state dictate what happens to the registry entries. But what if there is no successor specification?  I'd be concerned about situations in which the link that registry entries pointed to disappears entirely, or is updated to point to a new specification that no longer defines the registry entry. 

It seems like a successor specification should be able to clarify the meaning of a registry entry, or deprecate it.  In some limited circumstances it might even make sense to change the meaning entirely (e.g. if all implementations interpreted the entry in a certain way which might not have been obvious).  If the new specification makes the old one disappear, then it is possible that deletion would make sense, if the old entry would otherwise be left linking to something that is no longer available or doesn't define the entry any more.  If the specification that defines a registry entry can be retrievable indefinitely then deletion makes less sense, since the link can just be left as is until it is updated by a new specification.

youennf:
> I don't think we'd ever add mandatory metadata to `VideoFrameMetadata`.

There are two meaning in mandatory metadata:
1. Mandatory metadata in the sense of required dictionary member. I agree we do not want that.
2. Mandatory metadata in the sense it MUST be supported. This would happen if a UA supports the faceDetection MediaStreamTrack constraint. In any case, the actual face detection metadata field remains optional.

Anyway, can somebody summarise what are the actual editorial steps required to merge this PR?
We can always make improvements on the actual framework. Landing this PR would allow to make progress on some of the metadata in parallel to those improvements.

video_frame_metadata_registry.src.html
```diff
@@ -0,0 +1,70 @@
...
+spec: WEBCODECS; urlPrefix: https://w3c.github.io/webcodecs/#
+    type: dictionary
+        text: VideoFrame; url: dictdef-videoframe
+</pre>
+
+
+Registration Entry Requirements {#registration-entry-requirements}
+==================================================================
+
+A registration entry is a document describing one or several metadata entry.
```
chrisn, line 33: Suggest: metadata entries

chrisn (changes requested):
With these changes, I think this is ready to merge. We scan split off the discussion of deletion and deprecation to a new issue.

video_frame_metadata_registry.src.html
```diff
@@ -0,0 +1,70 @@
...
+A registrating entry has the following requirements:
+1. Each metadata entry is defined as a separate {{VideoFrameMetadata}}
+    dictionary member.
+2. Each metadata entry MUST be serializable.
+3. Each metadata entry should have clearly defined semantics.
```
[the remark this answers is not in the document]
chrisn: The other point here is from @aboba: 

> However, before a registry entry is reviewed by MEDIA WG, there probably needs to be some minimum requirements. For example, that the registry entry refer to a specification that has achieved some level of consensus.

This could apply also to the codecs registry, consider also in a subsequent PR?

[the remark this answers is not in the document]
youennf: Done

chrisn (approved):

chrisn:
Thanks @youennf, this looks good. We can discuss procedural requirements in a new issue if you prefer, and merge this one. Once we have those defined, we can run a CfC in the Media WG to actually publish this registry.

youennf:
> We can discuss procedural requirements in a new issue if you prefer, and merge this one.


Right, let's do that in a follow-up.
@aboba, can you review the latest PR?

> Once we have those defined, we can run a CfC in the Media WG to actually publish this registry.

Should we have a registry entry to run the CfC or is it fine to keep it empty?

dalecurtis (approved):
Still lgtm; thanks @youennf

chrisn:
> Should we have a registry entry to run the CfC or is it fine to keep it empty?

I think it's fine to be empty.

I'll raise another PR for the remaining procedural bits. We can probably also drop this item from the Media WG meeting agenda for tomorrow.

chrisn:
@padenot, @aboba, is this OK to merge, and we can raise a separate PR for any remaining procedural requirements?

padenot:
Yes.

youennf:
@aboba, can you review again the PR and approve it if you think this is fine?

aboba (approved):

dalecurtis:
> Still lgtm; thanks @youennf -- though there is some PR build error:
> 
> `FATAL ERROR: Couldn't find 'webcodecs-video-frame-metadata-registry' in bibliography data. `

Bump on this one before submitting though.

tidoust:
> @tidoust, can you shed some light on the issue @dalecurtis is mentioning above?

I pushed an update to add the entry to the local biblio. The entry can be removed when the registry document comes to existence and gets added to official biblio databases.

dalecurtis:
Great, looks like we're all good here. Thanks @youennf!

Issue #384, opened 2021-10-22
HDR support 


ZmGorynych:
Video Color Space Interface currently supports SD (BT.601) and HD (BT.709), but not PQ and HLG. Syntax-wise, it seems that the values are taken from x264 colorimetry command line parameters. I would suggest using values defined in the CICP spec (see H.273 sec 8.1-8.3). Alternatively, the list of x265 colorimetry command-line parameters will contain the necessary human-readable values.

Also, how can we support backwards-compatible HDR dynamic metadata such as DolbyVision profiles 8.x and Samsung HDR10+? In both cases the underlying stream is a perfectly valid HDR10 stream while additional metadata from either one is used to enhance the experience. This way a single stream can be decoded as HDR10, DolbyVision, and HDR10+ by different decoders.

chcunningham:
Re: Canvas HDR, the Color-on-the-web CG is tracking that deliverable [with a strawman here](https://github.com/w3c/ColorWeb-CG/blob/master/hdr_html_canvas_element.md).

Lets use this issue to track the WC spec work. Off the top of my head, we want:
1. Additional color space values
2. Higher bit depth video pixel formats

@sandersdan, what else?

sandersdan:
Those are the most important; we'll also need to update the relevant codec registrations, and verify that the codec and format detection APIs are sufficient.

sandersdan:
# Proposal

This covers the most basic changes required to support HDR and WCG video in WebCodecs. This is becoming feasible to implement in Chrome as the infrastructure for `<canvas>` is becoming available (c.f. https://groups.google.com/a/chromium.org/g/blink-dev/c/S609KGTkSSk/m/1EW4CWn3BgAJ).

## Pixel Formats

- `I420P10`, `I420P12`, `I420AP10`, `I420AP12`, `I422P10`, `I422P12`, `I422A`, `I422AP10`, `I422AP12`, `I444P10`, `I444P12`, `I444A`, `I444AP10`, `I444AP12`
  - These are all the YUV formats supported by libavif that are not currently specified.
- `P010`, `P012`, `P016`
  - These are the NV12-like formats likely to be supported by hardware decoders.
- `RGBAF16`
  - Matches the high-bit-depth `<canvas>` format (RGBA 16-bit float).

The above list excludes:

- `XR30`, `XB30`
  - These formats pack 10-bit RGB samples into four bytes. I didn't see these being used by any codecs, but they are sometimes used as capture formats.

### Alternatively

[...]

PR #566, opened 2022-09-26
Add WCG color spaces.

sandersdan:
This implements the color spaces portion of #384, but does not add pixel formats or HDR metadata.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/pull/566.html" title="Last updated on Sep 26, 2022, 6:31 PM UTC (b8dd618)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/566/1db4c21...b8dd618.html" title="Last updated on Sep 26, 2022, 6:31 PM UTC (b8dd618)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -4461,9 +4461,11 @@ Color primaries describe the color gamut of video samples.
 
 <xmp class='idl'>
 enum VideoColorPrimaries {
-  "bt709",      // BT.709, sRGB
-  "bt470bg",    // BT.601 PAL
-  "smpte170m",  // BT.601 NTSC
+  "bt709",
+  "bt470bg",
+  "smpte170m",
+  "bt2020",
+  "smpte432",
 };
 </xmp>
 
@@ -4483,6 +4485,16 @@ enum VideoColorPrimaries {
     Color primaries used by BT.601 NTSC, as described by [[H.273]]
     section 8.1 table 2 value 6.
   </dd>
+  <dt><dfn enum-value for=VideoColorPrimaries>bt2020</dfn></dt>
+  <dd>
+    Color primaries used by BT.2020 and BT.2100, as described by [[H.273]]
+    section 8.1 table 2 value 9.
+  </dd>
+  <dt><dfn enum-value for=VideoColorPrimaries>smpte432</dfn></dt>
+  <dd>
+    Color primaries used by P3 D65, as described by [[H.273]]
+    section 8.1 table 2 value 12.
+  </dd>
 </dl>
 
 Video Transfer Characteristics {#videotransfercharacteristics}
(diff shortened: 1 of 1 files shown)


index.src.html
```diff
@@ -4492,9 +4504,12 @@
...
 enum VideoTransferCharacteristics {
-  "bt709",         // BT.709
-  "smpte170m",     // BT.601 (functionally the same as bt709)
-  "iec61966-2-1",  // sRGB
+  "bt709",
+  "smpte170m",
+  "iec61966-2-1",
+  "linear",
+  "pq",
+  "hlg",
```
sandersdan, line 4512: The alternative names here are "smpte2084" and "arib-std-b67", which would align with FFmpeg naming.

chrisn:
Should we align these with [Media Capabilities](https://w3c.github.io/media-capabilities/) ColorGamut and TransferFunction? Are there cases where WebCodecs needs a different set of possible values than Media Capabilities?

sandersdan:
> Should we align these with [Media Capabilities](https://w3c.github.io/media-capabilities/) ColorGamut and TransferFunction?

The short answer is that we diverged with the first version of WebCodecs, which we decided was acceptable since WebCodecs does not integrate with MediaCapabilities. Some of the values I picked in this PR do match between the two, at least.

Perhaps more relevant is alignment with https://github.com/w3c/webcodecs/issues/384, but that is not yet finalized and also has a different technical interpretation.

dalecurtis (approved):
Names lgtm since we're using what's named in H.273.

dalecurtis:
Any last comments on these names? It's been 3 weeks, so we'll go ahead and merge tomorrow otherwise.

index.src.html
chrisn: Looks OK as is, but if you're considering changing, I'd suggest referencing BT2100, so something like `bt2100-2-pq` and `bt2100-2-hlg`.

chrisn (approved):
Thanks @dalecurtis, sorry for the delay. I left one minor comment to consider, but otherwise looks good.

index.src.html
chrisn: To clarify, preference is for `pq` and `hlg`

dalecurtis: Ah great :) I think Dan was just playing devils advocate above.

chrisn: I see! My comment was really: if you're going in that direction, prefer BT.2100 rather than the SMPTE and ARIB standards.

aboba (approved):

dalecurtis:
Thanks all. I think we have sufficient approval to merge now.

Issue #527, opened 2022-08-11
add opus inbandfec & dtx configure parameter to webcodecs audio encoder configure


bdrtc:
webcodecs support many audio codecs, some codec has its private encode configure parameters, opus for example, inbandfec、 dtx、complexity is importent for realtime comunication, but we find those paramerter was not described in spec and no brower implement them currently,
we sugguest the spec describe a method for such case.

aboba:
Duplicate of https://github.com/w3c/webcodecs/issues/244

bdrtc:
close this as solved via #594

Issue #588, opened 2022-10-19
framerate vs frameRate


dontcallmedom:
[`VideoEncoderConfig` uses lower-case `framerate`](https://w3c.github.io/webcodecs/#dom-videoencoderconfig-framerate) (as does media capabilities), when [other specs ](https://dontcallmedom.github.io/webdex/f.html#frameRate%40%40screen-capture%25%25dfn)(screen capture, media capture and streams, webxr) use the camel-case `frameRate`.

Aligning would be an improvement from a developer ergonomics perspective

padenot:
Too bad, but it's a bit late at this point indeed. I assume this attribute sees a lot of use out there.

dontcallmedom:
the future being longer than the past, can we move to frameRate and have implementation provides a backwards compat layer?

dalecurtis:
I don't mind accepting both, but defer to @padenot @aboba @youennf.

youennf:
bitrate is a similar field in VideoEncoderConfig and is consistently lower cased in all APIs AFAIK.
Is it fine to have frameRate but bitrate in the same dictionary?

dalecurtis:
We seem to use `bitrate` everywhere AFAICT. `bitrate` does seem to have "formal" recognition from dictionaries though: https://www.merriam-webster.com/dictionary/bit%20rate whereas I can't find an equivalent for `framerate`.

I personally prefer `framerate` and `bitrate` but am always sympathetic to consistency arguments.

dalecurtis:
A similar CamelCase or flatcase issue is coming up with https://github.com/w3c/webcodecs/pull/594

Djuffin:
I don't think we'll be renaming anything at this point. So let's live with what we have.

PR #594, opened 2022-10-20
Add OpusEncoderConfig parameters

### LINKED ISSUE #244: opus configuration (e.g. in-band fec)
in issue #9 @murillo128 requested a way to configure opus in-band fec.

https://www.opus-codec.org/docs/html_api/group__encoderctls.html#ga5b67dc832aa46c1c2f35752c46380545

tguilbert-google:
This PR adds in-band-FEC, DTX and complexity configuration knobs to the Opus encoder.

This should fix #244 and #527

### DIFF
--- a/opus_codec_registration.src.html
+++ b/opus_codec_registration.src.html
@@ -172,7 +172,7 @@ To check if an {{OpusEncoderConfig}} is valid, run these steps:
     Configures the encoder's expected packet loss percentage. The valid range is
     `0` to `100`.
 
-    NOTE: The packet loss percentage might be updated over the course of the
+    NOTE: The packet loss percentage might be updated over the course of an
     encoding, and it is recommended for User Agents to support these reconfigurations.
   </dd>
   <dt><dfn dict-member for=OpusEncoderConfig>useinbandfec</dfn></dt>
--- a/opus_codec_registration.src.html
+++ b/opus_codec_registration.src.html
@@ -131,9 +131,9 @@ OpusEncoderConfig {#opus-encoder-config}
 <xmp>
 dictionary OpusEncoderConfig {
   OpusBitstreamFormat format = "opus";
-  [EnforceRange] unsigned long frameDuration;
+  [EnforceRange] unsigned long frameDuration = 20;
   [EnforceRange] unsigned long complexity;
-  [EnforceRange] unsigned long packetlossperc;
+  [EnforceRange] unsigned long packetlossperc = 0;
   boolean useinbandfec = false;
   boolean usedtx = false;
 };
@@ -158,24 +158,31 @@ To check if an {{OpusEncoderConfig}} is valid, run these steps:
   <dt><dfn dict-member for=OpusEncoderConfig>frameDuration</dfn></dt>
   <dd>
     Configures the frame duration, in milliseconds, of output {{EncodedAudioChunk}}s.
+    If not specified, the default is `20` milliseconds.
   </dd>
   <dt><dfn dict-member for=OpusEncoderConfig>complexity</dfn></dt>
   <dd>
     Configures the encoder's computational complexity. The valid range is `0` to `10`,
-    with `10` representing the highest complexity.
+    with `10` representing the highest complexity. If no value is specificied, the
+    default value is platform-specific: User Agents <em class="rfc2119">SHOULD</em>
+    set a default of `5` for mobile platforms, and a default of `9` for all other platforms.
   </dd>
   <dt><dfn dict-member for=OpusEncoderConfig>packetlossperc</dfn></dt>
   <dd>
     Configures the encoder's expected packet loss percentage. The valid range is
-    `0` to `100`.
+    `0` to `100`. If no value is specified, the default is `0`.
+
+    NOTE: The packet loss percentage might be updated over the course of the encoding, and
+      it is recommended for User Agents to support these reconfigurations.
   </dd>
   <dt><dfn dict-member for=OpusEncoderConfig>useinbandfec</dfn></dt>
   <dd>
-    Configures whether the encoder should generate Opus in-band fec or not.
+    Specifies whether the encoder provides Opus in-band FEC. If no value is specified,
+    the default is `false`.
   </dd>
   <dt><dfn dict-member for=OpusEncoderConfig>usedtx</dfn></dt>
   <dd>
-    Configures whether the encoder should use DTX or not.
+    Specifies if the encoder uses DTX. If no value is specified, the default is `false`.
   </dd>
 </dl>
 
--- a/opus_codec_registration.src.html
+++ b/opus_codec_registration.src.html
@@ -131,7 +131,7 @@ OpusEncoderConfig {#opus-encoder-config}
 <xmp>
 dictionary OpusEncoderConfig {
   OpusBitstreamFormat format = "opus";
-  [EnforceRange] unsigned long frameDuration = 20;
+  [EnforceRange] unsigned long frameDuration;
   [EnforceRange] unsigned long complexity;
   [EnforceRange] unsigned long packetlossperc = 0;
   boolean useinbandfec = false;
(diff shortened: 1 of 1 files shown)


opus_codec_registration.src.html
```diff
@@ -132,13 +132,21 @@
...
+  boolean useinbandfec;
+  boolean usedtx;
 };
 </xmp>
 </pre>
 
 To check if an {{OpusEncoderConfig}} is valid, run these steps:
 1. If {{OpusEncoderConfig/frameDuration}} is not a valid `ptime` value,
     which is described in Section 6.1 of [[RFC7587]], return `false`.
+1. If {{OpusEncoderConfig/complexity}} is specified and not within the range of
```
tguilbert-google, line 146: Since dictionary members are optional by default, is `If {{member}} is specified and ...` to be omitted?

Similar question for specifying the lower bound of `0`, when the `[EnforceRange] unsigned long` guarantees the member cannot be a negative number.

opus_codec_registration.src.html
```diff
@@ -132,13 +132,21 @@
 dictionary OpusEncoderConfig {
   OpusBitstreamFormat format = "opus";
   [EnforceRange] unsigned long frameDuration;
+  [EnforceRange] unsigned long complexity;
+  [EnforceRange] unsigned long packetlossperc;
```
dalecurtis, line 136: We should figure out if we want to be consistent in naming here. `frameDuration` is camel case, it seems like these should be `packetLossPerc`, `useInbandFec`, and `useDtx` -- if we're intentionally not doing that for consistency with webrtc or elsewhere it'd be good to note that in the commit at least.

opus_codec_registration.src.html
dalecurtis: precision never hurts, so I think it's fine to be a bit more verbose.

opus_codec_registration.src.html
```diff
@@ -151,6 +159,24 @@
...
+    with `10` representing the highest complexity.
+  </dd>
+  <dt><dfn dict-member for=OpusEncoderConfig>packetlossperc</dfn></dt>
+  <dd>
+    Configures the encoder's expected packet loss percentage. The valid range is
+    `0` to `100`.
+  </dd>
+  <dt><dfn dict-member for=OpusEncoderConfig>useinbandfec</dfn></dt>
+  <dd>
+    Configures whether the encoder should generate Opus in-band fec or not.
```
dalecurtis, line 174: FEC. Is there a link to the parameter definitions in the opus spec that we can use for each of these?

opus_codec_registration.src.html
tguilbert-google: I went with no caps to be consistent with [section 6.1 of RFC758](https://datatracker.ietf.org/doc/html/rfc7587#section-6.1:~:text=useinbandfec%3A%20%20specifies%20that,default%20is%200.), and the [ORTC spec](https://draft.ortc.org/#opus-codec-settings*).

opus_codec_registration.src.html
tguilbert-google: I think we can link to section 6.1 of RFC7587 for useinbandfec and usedtx parameters. We could additionally link to [section 3.2](https://datatracker.ietf.org/doc/html/rfc7587#section-3.2) for complexity. There packetlossperc isn't defined in RFC7587, so it might be best to link to ORTC for all parameters?

I'm not sure what the "hierarchy" of specs is, and if we can mix & match.

opus_codec_registration.src.html
aboba: We modeled the Opus parameter names in ORTC after RFC 7587 Section 6.1, since that was also used in the WebRTC API SDP.

opus_codec_registration.src.html
aboba: RFC 7587 Section 6.1 defines `useinbandfec` and `usedtx`  in terms of decoder preference rather than an encoder setting. So linking to Section 6.1 might be a bit confusing.   You might define them like this: 

useinbandfec:  specifies whether the encoder provides Opus in-band FEC.  Possible values are 1 (yes) and 0 (no). 
If no value is specified, useinbandfec is assumed to be 0. 

usedtx:  specifies if the encoder uses DTX.  Possible values are 1 (yes) and 0 (no). If no value is specified, the default is 0.

bdrtc: packetlossperc: the packet loss perc parameter could be changed during encoding, this need the OpusEncoderConfig support realime configuration, do we need add describe about this?

aboba: There have been some questions relating to re-configuration so it might be a good idea to clarify that implementations are required to support this. However, the ability to re-configure encoders is not specific to Opus or `packetlossperc`.

bdrtc: > There have been some questions relating to re-configuration so it might be a good idea to clarify that implementations are required to support this. However, the ability to re-configure encoders is not specific to Opus or `packetlossperc`.

agree,  `packetlossperc` should be 0 by default, and user can re-configure it during encoding  in accordance with uplink audio loss rate（ask encoder to generate inbandFec packet)

opus_codec_registration.src.html
aboba: Do you want to define defaults for these as well?

aboba (changes requested):

opus_codec_registration.src.html
bdrtc: `complexity`: the default value are different in pc/mobile platform to save encoder complexity, webrtc use 5 in Android, iOS and/or ARM device, others is 9 by default.

opus_codec_registration.src.html
tguilbert-google: Thanks, leaving it as is.

opus_codec_registration.src.html
tguilbert-google: > [...] support realime configuration [...]
> [...] the ability to re-configure encoders [...]

Is `flush()` followed by `configure()` acceptable here? Or is this about having the ability to only update the `packetlossperc` parameter mid-stream without flushing?

opus_codec_registration.src.html
tguilbert-google: I will leave no-caps for now, but will update if #588 resolves one way or the other.

opus_codec_registration.src.html
tguilbert-google: > [...] the default value are different in pc/mobile platform [...]

I'm not sure how to specify this in spec-text. I've added tentative text, but I don't think that "mobile platforms" is defined normatively.

I don't know how to specify 2 defaults in the dictionary either.

tguilbert-google: > Possible values are 1 (yes) and 0 (no)

The rest of the WebCodec spec uses `true` and `false` directly, so I am favoring these over `1` and `0`.

aboba: That makes sense.

aboba (approved):
Thanks for the updates. LGTM.

opus_codec_registration.src.html
bdrtc: > > [...] the default value are different in pc/mobile platform [...]
> 
> I'm not sure how to specify this in spec-text. I've added tentative text, but I don't think that "mobile platforms" is defined normatively.
> 
> I don't know how to specify 2 defaults in the dictionary either.



> That makes sense.



> > [...] support realime configuration [...]
> > [...] the ability to re-configure encoders [...]
> 
> Is `flush()` followed by `configure()` acceptable here? Or is this about having the ability to only update the `packetlossperc` parameter mid-stream without flushing?

it should be  only update the `packetlossperc` parameter, not flush.

bdrtc: > > [...] the default value are different in pc/mobile platform [...]
> 
> I'm not sure how to specify this in spec-text. I've added tentative text, but I don't think that "mobile platforms" is defined normatively.
> 
> I don't know how to specify 2 defaults in the dictionary either.

add this part as recommendation is more suitable, i have seen your latest update, that make sence 👍 .

opus_codec_registration.src.html
```diff
@@ -150,6 +158,31 @@
   <dt><dfn dict-member for=OpusEncoderConfig>frameDuration</dfn></dt>
   <dd>
     Configures the frame duration, in milliseconds, of output {{EncodedAudioChunk}}s.
+    If not specified, the default is `20` milliseconds.
```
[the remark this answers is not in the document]
tguilbert-google: Thanks, I've removed the text description of the defaults.

tguilbert-google:
I will leave some time for @padenot to comment before merging.

I am working on the Chromium implementation.

opus_codec_registration.src.html
dalecurtis: Are you still planning to linkify FEC and DTX per this conversation?

tguilbert-google: I lost track of the original comment. I've added links to RFC6716 (matching the frameDuration link added in #602)

opus_codec_registration.src.html
```diff
@@ -151,6 +159,31 @@
   <dd>
     Configures the frame duration, in milliseconds, of output {{EncodedAudioChunk}}s.
   </dd>
+  <dt><dfn dict-member for=OpusEncoderConfig>complexity</dfn></dt>
+  <dd>
+    Configures the encoder's computational complexity, as described 2.1.9. of
```
dalecurtis, line 164: in 2.1.9

opus_codec_registration.src.html
```diff
@@ -151,6 +159,31 @@
...
+    complexity. If no value is specificied, the default value is platform-specific:
+    User Agents <em class="rfc2119">SHOULD</em> set a default of `5` for mobile
+    platforms, and a default of `9` for all other platforms.
+  </dd>
+  <dt><dfn dict-member for=OpusEncoderConfig>packetlossperc</dfn></dt>
+  <dd>
+    Configures the encoder's expected packet loss percentage. The valid range is
+    `0` to `100`.
+
+    NOTE: The packet loss percentage might be updated over the course of the
```
dalecurtis, line 175: an encoding

opus_codec_registration.src.html
```diff
@@ -151,6 +159,31 @@
...
+    encoding, and it is recommended for User Agents to support these reconfigurations.
+  </dd>
+  <dt><dfn dict-member for=OpusEncoderConfig>useinbandfec</dfn></dt>
+  <dd>
+    Specifies whether the encoder provides Opus in-band FEC, as described by
```
dalecurtis, line 180: Worth spelling out Forward Error Correction (FEC) and whatever DTX stands for?

dalecurtis (approved):
lgtm % nits

opus_codec_registration.src.html
tguilbert-google: Done.

opus_codec_registration.src.html
tguilbert-google: Done.

opus_codec_registration.src.html
tguilbert-google: Done.

tguilbert-google:
This has been open for a few extra days, I will go ahead and merge. If there's anything that needs updating, please let me know @padenot.

Thanks for the reviews!

PR #597, opened 2022-10-21
Add requirements to VideoFrame metadata registry definition

chrisn:
Following the discussion in #559, this PR covers further changes to adding, changing, and deprecating registry entries.

### DIFF
--- a/video_frame_metadata_registry.src.html
+++ b/video_frame_metadata_registry.src.html
@@ -37,9 +37,9 @@ and has the following requirements:
 2. Each metadata entry must be serializable.
 3. Each metadata entry must be defined by a W3C specification that has
     reached consensus in the originating Working Group.
-4. Each metadata entry must have clearly defined semantics. In particular,
-    its interactions with the media processing pipeline (encoders, decoders,
-    renderers, etc.) should be well defined.
+4. The specification defining each metadata entry must provide clearly defined
+    semantics. In particular, its interactions with the media processing
+    pipeline (encoders, decoders, renderers, etc.) should be well defined.
 5. A candidate registration entry must be announced by filing an issue in the
     WebCodecs GitHub issue tracker so they can be discussed and evaluated for
     compliance before being added to the registry. If the Media Working Group
--- a/video_frame_metadata_registry.src.html
+++ b/video_frame_metadata_registry.src.html
@@ -38,8 +38,8 @@ and has the following requirements:
 3. Each metadata entry must be defined by a W3C specification that has
     reached consensus in the originating Working Group.
 4. The specification defining each metadata entry must provide clearly defined
-    semantics. In particular, its interactions with the media processing
-    pipeline (encoders, decoders, renderers, etc.) should be well defined.
+    semantics. In particular, interactions with the media processing pipeline
+    (encoders, decoders, renderers, etc.) should be well defined.
 5. A candidate registration entry must be announced by filing an issue in the
     WebCodecs GitHub issue tracker so they can be discussed and evaluated for
     compliance before being added to the registry. If the Media Working Group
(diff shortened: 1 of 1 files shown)


aboba (commented):
Lines 40-42:  You might say, "The specification defining the metadata entry must provide clearly defined semantics."  That is, the semantics should be defined in the specification, not the registration request.

chrisn:
Thanks, I have pushed an update to clarify this.

dalecurtis (approved):

aboba (approved):

video_frame_metadata_registry.src.html
```diff
@@ -30,21 +30,30 @@
...
-A registration entry is a document describing one or several metadata entry,
+A registration entry is a document describing one or several metadata entries,
 and has the following requirements:
 1. Each metadata entry is defined as a separate {{VideoFrameMetadata}}
     dictionary member.
 2. Each metadata entry must be serializable.
-3. Each metadata must be defined by a W3C specification and have clearly defined semantics.
-    In particular, its interactions with the media processing pipeline
+3. Each metadata entry must be defined by a W3C specification that has
+    reached consensus in the originating Working Group.
```
tidoust, line 39: Three questions:
1. Does this need to be restricted to **W3C** specifications?
2. Should this be restricted to normative specifications?
3. What does it mean for a W3C specification to have "reached consensus in the originitaing Working Group"? 

For 1., I realize that it's hard to come up with an explicit set of requirements for a specification to be considered good enough. A document that comes to mind are the [considerations the Director takes into account when evaluating normative references from W3C documents](https://www.w3.org/2013/09/normative-references). These considerations include considerations such as:
- Is the specification produced by a group that follows the [OpenStand principles](http://open-stand.org/principles/)?
- Is the specification available in English and at no cost?
- Are the technologies in the referenced parts available under terms that are compatible with the [W3C Royalty-Free licensing requirements](https://www.w3.org/Consortium/Patent-Policy/#sec-Requirements)?

For 2., I'm wondering whether the group is ok with specifications on the Note track, as done for codec registrations but perhaps not such a good example to follow?

For 3., a status on the Recommendation track (e.g. Candidate Recommendation), would make things clearer. Also, I'm thinking that the group might want to add entries even when a spec is still under development?

I'm having a hard time coming up with a proposal that is not convoluted, so perhaps it's fine to leave this a bit under specified. If the group wanted to open the registry to non-W3C specs, the following wording could be considered:

```suggestion
3. Each metadata entry must be defined in an established standard with licensing terms that are consistent with royalty-free implementation of W3C Recommendations. The Media Working Group will evaluate candidate specifications following [considerations used when evaluating normative references from W3C specifications](https://www.w3.org/2013/09/normative-references) [[NORMATIVE-REFERENCES]].
```

youennf (approved):
LGTM.
With regards to the process, I think keeping it a bit unspecified until we exercise it seems ok.
In general, I would think that metadata entries being normative, they should follow the REC track.
The restriction to W3C specifications (but they might refer to non W3c specs themselves) is fine for now as well, given spec authors will anyway need to interact deeply with the Media WG.

tidoust (approved):

video_frame_metadata_registry.src.html
chrisn: Thanks @tidoust, my thoughts on your questions: 

1. We discussed limiting to W3C specifications in https://github.com/w3c/webcodecs/pull/559#issuecomment-1267387427, my interpretation is partly as a way to ensure compatible licensing
2. My understanding is that metadata entries should be normative, but they may be optional to implement, per https://github.com/w3c/webcodecs/pull/559#issuecomment-1267420661
3. I expect that we will want to add entries in draft state. I saw consensus here as: Media WG having consensus to add the registration, and the originating WG having a draft spec to add and consensus on its purpose and scope, it's OK if details are still in development (https://github.com/w3c/webcodecs/pull/559#pullrequestreview-1127856382)

dalecurtis:
Are we ready to merge this one?

chrisn:
I think so, @padenot is this OK from your perspective?

chrisn:
@padenot I'd like to merge this, as we now have incoming proposals to add to the registry. Thanks!

padenot:
Sure, sorry for missing this, go ahead!

Issue #332, opened 2021-08-10
The file produced by the 'capture file' sample is broken in some way


sashazxtt:
The Webm file produced by this sample: https://w3c.github.io/webcodecs/samples/capture-to-file/capture-to-file.html
is broken in some way. It doesn't properly seek. In VLC the current timestamp becomes some huge value. The Windows video player throws an error. Chrome simply stops playing.

The duration (always) and current time are displaying properly up to some point (30 or se seconds).

nixxquality:
I noted the same. It's an issue with the webm writer that I copied over to my own project.

After around 31-32 seconds my media player mpv interprets the timestamps as becoming negative.
Seeking into past 32 seconds doesn't work.

nixxquality:
Going through the modified webm-writer code and comparing it to the [original](https://github.com/thenickdude/webm-writer-js/)

The timestamp issue past 32 seconds is because `MAX_CLUSTER_DURATION_MSEC` was changed from `5000` to `5000000`.

```
> ffmpeg -i C:\Users\Linus\Videos\test.webm -c copy .\Videos\testt.webm
[webm @ 000001887bc37480] Application provided invalid, non monotonically increasing dts to muxer in stream 0: 65548 >= 28
```
Looks like a classic overflow, but why?

```js
// all irrelevant parts cut out
    function addFrameToCluster(frame) {
        if (clusterDuration >= MAX_CLUSTER_DURATION_MSEC) {
          flushClusterFrameBuffer();
        }
      }

    function flushClusterFrameBuffer() {
        cluster.data.push(createSimpleBlockForframe(clusterFrameBuffer[i]));
    }

    function createSimpleBlockForframe(frame) {
      bufferStream.writeU16BE(frame.timecode); // 🚨
    }
```

The file being unseekable is a separate issue, though. I'm not sure why that happens.

nixxquality:
Don't get me wrong, my solution is definitely not proper and I wouldn't recommend it.
It's merely replacing one non-standard hack that only works in some cases with another non-standard hack that only works in some media players.
The real solution would be fixing the cue writing.

dalecurtis:
Probably we should just replace the webm writer with mp4box.js since it's better supported if there are issues with the upstream version.

PR #618, opened 2022-12-07
Add nixxquality's fixes

josephrocca:
See: https://github.com/w3c/webcodecs/issues/332#issuecomment-1077442192


josephrocca:
Copying a recent comment from nixxquality here for context:

> Don't get me wrong, my solution is definitely not proper and I wouldn't recommend it.
> It's merely replacing one non-standard hack that only works in some cases with another non-standard hack that only works in some media players.
> The real solution would be fixing the cue writing.

dalecurtis:
Don't have plans to replace with mp4box.js anytime soon so just going to land this for now.

PR #620, opened 2022-12-09
 Add a way of specifying blockSize and compressLevel to FLAC encod…

### LINKED ISSUE #595: Add a way of specifying frame duration to FLAC encoder config
This is a followup to #405. #551 added a frameDuration attribute to the Opus encoder configuration, and closed #405 since Opus is the only codec with this tuning knob supported _by any browser_.

I'm the author of the polyfill: https://github.com/ennuicastr/libavjs-webcodecs-polyfill . The polyfill does support FLAC encoding.

I'll be adding support for this tuning knob to the Opus encoder in libavjs-webcodecs-polyfill Soon™, but I'd like to also add it to FLAC, and ideally do so following the spec instead of deviating from the spec with the possibility of Chrome later deviating in a different way. That is, I'd like to see similar for FLAC.

bdrtc:
This PR adds blockSize, compressLevel configuration knobs to the FLAC encoder.

This should fix #595 , 
@Yahweasel @dalecurtis

### DIFF
--- a/flac_codec_registration.src.html
+++ b/flac_codec_registration.src.html
@@ -137,7 +137,7 @@ To check if an {{FlacEncoderConfig}} is valid, run these steps:
 1. If {{FlacEncoderConfig/blockSize}} is not a valid block size,
 	which is described section 5.1 of [[FLAC-DRAFT]], return `false`.
 1. If {{FlacEncoderConfig/compressLevel}} is specified and not within the range of
-    `0`(fastest, least compression) and `8`(slowest, most compression) inclusively, return `false`.
+    `0` (fastest, least compression) and `8` (slowest, most compression) inclusively, return `false`.
 2. Return `true`.
 
 <dl>
--- a/flac_codec_registration.src.html
+++ b/flac_codec_registration.src.html
@@ -16,7 +16,8 @@ Abstract: This registration is entered into the [[webcodecs-codec-registry]].
     (2) the codec-specific {{EncodedAudioChunk}}
     [=EncodedAudioChunk/[[internal data]]=] bytes, (3) the
     {{AudioDecoderConfig.description}} bytes, and (4) the values of
-    {{EncodedAudioChunk}} [=EncodedAudioChunk/[[type]]=].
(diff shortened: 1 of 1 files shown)


dalecurtis (approved):

bdrtc:
@padenot @aboba .

flac_codec_registration.src.html
```diff
@@ -94,6 +102,59 @@
...
+dictionary FlacEncoderConfig {
+  [EnforceRange] unsigned long blockSize = 0;
+  [EnforceRange] unsigned long compressLevel = 5;
+};
+</xmp>
+</pre>
+
+To check if an {{FlacEncoderConfig}} is valid, run these steps:
+1. If {{FlacEncoderConfig/blockSize}} is not a valid block size,
+	which is described section 5.1 of [[FLAC-DRAFT]], return `false`.
```
chrisn, line 138: Do we need both the `FLAC` and `FLAC-DRAFT` references? The FLAC reference also describes the valid block size. That said, if this is going through the IETF, we should prefer the IETF-DRAFT, then update the registration once it becomes an RFC

chrisn (changes requested):

flac_codec_registration.src.html
bdrtc: IETF-DRAFT should be more formal, we will record this, and update when it become final RFC.

Yahweasel:
> Do any reviewers have further comments here ? Thanks for your time 1st_place_medal

I have none. Works for me.

dalecurtis:
@chrisn did you want to stamp and merge?

chrisn:
There's just one change remaining, I suggest removing the `FLAC-DRAFT` reference and updating `FLAC` to point to the IETF draft.

bdrtc:
> There's just one change remaining, I suggest removing the `FLAC-DRAFT` reference and updating `FLAC` to point to the IETF draft.

done, update FLAC reference  to  IETF draft.   @chrisn

chrisn (approved):
Thank you @bdrtc, looks good!

PR #630, opened 2023-01-24
Fix memory layout of NV12 image example

tidoust:
According to the text, which I hope I read correctly:
- The coded width and coded height must be even. So image cannot be 16x9. This update makes it 16x10.
- There should be one Y byte per pixel, so 16x10 Y bytes (example only had 14 bytes per row).
- The U and V components are sub-sampled horizontally and vertically by a factor of 2, and interleaved, so there should 16x10/4 = 40 UV blocks (example only sub-sampled horizontally).

This update also fixes occurrences of "horizontaly".

***
<a href="https://pr-preview.s3.amazonaws.com/tidoust/webcodecs/pull/630.html" title="Last updated on Jan 24, 2023, 2:37 PM UTC (6728d4e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/630/47b8051...tidoust:6728d4e.html" title="Last updated on Jan 24, 2023, 2:37 PM UTC (6728d4e)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -4268,7 +4268,7 @@ is its own [=equivalent opaque format=].
     planes of Chroma, denoted Y, U and V, and present in this order. It is also
     often refered to as Planar YUV 4:2:0.
 
-    The U an V planes are [=sub-sampled=] horizontaly and vertically by a
+    The U an V planes are [=sub-sampled=] horizontally and vertically by a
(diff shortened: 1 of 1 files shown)


dalecurtis (approved):

dalecurtis:
Nice catch!

Issue #270, opened 2021-06-07
Support per-frame QP configuration by VideoEncoder 


daijh:
Customize BRCs by applications, in addition to CBR/VBR provided by VideoEncoder.

Some applications prefer to implement its own BRC according to per-frame QP, to better response to network bandwidth estimation. One such per-frame QP BRC implement can be found:
https://github.com/intel/libmebo/blob/master/explainer.md

```
dictionary VideoEncoderEncodeOptions {
  boolean keyFrame = false;

  // proposal
  unsigned long QP;
};
```

daijh:
#56

chcunningham:
Thanks for filing @daijh. Can you say more on why per-frame is important vs making this part of the config (affecting N frames). I'm surprised we would want to adjust it with every frame (e.g. 30 times a second).

daijh:
@chcunningham
The BRC is something intrinsic of the video encoder, and it usually cannot be much changed after platform shipped.
Some developers would like to use their own BRC cross all platforms, instead of platform intrinsic BRC.
Per-frame QP enables the developer to implement customized BRC, by set QP of every frame according to previous encoded frames size and bitrate budget, as mentioned in #56

For example:
- Intrinsic BRC is not implement (e.g. VBR is not supported by some HW encoders) or the implementation is not what developers want.

Djuffin:
As I see it we’ll need to make at least 3 following changes in the API. I'm  curious to know what the community thinks 

### New bitrate mode. 
- We’ll introduce a new value to [BitrateMode](https://www.w3.org/TR/mediastream-recording/#bitratemode) enum : ‘qp’ 
### QP range discovery. Either 
1. A new optional field to [VideoEncoderConfig](https://www.w3.org/TR/webcodecs/#video-encoder-config) dictionary : [ULongRange](https://www.w3.org/TR/mediacapture-streams/#dictionary-ulongrange-members) `qpRange` 
[...]

PR #633, opened 2023-02-01
Introducing a codec specific 'quantizer' mode for VideoEncoder

Djuffin:
A codec specific quantizer is set for each video frame for constant quality or fine tuned external rate control.

Addressing Issue [270](https://github.com/w3c/webcodecs/issues/270)

***
<a href="https://pr-preview.s3.amazonaws.com/Djuffin/webcodecs/pull/633.html" title="Last updated on Feb 28, 2023, 7:38 AM UTC (40954c3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/633/704c167...Djuffin:40954c3.html" title="Last updated on Feb 28, 2023, 7:38 AM UTC (40954c3)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -2073,7 +2073,7 @@ dictionary VideoEncoderConfig {
   HardwareAcceleration hardwareAcceleration = "no-preference";
   AlphaOption alpha = "discard";
   DOMString scalabilityMode;
-  BitrateMode bitrateMode = "variable";
+  VideoEncoderBitrateMode bitrateMode = "variable";
   LatencyMode latencyMode = "quality";
 };
 </xmp>
@@ -2180,8 +2180,8 @@ To check if a {{VideoEncoderConfig}} is a <dfn>valid VideoEncoderConfig</dfn>,
 
   <dt><dfn dict-member for=VideoEncoderConfig>bitrateMode</dfn></dt>
   <dd>
-    Configures encoding to use a {{BitrateMode/constant}} or
-    {{BitrateMode/variable}} bitrate as defined by [[MEDIASTREAM-RECORDING]].
+    Configures encoding to use one of the rate control modes specified by
+    {{VideoEncoderBitrateMode}}.
 
     NOTE: The precise degree of bitrate fluctuation in either mode is
         implementation defined.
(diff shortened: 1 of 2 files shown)


aboba:
So quantizer support is only for VP9?

Djuffin:
> So quantizer support is only for VP9?
No, this PR is deliberately limited to VP9 to demonstrate the idea and keep the change small.
AV1 and H264 will follow.

aboba (approved):

solmaks:
If there is no technical reason to exclude other codecs, can we come up with codec agnostic API surface from the get go?

aboba:
Could we have "VideoEncoderEncodeOptions" instead of "VideoEncoderEncodeOptionsForVp9"?

Also, might be helpful to expose the QPmin and QPmax to  use [EnforceRange] so quantizer values can't be set outside the range..

Djuffin:
> If there is no technical reason to exclude other codecs, can we come up with codec agnostic API surface from the get go?

> Could we have "VideoEncoderEncodeOptions" instead of "VideoEncoderEncodeOptionsForVp9"?

We can certainly simply put quantizer into VideoEncoderEncodeOptions, like it was originally proposed in https://github.com/w3c/webcodecs/issues/270, but there is a [concern](https://github.com/w3c/webcodecs/issues/270#issuecomment-1203219196) that QP is too codec specific in its range and effect on encoding.  It's like having very different parameters assigned to the same name.

> Also, might be helpful to expose the QPmin and QPmax and to use [EnforceRange] so quantizer values can't be set outside the range.

I though [EnforceRange] only enforces integer range.
QPmin and QPmax are part of codec spec, power users that use per-frame QP probably know it already, don't they?

Djuffin:
@padenot @youennf 
What do you think about quantizer being codec specific rather than a generic parameter?

stepancar (approved):

solmaks:
Introducing an API surface that will need changes/updates for every new video codec is an unnecessary complexity (and extra work for everyone). For instance, getting this available for H264 will required a spec iteration (and more code in implementation), same for AV1, same for H265, same for some new codec 10 year later down the road. Is there a concern that QP does not apply to or does not have a similar effect (higher QP means higher compression, lower QP means lower compression) on current generation of video codecs used in the industry?

Djuffin:
I don't feel very strongly about it, but since QP has different ranges for different codecs it seems logical to put them into codec specific parts of the spec.

aboba:
It is probably beyond the scope of this PR, but I'd note that the potential addition of segmentation metadata (for face detection and/or background blur) to the VideoFrame has raised the question of per-macroblock QP, which would allow the encoder to focus its efforts on "important" macro blocks (e.g. the macroblocks corresponding to the foreground, rather than a (blurred) background).

Djuffin:
This PR was approved at the last meeting of the media working group.

PR #651, opened 2023-03-10
Fix usage of {Audio/Video}Decoder.isConfigSupported() in example

chrisguttandin:
I've noticed that the usage of `isConfigSupported()` in the example is probably not as intended. So far the assertion only checks if `isConfigSupported()` returns a truthy value which is always the case.

I've changed the code to match the usage here: 
https://github.com/w3c/webcodecs/blob/34e21cb731360339b249ceeb6234129a011b92c9/samples/capture-to-file/encode-worker.js#L38-L39

Please let me know if there is anything I should change.

### DIFF
--- a/samples/lib/audio_renderer.js
+++ b/samples/lib/audio_renderer.js
@@ -31,7 +31,8 @@ export class AudioRenderer {
 
     debugLog(config);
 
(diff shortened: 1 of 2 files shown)


chrisn (approved):
Looks good, thank you!

PR #659, opened 2023-04-01
Per frame quantizer option for Av1.

Djuffin:
This value is used by VideoEncoder.encode() if VideoEncoder is configured with "quantizer" bitrate mode.

Follow up for: https://github.com/w3c/webcodecs/issues/270


aboba:
LGTM.

aboba (approved):

Issue #646, opened 2023-03-01
Support for Screen Content Coding


aboba:
Currently there is no parameter defined in the AV1 registry to enable or force screen content coding (e.g. set seq_force_screen_content_tools to SELECT_SCREEN_CONTENT_TOOLS). 

Reference: [AV1 Bitstream Spec](https://aomediacodec.github.io/av1-spec/av1-spec.pdf) page 31

Djuffin:
WebCodecs Editors decision: let's put the knob into VideoEncoderConfig so VideoEncoder can explicitly refuse it if it's not supported.

PR #652, opened 2023-03-16
Enable configuration of AV1 screen content coding tools

aboba:
Fixes https://github.com/w3c/webcodecs/issues/646

PR #662, opened 2023-04-11
Enable configuration of AV1 screen content coding tools

aboba:
Fixes https://github.com/w3c/webcodecs/issues/646

Rebase of https://github.com/w3c/webcodecs/pull/652

### DIFF
--- a/av1_codec_registration.src.html
+++ b/av1_codec_registration.src.html
@@ -74,6 +74,47 @@ If an {{EncodedVideoChunk}}'s {{EncodedVideoChunk/[[type]]}} is
 contain a frame with a `frame_type` of `KEY_FRAME` as defined in Section
 6.8.2 of [[AV1]].
 
+VideoEncoderConfig extensions {#videoencoderconfig-extensions}
+==============================================================
+
+<pre class='idl'>
+<xmp>
+partial dictionary VideoEncoderConfig {
+  AV1EncoderConfig av1;
+};
+</xmp>
+</pre>
+
+<dl>
+  <dt><dfn dict-member for=VideoEncoderConfig>av1</dfn></dt>
+  <dd>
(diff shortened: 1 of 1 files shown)


Djuffin:
Is it really a per frame setting? 
It seems that https://github.com/w3c/webcodecs/pull/652 does the same but on the VideoEncoderConfig level

aboba:
PR redone to make it an encoder configuration option, as discussed in the Editor's meeting.

av1_codec_registration.src.html
```diff
@@ -74,6 +74,47 @@
 contain a frame with a `frame_type` of `KEY_FRAME` as defined in Section
 6.8.2 of [[AV1]].
 
+VideoEncoderConfig extensions {#videoencoderconfig-extensions}
```
Djuffin, line 77: If I understand the structure of this document. VideoEncoderConfig needs to be added to the 'Abstract' section as well.

aboba: Fixed.

Djuffin (approved):

aboba:
Discussed at MEDIA WG meeting: https://lists.w3.org/Archives/Public/public-media-wg/2023Apr/0007.html

aboba:
Editor's meeting: merge it!

fippo:
Late comment but since this hasn't shipped yet... can we stop saying this is "forced" and instead say it gets enabled (i.e. `enableScreenContentTools`)?
Forcing may have implications e.g. when you try to do "force" it on a hardware encoder where it is not supported. Does that imply it should fail or should this simply not get enabled.

Wordsmithing ;-)

fippo:
(force... is used by the AV1 spec so keeping this consistent is good)

PR #667, opened 2023-04-19
Add per-frame QP support to AVC

aboba:
This value is used by VideoEncoder.encode() if VideoEncoder is configured with "quantizer" bitrate mode. 

Followup for: #270


avc_codec_registration.src.html
```diff
@@ -187,6 +188,42 @@
...
+</pre>
+
+<dl>
+  <dt><dfn dict-member for=VideoEncoderEncodeOptions>avc</dfn></dt>
+  <dd>
+    Contains codec specific encode options for the [[ITU-T-REC-H.264]] codec.
+  </dd>
+</dl>
+
+VideoEncoderEncodeOptionsForAv1 {#av1-encode-options}
```
Djuffin, line 209: Should be Avc instead of Av1

aboba: Fixed.

aboba:
Editor's meeting: merge it!

PR #671, opened 2023-05-10
Clone configuration should perform a deep copy

### LINKED ISSUE #485: Clarify "Clone Configuration" to eliminate TOCTOU race possibility
Clone Configuration is used in the various isConfigSupported() steps ([example](https://w3c.github.io/webcodecs/#dom-audiodecoder-isconfigsupported)) to make a copy of the config as seen by the UA in assessing "support". The steps in "Clone Configuration" make a recursive shallow copy of the given config. For configs that contain a "description" BufferSource member, this should mean the output config points to the same BufferSource as the input config (copy by reference). This opens the possibility of TOCTOU races where the user may pass the config to isConfigSupported() and then modify the buffer while isConfigSupported() is asynchronously checking for support. It seems better to instead make a deep copy of the BufferSource in the synchronous portion of isConfigSupported().

Similarly, we should probably make deep copy (using the same clone algo) in the synch steps of configure().

padenot:
This fixes #485.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/pull/671.html" title="Last updated on May 10, 2023, 5:13 PM UTC (052d98c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/671/3ec4530...052d98c.html" title="Last updated on May 10, 2023, 5:13 PM UTC (052d98c)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -1766,7 +1766,7 @@ Run these steps:
         `config[m]`.
     3. Otherwise, assign a copy of `config[m]` to `clone[m]`.
 
-Note: This implements a "deep-copy". Those configuration objects are
+Note: This implements a "deep-copy". These configuration objects are
 frequently used as the input of asynchronous operations. Copying means that
 modifying the original object while the operation is in flight won't change the
 operation's outcome.
--- a/index.src.html
+++ b/index.src.html
@@ -1764,7 +1764,12 @@ Run these steps:
     2. If `config[m]` is a nested dictionary, set `clone[m]` to the result of
         recursively running the <a>Clone Configuration</a> algorithm with
         `config[m]`.
-    3. Otherwise, assign the value of `config[m]` to `clone[m]`.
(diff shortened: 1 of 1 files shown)


index.src.html
```diff
@@ -1764,7 +1764,12 @@
     2. If `config[m]` is a nested dictionary, set `clone[m]` to the result of
         recursively running the <a>Clone Configuration</a> algorithm with
         `config[m]`.
-    3. Otherwise, assign the value of `config[m]` to `clone[m]`.
+    3. Otherwise, assign a copy of `config[m]` to `clone[m]`.
+
+Note: This implements a "deep-copy". Those configuration objects are
```
dalecurtis, line 1769: "These"

dalecurtis (approved):
In Chrome's case we don't intentionally make a deep copy, media code just can't handle Blink GC'd types, so we copy out of the GC type into our own.

aboba (approved):

Issue #104, opened 2020-11-09
Detach ImageDecoder inputs


chcunningham:
ImageDecoder isn't in the draft spec yet, but see here for latest interface in Chromium.

This is issue is a reminder that we should consider specifying that inputs are detached after the call to decode returns  (neutering the buffer from JS pov). [Similar to WebAudio's decodeAudio()](https://www.w3.org/TR/webaudio/#dom-baseaudiocontext-decodeaudiodata). Detaching the inputs transfers ownership for the UA to safely perform asynch decoding without requiring a copy.

sandersdan:
Previously we decided against this for two reasons:

1. It's not compatible with SharedArrayBuffer. In retrospect, it's probably fine to fall back to copying in this case, but that would mean inconsistent behavior.

2. It doesn't allow the input buffers to be reused, and therefore creates extra GC pressure. In practice it doesn't seem like this garbage is consequential.

The primary alternative to transferring inputs is to always copy inputs immediately.

chcunningham:
Title update: problem isn't unique to ImageDecoder. Lets fix this everywhere.

Agree with points 1 and 2 above.

chcunningham:
Dropping some questions for TAG

* is it better to always transfer when possible (e.g. yes for ArrayBuffer and ArrayBufferView, no for SAB) or to make it user configurable (e.g. transfer list)? For reference, inputs to WebAudio's [decodeAudioData()](https://webaudio.github.io/web-audio-api/#dom-baseaudiocontext-decodeaudiodata) are automatically transferred. 
* if configurable, is a transfer list the recommended approach? We've also considered dual static create methods like EncodedVideoChunk::TransferFrom() and EncodedVideoChunk::CopyFrom().

padenot:
>     2\. It doesn't allow the input buffers to be reused, and therefore creates extra GC pressure. In practice it doesn't seem like this garbage is consequential.

There are evidences in the Web Audio issue tracker that folks in the video game industry find this problematic and like "zero-garbage" paths.

I don't have a solution however, but a chat with representatives from this industry is planned.

padenot:
An update after talking with a couple of people from different industries: it's getting increasingly clear that detaching is needed, but that also internal copy is useful. We'll probably have to support both.

padenot:
[...]

Issue #212, opened 2021-05-03
Improve memory locality for higher performances and reduced memory working set


padenot:
We can add a second parameter to `decode`/`encode` and the output callbacks (for audio, video and images), that would be respectively an output buffer or handle, and the input buffer or handle (and also mirror it for encode vs. decode), to allow authors to reuse memory and avoid **lots** of memory traffic. GCs are amazing, but it's always better to not GC, reuse allocations, and always use the same bit of the address space, especially in very demanding multimedia scenarios.

This is **really** compelling whenever the input (resp. output for decode) is regular memory, since it means the working set is going to be pretty tight (for example help with https://github.com/w3c/webtransport/issues/231), but is really nice even when some buffers are backed by (say) GPU memory (because you're not doing malloc/free pairs for input/output buffers for respectively decoding and encoding). I'm not sure if GPU-memory backed surfaces can work in the same way.

Essentially, the proposal is to "pass through" the input to the output so that it can be reused. This reduces allocator use, but more importantly greatly improves memory locality.

When **not** using this new parameter, the implementation allocates as usual. This means that authors don't have to try to guess the buffer size, it's allocated once and then can be resized. The buffer is released naturally because it isn't referenced, immediately. In short, this is not a breaking change, but allows developers that want to control their memory allocation to be able to do so (and I believe this is something that is important for quite a few domains).

When the provided buffer is too small, the browser can allocate a new buffer, or use more complex tricks (realloc, or maybe it finds that the array buffer backing memory storage was in fact big enough, etc.).

When the provided buffer is way too big, the implementation can decide to swap it for a buffer that has a more appropriate size.

aboba:
@padenot This sounds interesting.  Having some WebIDL might help in walking through the implications for various use cases (and related APIs, particularly those based on WHATWG Streams).  

A use case that comes to mind is media capture of a track, followed by WebCodecs encode and then sending (e.g. via WebTransport, RTCDataChannel, etc.). 

There is work underway to support [zero-copy capture](https://bugs.chromium.org/p/chromium/issues/detail?id=1120900). My understanding is that this will (in concert with [MediaCapture Transform](https://w3c.github.io/mediacapture-transform/)) provide a `VideoFrame` in a GPU buffer, and can then be provided to WebCodecs for Encoding, yielding an `encodedVideoChunk` that can then be sent via a transport. 

[...]

PR #676, opened 2023-05-18
Add VideoFrameBufferInit.transfer

Djuffin:
This should allow to avoid extra copies while constructing VideoFrame from pixel data

Addressing: https://github.com/w3c/webcodecs/issues/104

***
<a href="https://pr-preview.s3.amazonaws.com/Djuffin/webcodecs/pull/676.html" title="Last updated on Jun 2, 2023, 3:13 AM UTC (4897397)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/676/3a96dbd...Djuffin:4897397.html" title="Last updated on Jun 2, 2023, 3:13 AM UTC (4897397)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -3277,6 +3277,8 @@ dictionary VideoFrameBufferInit {
   [EnforceRange] unsigned long displayHeight;
 
   VideoColorSpaceInit colorSpace;
+
+  sequence<ArrayBuffer> transfer = [];
 };
 
 dictionary VideoFrameMetadata {
@@ -3421,23 +3423,33 @@ dictionary VideoFrameMetadata {
 3. Let |overrideRect| be `undefined`.
 4. If |init|.{{VideoFrameBufferInit/visibleRect}} [=map/exists=], assign its
     value to |overrideRect|.
-4. Let |parsedRect| be the result of running the [=VideoFrame/Parse Visible
+5. Let |parsedRect| be the result of running the [=VideoFrame/Parse Visible
         Rect=] algorithm with |defaultRect|, |overrideRect|,
         |init|.{{VideoFrameBufferInit/codedWidth}},
         |init|.{{VideoFrameBufferInit/codedHeight}}, and
         |init|.{{VideoFrameBufferInit/format}}.
-5. If |parsedRect| is an exception, return |parsedRect|.
-6. Let |optLayout| be `undefined`.
-7. If |options|.{{VideoFrameBufferInit/layout}} [=map/exists=], assign its value
+6. If |parsedRect| is an exception, return |parsedRect|.
+7. Let |optLayout| be `undefined`.
+8. If |options|.{{VideoFrameBufferInit/layout}} [=map/exists=], assign its value
     to |optLayout|.
-8. Let |combinedLayout| be the result of running the [=VideoFrame/Compute
+9. Let |combinedLayout| be the result of running the [=VideoFrame/Compute
     Layout and Allocation Size=] algorithm with |parsedRect|,
     |init|.{{VideoFrameBufferInit/format}}, and |optLayout|.
-9. If |combinedLayout| is an exception, throw |combinedLayout|.
-10. If `data.byteLength` is less than |combinedLayout|’s
+10. If |combinedLayout| is an exception, throw |combinedLayout|.
+11. If `data.byteLength` is less than |combinedLayout|’s
     [=combined buffer layout/allocationSize=], throw a {{TypeError}}.
-11. Let |resource| be a new [=media resource=] containing a copy of |data|. Use
-        {{VideoFrameBufferInit/visibleRect}} and {{VideoFrameBufferInit/layout}}
+12. If |init|.{{VideoFrameBufferInit/transfer}} contains more than one reference
+     to the same {{ArrayBuffer}}, then throw a {{DataCloneError}} {{DOMException}}.
+13. For each |transferable| in |init|.{{VideoFrameBufferInit/transfer}}:
+     1. If {{platform object/[[Detached]]}} internal slot is `true`,
+         then throw a {{DataCloneError}} {{DOMException}}.
+14. If |init|.{{VideoFrameBufferInit/transfer}} contains an {{ArrayBuffer}}
+    referenced by |data| the User Agent <em class="rfc2119">MAY</em> choose to:
+    1. Let |resource| be a new [=media resource=] referencing pixel data in
+        |data|.
+15. Otherwise:
+    1. Let |resource| be a new [=media resource=] containing a copy of |data|.
+        Use {{VideoFrameBufferInit/visibleRect}} and {{VideoFrameBufferInit/layout}}
         to determine where in |data| the pixels for each plane reside.
 
         The User Agent <em class="rfc2119">MAY</em> choose to allocate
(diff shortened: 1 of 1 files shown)


index.src.html
```diff
@@ -3415,29 +3417,39 @@
   VideoFrame(data, init)
 </dfn>
 1. If |init| is not a [=valid VideoFrameBufferInit=], throw a {{TypeError}}.
-2. Let |defaultRect| be «[ "x:" → `0`, "y" → `0`,
+1. Let |defaultRect| be «[ "x:" → `0`, "y" → `0`,
```
dalecurtis, line 3420: Numbers look all wrong?

Djuffin: It's actually an easier way format lists in markdown

https://carpentries-incubator.github.io/markdown-intro/06-lists/index.html

dalecurtis: Huh neat, TIL.

dalecurtis (approved):

index.src.html
```diff
@@ -3415,29 +3417,39 @@
...
-10. If `data.byteLength` is less than |combinedLayout|’s
+1. If |combinedLayout| is an exception, throw |combinedLayout|.
+1. If `data.byteLength` is less than |combinedLayout|’s
     [=combined buffer layout/allocationSize=], throw a {{TypeError}}.
-11. Let |resource| be a new [=media resource=] containing a copy of |data|. Use
-        {{VideoFrameBufferInit/visibleRect}} and {{VideoFrameBufferInit/layout}}
+1. For each |transferable| in |init|.{{VideoFrameBufferInit/transfer}}:
+    1. If {{platform object/[[Detached]]}} internal slot is `true`,
+        then throw a {{DataCloneError}} {{DOMException}}.
+1. If |init|.{{VideoFrameBufferInit/transfer}} contains an ArrayBuffer referenced
```
padenot, line 3444: Markup on `ArrayBuffer`.

index.src.html
```diff
@@ -3415,29 +3417,39 @@
...
+    by |data|:
+    1. Let |resource| be a new [=media resource=] referencing pixels in |data|
```
padenot, line 3446: `s/pixels/pixel data/` to match the rest of the text.

index.src.html
```diff
@@ -3415,29 +3417,39 @@
...
+        without a copy.
```
padenot, line 3447: `without a copy` sounds strange. Either you reference the data, or you copy, but referencing the copy is not what we want here (since everything is a reference, it would just mean to copy, but reference the old data?).

Generally, `referencing ...` is enough to mean there's no copy.

We've preferred assigning the reference to something else, and, when necessary, add a non-normative note to make it extra-clear that there are no copies. The normative text for this is https://w3c.github.io/webcodecs/#raw-media-memory-model-reference-counting if we need a link.

index.src.html
```diff
@@ -3415,29 +3417,39 @@
...
+        Use {{VideoFrameBufferInit/visibleRect}} and {{VideoFrameBufferInit/layout}}
+        to determine where in |data| the pixels for each plane reside.
```
padenot, line 3449: This is very hand-wavy, is there an algorithm to link to? Or maybe it just follows from the rest of the prose and can be omitted.

index.src.html
```diff
@@ -3415,29 +3417,39 @@
...
+1. Otherwise:
+    1. Let |resource| be a new [=media resource=] containing a copy of |data|.
+        Use {{VideoFrameBufferInit/visibleRect}} and {{VideoFrameBufferInit/layout}}
```
padenot, line 3452: This is however useful because it can lead to copying less data, but really the paragraph below is enough.

padenot (commented):
Looks good, the prose needs a bit of fixing.

We should also be doing the same for the audio side, not only because we've strived to make the API all symmetric (audio / video) and mirrored (encoding / decoding), but also because this will save a lot of small and very frequent allocations: while you might be doing 30-60 allocations per second for a video scenario (in addition to horribly large copies), you can be doing hundreds of allocations per second on the audio side: imagine an `AudioWorkletProcessing` feeding an `AudioEncoder`, buffers are 128-frames long, sample-rate is e.g. 48000Hz, we're looking at 375 buffer allocations (and small copies, that's less problematic) per second. We can then implement https://github.com/w3c/webcodecs/issues/212#issuecomment-832075352, and make it, so there is no allocation after some time, by naturally pooling the buffers. Authors that don't mind allocation can simply not specify the parameters and/or the transfer list.

Unrelated, please don't mix functional and non-functional (formatting / refactoring) changes in the same commit, it makes things unnecessarily hard to review and mistakes eventually slip through because of this.

In general, it's a lot better to use regular numbering in markdown, because it's not rare that a step references another step (jumps or complex conditionals), and having the number makes it easier to notice that things need to be offset when inserting or deleting steps of an algorithm. This can be fixed by adding explicit labels to jump to, but we haven't done that.

index.src.html
Djuffin: I dropped `without a copy`

index.src.html
Djuffin: This is a `non-transfer` case that we have in the spec now. It's a copy of the step 11.

padenot (approved):

index.src.html
```diff
@@ -3421,23 +3423,33 @@
...
-10. If `data.byteLength` is less than |combinedLayout|’s
+10. If |combinedLayout| is an exception, throw |combinedLayout|.
+11. If `data.byteLength` is less than |combinedLayout|’s
     [=combined buffer layout/allocationSize=], throw a {{TypeError}}.
-11. Let |resource| be a new [=media resource=] containing a copy of |data|. Use
-        {{VideoFrameBufferInit/visibleRect}} and {{VideoFrameBufferInit/layout}}
+12. For each |transferable| in |init|.{{VideoFrameBufferInit/transfer}}:
+     1. If {{platform object/[[Detached]]}} internal slot is `true`,
+         then throw a {{DataCloneError}} {{DOMException}}.
+13. If |init|.{{VideoFrameBufferInit/transfer}} contains an {{ArrayBuffer}}
```
Djuffin, line 3444: Can we reference  transfer algorithms from structuredClose()  here?

index.src.html
sandersdan: I would prefer if implementations were allowed to copy. In general that's not observable here (unless we add buffer return), but while copying Chrome does try to minimize the coded size and that would be visible if it were not allowed on this path. It also allows us to shift the visible offset to the origin, which is more compatible; not all code is written to handle offsets (because they are rare in practice).

Chrome also expands the Y plane for odd-sized frames so that the coded size is always even. At some point in the future we may trust all of our processing paths enough to remove that, but for now we are likely to prefer to copy in that case.

padenot: I think it's preferable to be explicit, since it's opt-in.

sandersdan: I do not want to throw in edge cases like this, especially the odd-size one which I expect to go away in the future. I don't think developers should need to be aware that transfers can fail and always have to implement fallbacks; I can't imagine the fallback would ever be anything other than trying again without the transfer.

index.src.html
```diff
@@ -3421,23 +3423,33 @@
...
-9. If |combinedLayout| is an exception, throw |combinedLayout|.
-10. If `data.byteLength` is less than |combinedLayout|’s
+10. If |combinedLayout| is an exception, throw |combinedLayout|.
+11. If `data.byteLength` is less than |combinedLayout|’s
     [=combined buffer layout/allocationSize=], throw a {{TypeError}}.
-11. Let |resource| be a new [=media resource=] containing a copy of |data|. Use
-        {{VideoFrameBufferInit/visibleRect}} and {{VideoFrameBufferInit/layout}}
+12. For each |transferable| in |init|.{{VideoFrameBufferInit/transfer}}:
+     1. If {{platform object/[[Detached]]}} internal slot is `true`,
+         then throw a {{DataCloneError}} {{DOMException}}.
```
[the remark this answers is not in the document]
Djuffin: Yeah, you're right it throws.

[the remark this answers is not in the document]
Djuffin: Done

index.src.html
Djuffin: No. It turned out that transfer process is inseparable from closing in structuredClose() spec.

index.src.html
Djuffin: Dropped the wording.

index.src.html
Djuffin: I made transferring optional. Please take a look.

Djuffin:
I updated the text according to the feedback and WG comments. PTAL

sandersdan (approved):

Djuffin:
youennf@ Could you please take another look and approve if you don't have any more comments?

Djuffin:
@youennf  Could you please take another look and approve if you don't have any more comments?

aboba (approved):

PR #679, opened 2023-06-07
Define videoSource before videoSelect.onchange to avoid ReferenceError.

HuaChunbo:
This fixes the error popup when doing video source switch.


chrisn (approved):

chrisn:
Thank you!

PR #685, opened 2023-06-15
Fix usage of RFC2119 words in privacy and security section

padenot:
This fixes #648.


chrisn (commented):
Looks good, although I'm now wondering about this:

> it SHOULD not be possible for a site to mutate a codec input or output 

given the spec includes mitigations as part of the normative requirements. Also, we should probably add "This section is non-normative." to both these sections.

padenot:
So we can't add rfc2119 verbs to non-normative sections see the [manual of style](https://www.w3.org/Guide/manual-of-style/#N-I-sections), so we need to make a decision here.

I think this recommendation is fair (although I'm now wondering why it mentions "output" as well), but I'm wondering what others think, @dsanders11, @youennf?

I think there's agreement that mutating memory that's been passed to as an input is dangerous (that's why there are copies for now, and other mechanisms are being considered), but do implementations prevent mutating the output of codecs (e.g. mutating the encoded bytes after encoding, or mutating the media after decoding)?

chrisn:
> I think this recommendation is fair

Yes, I agree.

Another instance I just noticed is:

> User Agents SHOULD mitigate this risk by extensively fuzzing their implementation

which could be rephrased as "We recommended that user agents mitigate this risk by"

aboba (approved):

chrisn:
Hang on, why merge this when there are unresolved comments?

aboba:
Opened Issue for the remaining items: https://github.com/w3c/webcodecs/issues/689

PR addressing them is here: https://github.com/w3c/webcodecs/pull/690

chrisn:
Thanks! I'll file a PR for https://github.com/w3c/webcodecs/pull/685#issuecomment-1602409644

PR #693, opened 2023-06-23
Add to codec registry requirements

### LINKED ISSUE #426: WebCodecs registration policy
In the [last Media WG meeting](https://www.w3.org/2021/12/14-mediawg-minutes.html#t02) we talked about criteria for adding new WebCodecs registrations. @aboba [suggested](https://lists.w3.org/Archives/Public/public-media-wg/2021Dec/0005.html): 

> 1.  Specification required.  This could be a reference to a specification published by an SDO, or a website (e.g. VP9).  There may be some stability requirement (e.g. a reference to a personal github may not qualify).  The specification should be available at least to the designated Expert Reviewer.
> 2.  Expert review required. The expert could be one of the spec editors, or someone designated by the WG Chairs.  The idea is to designate someone to review the proposal and produce a written review with a recommendation (with the reasoning behind it), posted to the WG mailing list.
> 3.  WG discussion.  The WG (and the submitter) should have the opportunity to respond to the expert review.
> 4.  Consensus determination.  The WG Chairs determine whether the Expert review has consensus or not.

chrisn:
This is intended to resolve #426.

- Added a requirement for a public specification that is stable
- Added that the WG may consult external expertise as part of its review (but not go so far as producing a written review with a recommendation, as #426 suggested)
- Fixes the numbering in the source document

@aboba for review

### DIFF
--- a/codec_registry.src.html
+++ b/codec_registry.src.html
@@ -45,7 +45,7 @@ Registration Entry Requirements {#registration-entry-requirements}
 ==================================================================
 
 1. Each entry must include a unique codec string, a common name string, and a
-    link to a public specification.
+    link to a stable public specification.
 2. The codec string must be crafted as follows:
     1. If the codec string contains a fixed prefix with variable suffix values,
         the suffix must be represented by an asterisk and the registration's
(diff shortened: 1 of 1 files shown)


aboba (changes requested):
Line 48: What does "stable" mean in this context?  Taking the IETF as an example, would an Internet-Draft be considered stable upon WG adoption? Upon approval for publication as an RFC?

chrisn:
This gives a pretty good set of considerations: https://www.w3.org/2013/09/normative-references#stability

But, adding new registrations is pretty infrequent so I'm wondering whether we really need say anything, and leave to the WG evaluation at the time.

aboba:
@chrisn I think we should leave it to WG evaluation. 

Note that there are other requirements in the normative reference gujde that codec specifications may not be able to meet.  For example, the specification for EVC  (,"ISO/IEC 23094-1 Essential Video Coding")  is behind a paywall at https://www.iso.org/standard/57797.html  In IETF, the typical requirement is only that versions of the codec specification can be made available to reviewers upon request.

chrisn:
To allow for paywalled specs we could change the wording from "must include ... a link to a public specification" to "must include ... a reference to the codec's specification" - and we'd similarly want the spec to be made available for review (we have done this from time to time in MEIG).

PR #695, opened 2023-06-25
Add H.265/VP8/VP9/AV1 samples to video decoding page.

StaZhu:
This should be helpful for developer to test the browser implement or compare the performance between different codecs on different platforms and devices.

Note that the previous version of mp4box is not able to parse AV1 video codec string correctly, so to support AV1, we also need to upgrade mp4box to version 0.5.2.

Safari doesn't support `gl.createShader` so change the default renderer to `2d`.

All sample videos are encoded from `bbb-1920x1080-cfg06.mp4` using ffmpeg and shaka packager.


dalecurtis:
We've been meaning to trim down these videos actually. Can you reduce them all to only about 5-10s content, or whatever fits in about ~5mb per clip?

dalecurtis:
Is measuring decoding performance that useful? I think we just want to show how to use the various codecs. Speed testing codecs seems like something folks can trivially figure out.

StaZhu:
> Is measuring decoding performance that useful? I think we just want to show how to use the various codecs. Speed testing codecs seems like something folks can trivially figure out.
How about let's find a compromise point. for example: reduce 20mins -> 4mins, then the H264,VP8 sample will be 18MB approximately, and the H265,VP9,AV1 will be 9MB, is that still too large?

StaZhu:
Okey, I've uploaded the samples to 1mins now, please review the rest ! :)

dalecurtis:
@aboba did you want to review as well?

aboba (approved):

PR #705, opened 2023-07-19
Fix bug for `WebCodecs in Worker` sample page

StaZhu:
`VideoEncoder.isConfigSupported()` sometimes will return result after `transform()` callback's execution, which will resulting `encoder.encode()` fails to encode a VideoFrame.

We can instead change `start()` to an async function, then `transform()` will always be called after `encoder.configure()`.


aboba:
There is a similar issue on the decode side.  I'll submit a fix for that.

PR #707, opened 2023-07-20
Remove ambiguity around valid codec strings during support checks.

### LINKED ISSUE #686: https://wpt.fyi/results/webcodecs/video-encoder-config.https.any.html is expecting VideoEncoder.isConfigSupported to throw
According the spec, VideoEncoder.isConfigSupported only throws in case https://w3c.github.io/webcodecs/#valid-videoencoderconfig returns false.
The test (and Chrome) is expecting to throw in more cases, like if width or height is very big.
Either we should update the spec accordingly or VideoEncoder.isConfigSupported should return supported = false in that case.

dalecurtis:
User agents shouldn't throw a `TypeError` when an unrecognized codec string is 
encountered since this is harmful for interop (e.g., one UA may only recognize 
some versions of a codec string as valid according to its own codec specification).

This change removes the language around checking valid codec strings when 
determining if a config is valid. It instead adds that language explicitly to 
the "Check Configuration Support" algorithm so `false` or `NotSupportedError` 
can be triggered in `isConfigSupported` and`configure` respectively.

Fixes #686

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/pull/707.html" title="Last updated on Jul 21, 2023, 7:08 PM UTC (1e0fbe3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/707/c462aa4...1e0fbe3.html" title="Last updated on Jul 21, 2023, 7:08 PM UTC (1e0fbe3)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -1720,11 +1720,14 @@ Configurations{#configurations}
 <dfn>Check Configuration Support</dfn> (with |config|) {#config-support}
 ------------------------------------------------------------------------
 Run these steps:
-1. If |config| is an {{AudioDecoderConfig}} or {{VideoDecoderConfig}} and the
+1. If the <a>codec string</a> in |config|.codec is not a
+    <a>valid codec string</a> or is otherwise unrecognized by the UA,
+    return `false`.
+2. If |config| is an {{AudioDecoderConfig}} or {{VideoDecoderConfig}} and the
     User Agent can't provide a <a>codec</a> that can decode the exact profile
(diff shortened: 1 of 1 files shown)


Djuffin (approved):

dalecurtis:
Bump @padenot -- can you take a look?

padenot (approved):
Looks good, sorry for the delay.

Issue #719, opened 2023-09-25
The encoder/decoder configure methods is not using a queue to check for configuration support


youennf:
Following on https://github.com/web-platform-tests/interop/issues/382#issuecomment-1731669384 and https://github.com/web-platform-tests/wpt/pull/42127, we could improve the spec related to the configure methods.

In particular:
- configure should probably be using a parallel queue to check for configuration support, like isConfigSupported is.
- once the check is done and config is not supported, configure could queue a task to call the close decoder/encoder algorithm, like done in https://w3c.github.io/webcodecs/#dom-audiodecoder-decode. This would work nicely if we change the close decoder algorithm to synchronously call the error callback.

Other editorial notes while reading the spec (probably applicable to all decoders/encoders) :
- isConfigSupported is trying to resolve a promise from a parallel queue, we could instead queue a task to resolve the promise.
- [[[message queue blocked]]](https://w3c.github.io/webcodecs/#dom-audiodecoder-message-queue-blocked-slot) is accessed from both the work queue and the event task queue, for instance in https://w3c.github.io/webcodecs/#dom-audiodecoder-configure. It might be better to queue a task to set [[[message queue blocked]]](https://w3c.github.io/webcodecs/#dom-audiodecoder-message-queue-blocked-slot) to false and synchronously process the control queue.
- Ditto for [[[codec saturated]]](https://w3c.github.io/webcodecs/#dom-audiodecoder-codec-saturated-slot) which seems to be accessed from both the codec work queue and the event task queue.

youennf:
@dalecurtis, I filed this issue following on our WPT test discussions.
Let me know if that aligns with your thoughts.
I can also submit the corresponding PR(s) if that helps.

dalecurtis:
I'm not sure about `[[codec saturated]]` since that might have performance implications, but the rest sgtm and matches what our implementation (after the recent fixes). Your help with the PRs would be most welcome, thanks!

youennf:
Additional remark related to [[codec implementation]]as spec says it should only be used from the [[codec queue]].
But the spec currently signals or clears [[codec implementation]]  from reset/close algorithms, which can be called from JS directly, without queueing a task to the [[codec queue]].

PR #721, opened 2023-09-27
Update configure methods to use a parallel queue to run the the configuration check support

youennf:
Make close algorithm synchronously call the error callback and ensure the close algorithm is never called from a parallel queue. Update isConfigSupported to stop resolving the promise from a parallel queue.

This should partially fix https://github.com/w3c/webcodecs/issues/719.

***
<a href="https://pr-preview.s3.amazonaws.com/youennf/webcodecs/pull/721.html" title="Last updated on Sep 27, 2023, 1:34 PM UTC (cdc456f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/721/261401a...youennf:cdc456f.html" title="Last updated on Sep 27, 2023, 1:34 PM UTC (cdc456f)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -245,7 +245,7 @@ Each [=codec=] instance has an internal slot named
 
 Each [=codec=] instance has an internal slot named
 <dfn>[[codec implementation]]</dfn> that refers to the underlying platform
-encoder or decoder. Except for assignment, any steps that reference
+encoder or decoder. Except for the initial assignment, any steps that reference
 [=[[codec implementation]]=] will be enqueued to the [=[[codec work queue]]=].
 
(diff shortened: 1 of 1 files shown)


youennf:
This PR is changing configure to do the configuration check from the codec work queue.
Assignment of [[codec implementation]] is now done in the codec work queue, while previously it was made from the JS thread. I am not sure why this was written this way. Does somebody know why it was done so?

I also added `if needed` since we might not always want to create a new [[codec implementation]] each time configure is called (say only bitrate is changing for instance).

dalecurtis (approved):

PR #738, opened 2023-10-31
Resolve VideoFrame.copyTo with the list of layout

### LINKED ISSUE #510: `VideoFrame.copyTo` doesn't resolve its `Promise` with a value
We want to resolve with the `sequence<PlaneLayout>`: we compute the underlying values but we don't use them.

padenot:
This fixes #510.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/pull/738.html" title="Last updated on Oct 31, 2023, 2:32 PM UTC (441d96c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/738/3c9e353...441d96c.html" title="Last updated on Oct 31, 2023, 2:32 PM UTC (441d96c)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -3744,7 +3744,8 @@ A <dfn>computed plane layout</dfn> is a [=struct=] that consists of:
         {{TypeError}}.
     6. Let |p| be a new {{Promise}}.
     7. Let |copyStepsQueue| be the result of starting a new [=parallel queue=].
-    8. Enqueue the following steps to |copyStepsQueue|:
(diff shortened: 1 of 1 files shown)


Djuffin (approved):
Both Safari and Chromium do it this way. It should be a part of the spec.

padenot:
> Both Safari and Chromium do it this way. It should be a part of the spec.

It was part of the spec, it was clear from the signature all along, but we forgot to resolve with the value when writing this.

PR #752, opened 2023-12-06
Revert "Add screen content coding tools as an encoder configuration"

Djuffin:
`av1.forceScreenContentTools` was made practically obsolete by [contentHint](https://www.w3.org/TR/webcodecs/#dom-videoencoderconfig-contenthint), which does the same but in a codec-agnostic way.

It hasn't been implemented by any UA, and it's safe to remove from the spec.

This reverts commits:
b35c6a36cb45ceaea87344702aa1c53430daaa56
29f1a8a165c2a16dc9a0b7180bd23c5ac2f07215

Related issue: https://github.com/w3c/webcodecs/issues/646

### DIFF
--- a/av1_codec_registration.src.html
+++ b/av1_codec_registration.src.html
@@ -75,47 +75,6 @@ If an {{EncodedVideoChunk}}'s {{EncodedVideoChunk/[[type]]}} is
 contain a frame with a `frame_type` of `KEY_FRAME` as defined in Section
 6.8.2 of [[AV1]].
 
-VideoEncoderConfig extensions {#videoencoderconfig-extensions}
-==============================================================
-
-<pre class='idl'>
-<xmp>
-partial dictionary VideoEncoderConfig {
-  AV1EncoderConfig av1;
-};
-</xmp>
-</pre>
-
-<dl>
-  <dt><dfn dict-member for=VideoEncoderConfig>av1</dfn></dt>
-  <dd>
-    Contains codec specific configuration options for the AV1 codec.
-  </dd>
-</dl>
-
-AV1EncoderConfig {#av1-encoder-config}
---------------------------------------
-<pre class='idl'>
-<xmp>
-dictionary AV1EncoderConfig {
-  boolean forceScreenContentTools = false;
-};
-</xmp>
-</pre>
-
-<dl>
-  <dt><dfn dict-member for=AV1EncoderConfig>forceScreenContentTools</dfn></dt>
-  <dd>
-     Indicates whether the encoder should force use of screen content
-     coding tools. The default value (false) indicates that use of
-     screen content coding tools is not forced. A value of true
-     (corresponding to setting seq_force_screen_content_tools
-     to SELECT_SCREEN_CONTENT_TOOLS in Section 5.5.1
-     of [[AV1]]) indicates that use of screen content coding tools
-     is forced.
-  </dd>
-</dl>
-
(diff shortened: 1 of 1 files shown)


Djuffin:
cc @fippo

aboba:
The reason for including per-codec SCC configuration settings was to be able to detect if they were supported.  Our data shows that only a fraction of AV1 hardware acceleration implementations support acceleration of screen content coding. 

Using contentHint, is there a way to determine whether the desired hint can be applied?  For example, if a config contains `hwAcceleration="prefer-hardware"` and `contentHint: "text"`, and an attempt to configure a tool (e.g. palette or block copy) fails, does anything returned by `isConfigSupported` indicate an inability to apply the hint?  For example, `contentHint: "text"` could be omitted if the hint couldn't be applied.

fippo:
You [beat me to it](https://github.com/w3c/webcodecs/compare/main...fippo:webcodecs:remove-forceScreenContentTools?expand=1) and reverts are better ;-)

padenot:
@aboba's point is valid though. [Quoting](https://w3c.github.io/webcodecs/#dom-videoencoderconfig-contenthint) the spec:

> The User Agent MUST NOT refuse the configuration if it doesn’t support this content hint. See [isConfigSupported()](https://w3c.github.io/webcodecs/#dom-videoencoder-isconfigsupported).

aboba:
Since software implementations of AV1 encode/decode support SCC, the problem arises in situations where a (non-compliant) AV1 hardware encoder or decoder lacks SCC support.  In the case of a hardware encoder, the result will be sending at a higher bitrate than might be necessary with SCC, but probably still lower than alternatives, so it wouldn't influence codec choice.  However, in the case of a hardware decoder that doesn't support SCC, the result could be a decode failure, and in that situation, it would make sense to turn off SCC on the encoder. But that requires knowledge of whether the encoder is encoding with SCC and also information about that the decoder error was due to an inability to decode SCC.

taste1981:
For AV1 decoder SCC must be supported as part of conformance (unlike H.265 which contains SCC support in the form of a profile), so it is not likely for decoder to raise such "error".

There are platforms that supports AV1 encoding, but does not allow you to turn on screen content tools. I think the knob here still makes sense to me.

Djuffin:
Given that we already expose SCC via `contentHint`, I'm not convinced we really need an explicit knob here. @aboba if I remember correctly you wanted to check how many decoders fail to decode content using SCC. Do you still think that we shouldn't use `contentHint` for it. BTW it's already shipped in Chromium

aboba:
@Djuffin SCC support is a conformance requirement for AV1 decoders, and so far we have not encountered AV1 decoders that do not support SCC.  WebCodecs is supported in Safari Tech Preview, and `isConfigSupported()` indicates support for AV1 hardware accelerated decode in newer devices such as the iPhone Pro 15 as well as M3-based MacBooks. Support for AV1 hardware accelerated encode is not available on iOS or Android mobile devices (or notebooks, for that matter) so I can't speak to the prevalence of SCC support on AV1 hardware encoders.

aboba (approved):
The question here is whether it is useful to the application to know that SCC encoding cannot be supported (e.g. if a hardware encoder does not support SCC). In all likelihood, the application will choose a workaround along the lines of what is provided in ContentHints. Unless there is a reason to believe that the application can do better, I don't see the downside to the ContentHints approach.

fippo (approved):

Issue #92, opened 2020-10-16
API for conversion between pixel formats


chcunningham:
Currently the spec lists one format: I420. We'll want more, including some format to indicate "opaque", which will be used for GPU backed frames to avoid having to always copy to CPU memory. 

Particularly for "opaque", we want an API to convert to another format where the planes can be read (performing the copy to CPU along the way). This API could also be used to convert between various non-opaque formats. Something like...

let frame = await gpu_frame.convertTo(I420);

chcunningham:
The urgency of this API is reduced. Previously we expected this would be primary means to read back pixel data from GPU backed frames into the CPU space. Since then we've managed to add support (in chromium, not yet in the spec) for exposing pixel data from GPU backed video frames coming off of the users camera in the NV12 pixel format. 

A conversion API is still interesting to pursue, but there are no urgent use cases.

padenot:
> (in chromium, not yet in the spec)

https://github.com/w3c/webcodecs/pull/208, for reference, but this would still be good to have.

chcunningham:
triage note: marking extension, as this would be a new API.

alvestrand:
this means, of course, that any application that handles only a limited set of formats will be broken every time a new format is encountered. Seems a rough deal for app developers.

sandersdan:
> any application that handles only a limited set of formats will be broken every time a new format is encountered

There is a fallback path; you can createImageBitmap(videoFrame), then read back that ImageBitmap (presumably as RGBA).

This isn't great if the frame was mappable in the first place, but for GPU-backed frames it actually has acceptable performance.

alvestrand:
So this means that we have a fallback (effectively "read as RGBA"). Apps can choose between using a fast path if they support the current format and a slow path if they don't. For a lot of cases, this will be sufficient.

chcunningham:
Options for potential chromium impl in https://bugs.chromium.org/p/chromium/issues/detail?id=1139119

Kaiido:
> There is a fallback path; you can createImageBitmap(videoFrame), then read back that ImageBitmap (presumably as RGBA).

Is there anything that would enforce that conversion? Currently the [HTML specs](https://html.spec.whatwg.org/multipage/imagebitmap-and-animations.html#images-2:videoframe) for `createImageBitmap(VideoFrame)` ask to

[...]

PR #754, opened 2023-12-07
Spec optional conversion to RGB in VideoFrame.copyTo()

Djuffin:
This change should address the [issue](https://github.com/w3c/webcodecs/issues/92) of pixel format conversion in `VideoFrame.copyTo()`.

***
<a href="https://pr-preview.s3.amazonaws.com/Djuffin/webcodecs/pull/754.html" title="Last updated on Apr 16, 2024, 8:43 AM UTC (0df4aee)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/754/871ccb3...Djuffin:0df4aee.html" title="Last updated on Apr 16, 2024, 8:43 AM UTC (0df4aee)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -3767,9 +3767,8 @@ A <dfn>computed plane layout</dfn> is a [=struct=] that consists of:
     5. If `destination.byteLength` is less than |combinedLayout|'s [=combined
         buffer layout/allocationSize=], return a promise rejected with a
         {{TypeError}}.
-    6. If |options|.{{VideoFrameCopyToOptions/format}} does not equal
-        {{VideoFrame/[[format]]}} and |options|.{{VideoFrameCopyToOptions/format}}
-        is equal to one of {{RGBA}}, {{RGBX}}, {{BGRA}}, {{BGRX}} then:
+    6. If |options|.{{VideoFrameCopyToOptions/format}} is equal to one of
+        {{RGBA}}, {{RGBX}}, {{BGRA}}, {{BGRX}} then:
           1. Let |newOptions| be the result of running the <a>Clone Configuration</a>
                algorithm with |options|.
           2. Assign `undefined` to |newOptions|.{{VideoFrameCopyToOptions/format}}
@@ -3782,10 +3781,12 @@ A <dfn>computed plane layout</dfn> is a [=struct=] that consists of:
                algorithm with |options|.{{VideoFrameCopyToOptions/format}}, |rect|, and  `options.colorSpace`.
           7. Return the result of calling {{VideoFrame/copyTo()}} on |rgbFrame| with
                |destination| and |newOptions|.
-    7. Let |p| be a new {{Promise}}.
-    8. Let |copyStepsQueue| be the result of starting a new [=parallel queue=].
-    9. Let |planeLayouts| be a new [=list=].
-    10. Enqueue the following steps to |copyStepsQueue|:
+    7. Otherwise, if |options|.{{VideoFrameCopyToOptions/format}} is not `undefined` or `null`
+        throw a {{NotSupportedError}} {{DOMException}}.
+    8. Let |p| be a new {{Promise}}.
+    9. Let |copyStepsQueue| be the result of starting a new [=parallel queue=].
+    10. Let |planeLayouts| be a new [=list=].
+    11. Enqueue the following steps to |copyStepsQueue|:
         1. Let resource be the [=media resource=] referenced by
             [[resource reference]].
         2. Let |numPlanes| be the number of planes as defined by
@@ -3822,7 +3823,7 @@ A <dfn>computed plane layout</dfn> is a [=struct=] that consists of:
             10. Increment |planeIndex| by `1`.
             11. Append |layout| to |planeLayouts|.
         5. [=Queue a task=] to resolve |p| with |planeLayouts|.
-    11. Return |p|.
+    12. Return |p|.
 
 : <dfn method for=VideoFrame>clone()</dfn>
 :: Creates a new {{VideoFrame}} with a reference to the same
@@ -4328,9 +4329,9 @@ will enforce the following requirements:
     invalid to specify planes that overlap.
 : <dfn dict-member for=VideoFrameCopyToOptions>format</dfn>
 :: A {{VideoPixelFormat}} for the pixel data in the destination
-    {{BufferSource}}. Potential values are:
-    {{RGBA}}, {{RGBX}}, {{BGRA}}, {{BGRX}} and values of {{VideoFrame/format}}.
-    If unspecified or `null`, the {{VideoFrame/format}} is used.
+    {{BufferSource}}. Potential values are: {{RGBA}}, {{RGBX}}, {{BGRA}},
+    {{BGRX}}. If unspecified or `null`, the the destination
+    {{BufferSource}} will match {{VideoFrame/format}} .
 : <dfn dict-member for=VideoFrameCopyToOptions>colorSpace</dfn>
 :: A {{PredefinedColorSpace}} that <em class="rfc2119">SHALL</em> be used as
     a target color space for the pixel data in the destination
--- a/index.src.html
+++ b/index.src.html
@@ -3767,7 +3767,7 @@ A <dfn>computed plane layout</dfn> is a [=struct=] that consists of:
     5. If `destination.byteLength` is less than |combinedLayout|'s [=combined
         buffer layout/allocationSize=], return a promise rejected with a
         {{TypeError}}.
-    6. If |options|.{{VideoFrameCopyToOptions/format}} does not equal to
+    6. If |options|.{{VideoFrameCopyToOptions/format}} does not equal
         {{VideoFrame/[[format]]}} and |options|.{{VideoFrameCopyToOptions/format}}
         is equal to one of {{RGBA}}, {{RGBX}}, {{BGRA}}, {{BGRX}} then:
           1. Let |newOptions| be the result of running the <a>Clone Configuration</a>
@@ -4214,25 +4214,25 @@ A <dfn>computed plane layout</dfn> is a [=struct=] that consists of:
     2. Let |canvasSettings| be a new {{CanvasRenderingContext2DSettings}} with
         {{CanvasRenderingContext2DSettings/colorSpace}} set to |colorSpace|.
     3. Let |context| be the result of calling {{OffscreenCanvas/getContext()}}
-        on |canvas| with `"2d"` and |canvasSettings|
+        on |canvas| with `"2d"` and |canvasSettings|.
     3. Call {{CanvasDrawImage/drawImage()}} on |context| with
         `rect.x`, `rect.y`, `rect.width`, `rect.height`, `0`, `0`, `rect.width`,
         `rect.height`.
     4. Let |imageData| be the result of calling {{CanvasImageData/getImageData()}} with
         `0`, `0`, |rect.width|, |rect.height|.
     5. If |format| is equal to {{BGRA}} or {{BGRX}}:
-        1. Let |pixelIndex| be `0`
+        1. Let |pixelIndex| be `0`.
         2. While |pixelIndex| is less than `rect.width` multiplied by `rect.height`:
             1. Swap elements in |imageData|.{{ImageData/data}} at indexes `4 * pixelIndex`
                  and `4 * pixelIndex + 2`.
-            2. Increment |pixelIndex| by `1`
+            2. Increment |pixelIndex| by `1`.
     6. Let |init| be a new {{VideoFrameBufferInit}} constructed with
         {{VideoFrameBufferInit/format}} = |format|,
         {{VideoFrameBufferInit/codedWidth}} = `rect.width`,
-        {{VideoFrameBufferInit/codedHeight}} = `rect.height`
+        {{VideoFrameBufferInit/codedHeight}} = `rect.height`.
     7. Let |convertedFrame| be a new {{VideoFrame}} constructed with
-        |imageData|.{{ImageData/data}} and |init|
-    8. Return |convertedFrame|
+        |imageData|.{{ImageData/data}} and |init|.
+    8. Return |convertedFrame|.
 
 : <dfn for=VideoFrame>Copy VideoFrame metadata</dfn> (with |metadata|)
 :: 1. Let |metadataCopySerialized| be [$StructuredSerialize$](|metadata|).
@@ -4290,7 +4290,7 @@ value to {{ColorSpaceConversion/"none"}} disables color space conversion.
 
 VideoFrame CopyTo() Options {#videoframe-copyto-options}
 ------------------------------------------------------------
-Options to specify a rectangle of pixels to copy, their format and the offset
+Options to specify a rectangle of pixels to copy, their format, and the offset
 and stride of planes in the destination buffer.
 
 <xmp class='idl'>
@@ -4332,8 +4332,8 @@ will enforce the following requirements:
     {{RGBA}}, {{RGBX}}, {{BGRA}}, {{BGRX}} and values of {{VideoFrame/format}}.
     If unspecified or `null`, the {{VideoFrame/format}} is used.
 : <dfn dict-member for=VideoFrameCopyToOptions>colorSpace</dfn>
-:: A {{PredefinedColorSpace}} that <em class="rfc2119">SHALL</em> be used to as
-    a target color space for for the pixel data in the destination
+:: A {{PredefinedColorSpace}} that <em class="rfc2119">SHALL</em> be used as
+    a target color space for the pixel data in the destination
     {{BufferSource}}, but only if {{VideoFrameCopyToOptions/format}} is one of
     {{RGBA}}, {{RGBX}}, {{BGRA}}, {{BGRX}}, otherwise it is ignored.
     If unspecified or `null`, {{srgb}} is used.
--- a/index.src.html
+++ b/index.src.html
@@ -4322,7 +4322,7 @@ will enforce the following requirements:
 : <dfn dict-member for=VideoFrameCopyToOptions>format</dfn>
 :: A {{VideoPixelFormat}} for the pixel data in the destination
     {{BufferSource}}. Potential values are: {{RGBA}}, {{RGBX}}, {{BGRA}},
-    {{BGRX}}. If it does not [=map/exist=], the the destination
+    {{BGRX}}. If it does not [=map/exist=], the destination
     {{BufferSource}} will be in the same format as {{VideoFrame/format}} .
 : <dfn dict-member for=VideoFrameCopyToOptions>colorSpace</dfn>
 :: A {{PredefinedColorSpace}} that <em class="rfc2119">MUST</em> be used as
--- a/index.src.html
+++ b/index.src.html
@@ -3748,8 +3748,8 @@ A <dfn>computed plane layout</dfn> is a [=struct=] that consists of:
 
 : <dfn method for=VideoFrame>copyTo(|destination|, |options|)</dfn>
 :: Asynchronously copies the planes of this frame into |destination| according
-    to |options|. The format of the data is specified by {{VideoFrameCopyToOptions/format}},
-    if present, otherwise {{VideoFrame/format}}.
+    to |options|. The format of the data is |options|.{{VideoFrameCopyToOptions/format}},
+    if it [=map/exists=] or [=this=] {{VideoFrame}}'s {{VideoFrame/format}} otherwise.
 
     NOTE: Promises that are returned by several calls to
         {{VideoFrame/copyTo()}} are not guaranteed to resolve in the order they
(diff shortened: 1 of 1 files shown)


Djuffin:
There are probably some missing details, typos and omissions.
But I'd love to get some feedback on the general approach in this change.

index.src.html
```diff
@@ -3728,8 +3730,8 @@
 
 : <dfn method for=VideoFrame>copyTo(|destination|, |options|)</dfn>
 :: Asynchronously copies the planes of this frame into |destination| according
-    to |options|. The format of the data is the same as this {{VideoFrame}}'s
-    {{VideoFrame/format}}.
+    to |options|. The format of the data is specified by {{VideoFrameCopyToOptions/format}},
+    if present, or the same as {{VideoFrame/format}}.
```
sandersdan, line 3734: Recommend replacing "or" with "otherwise".

index.src.html
```diff
@@ -4166,10 +4185,36 @@
...
+        on |canvas| with `"2d"` and |canvasSettings|
+    3. Call {{CanvasDrawImage/drawImage()}} on |context| with
+        `rect.x`, `rect.y`, `rect.width`, `rect.height`, `0`, `0`, `rect.width`,
+        `rect.height`.
+    4. Let |imageData| be the result of calling {{CanvasImageData/getImageData()}} with
+        `0`, `0`, |rect.width|, |rect.height|.
+    5. If |format| is equal to {{BGRA}} or {{BGRX}}:
+        1. Let |pixelIndex| be `0`
+        2. While |pixelIndex| is less than `rect.width` multiplied by `rect.height`:
+            1. Swap items in |imageData|.{{ImageData/data}} at indexes `4 * pixelIndex`
```
sandersdan, line 4203: Nit: `items` is perhaps unclear, recommend "elements".

index.src.html
```diff
@@ -4256,6 +4303,16 @@
     to specify an offset and stride for each plane in the destination
     {{BufferSource}}. If unspecified, the planes will be tightly packed. It is
     invalid to specify planes that overlap.
+: <dfn dict-member for=VideoFrameCopyToOptions>format</dfn>
+:: A {{VideoPixelFormat}} for the pixel data in the destination
+    {{BufferSource}}. Explicity values cab be:
+    {{RGBA}}, {{RGBX}}, {{BGRA}}, {{BGRX}} and the same as {{VideoFrame/format}}.
```
sandersdan, line 4309: What happens if `frame.format` is `null`?

index.src.html
```diff
@@ -4256,6 +4303,16 @@
...
+    If unspecified, the {{VideoFrame/format}} is used.
+: <dfn dict-member for=VideoFrameCopyToOptions>colorSpace</dfn>
+:: A {{PredefinedColorSpace}} that <em class="rfc2119">SHALL</em> be used to as
+    a target color space for for the pixel data in the destination
+    {{BufferSource}}.
```
sandersdan, line 4314: What happens if this member is not present when doing a conversion?

What happens if it is present when not doing a conversion?

index.src.html
aboba: "or the same as" -> "otherwise"

index.src.html
aboba: "Explicity values cab" -> "Potential values are"
"and the same as" -> "and other values of"

index.src.html
Djuffin: I added explanation around it.

index.src.html
```diff
@@ -4166,10 +4191,36 @@
...
+        {{CanvasRenderingContext2DSettings/colorSpace}} set to |colorSpace|.
+    3. Let |context| be the result of calling {{OffscreenCanvas/getContext()}}
+        on |canvas| with `"2d"` and |canvasSettings|
+    3. Call {{CanvasDrawImage/drawImage()}} on |context| with
+        `rect.x`, `rect.y`, `rect.width`, `rect.height`, `0`, `0`, `rect.width`,
+        `rect.height`.
+    4. Let |imageData| be the result of calling {{CanvasImageData/getImageData()}} with
+        `0`, `0`, |rect.width|, |rect.height|.
+    5. If |format| is equal to {{BGRA}} or {{BGRX}}:
+        1. Let |pixelIndex| be `0`
```
chrisn, line 4207:
```suggestion
        1. Let |pixelIndex| be `0`.
```

index.src.html
```diff
@@ -4166,10 +4191,36 @@
...
+        2. While |pixelIndex| is less than `rect.width` multiplied by `rect.height`:
+            1. Swap elements in |imageData|.{{ImageData/data}} at indexes `4 * pixelIndex`
+                 and `4 * pixelIndex + 2`.
+            2. Increment |pixelIndex| by `1`
```
chrisn, line 4211:
```suggestion
            2. Increment |pixelIndex| by `1`.
```

index.src.html
```diff
@@ -4166,10 +4191,36 @@
...
+    6. Let |init| be a new {{VideoFrameBufferInit}} constructed with
+        {{VideoFrameBufferInit/format}} = |format|,
+        {{VideoFrameBufferInit/codedWidth}} = `rect.width`,
+        {{VideoFrameBufferInit/codedHeight}} = `rect.height`
```
chrisn, line 4215:
```suggestion
        {{VideoFrameBufferInit/codedHeight}} = `rect.height`.
```

index.src.html
```diff
@@ -4166,10 +4191,36 @@
...
+    7. Let |convertedFrame| be a new {{VideoFrame}} constructed with
+        |imageData|.{{ImageData/data}} and |init|
```
chrisn, line 4217:
```suggestion
        |imageData|.{{ImageData/data}} and |init|.
```

index.src.html
```diff
@@ -4166,10 +4191,36 @@
...
+    8. Return |convertedFrame|
```
chrisn, line 4218:
```suggestion
    8. Return |convertedFrame|.
```

index.src.html
```diff
@@ -4222,13 +4273,15 @@
 
 VideoFrame CopyTo() Options {#videoframe-copyto-options}
 ------------------------------------------------------------
-Options to specify a rectangle of pixels to copy and the offset and stride of
-planes in the destination buffer.
+Options to specify a rectangle of pixels to copy, their format and the offset
```
chrisn, line 4276:
```suggestion
Options to specify a rectangle of pixels to copy, their format, and the offset
```

index.src.html
```diff
@@ -4256,6 +4309,18 @@
     to specify an offset and stride for each plane in the destination
     {{BufferSource}}. If unspecified, the planes will be tightly packed. It is
     invalid to specify planes that overlap.
+: <dfn dict-member for=VideoFrameCopyToOptions>format</dfn>
+:: A {{VideoPixelFormat}} for the pixel data in the destination
+    {{BufferSource}}. Potential values are:
+    {{RGBA}}, {{RGBX}}, {{BGRA}}, {{BGRX}} and values of {{VideoFrame/format}}.
+    If unspecified or `null`, the {{VideoFrame/format}} is used.
+: <dfn dict-member for=VideoFrameCopyToOptions>colorSpace</dfn>
+:: A {{PredefinedColorSpace}} that <em class="rfc2119">SHALL</em> be used to as
```
chrisn, line 4318:
```suggestion
:: A {{PredefinedColorSpace}} that <em class="rfc2119">SHALL</em> be used as
```

index.src.html
```diff
@@ -4256,6 +4309,18 @@
...
+    a target color space for for the pixel data in the destination
```
chrisn, line 4319:
```suggestion
    a target color space for the pixel data in the destination
```

chrisn (changes requested):

index.src.html
```diff
@@ -4256,6 +4309,18 @@
...
+: <dfn dict-member for=VideoFrameCopyToOptions>format</dfn>
+:: A {{VideoPixelFormat}} for the pixel data in the destination
+    {{BufferSource}}. Potential values are:
+    {{RGBA}}, {{RGBX}}, {{BGRA}}, {{BGRX}} and values of {{VideoFrame/format}}.
+    If unspecified or `null`, the {{VideoFrame/format}} is used.
+: <dfn dict-member for=VideoFrameCopyToOptions>colorSpace</dfn>
+:: A {{PredefinedColorSpace}} that <em class="rfc2119">SHALL</em> be used as
+    a target color space for the pixel data in the destination
+    {{BufferSource}}, but only if {{VideoFrameCopyToOptions/format}} is one of
+    {{RGBA}}, {{RGBX}}, {{BGRA}}, {{BGRX}}, otherwise it is ignored.
```
sandersdan, line 4321: It's also necessary that format differs from the frame format, otherwise the result will be in the frame's color space.

We should probably enforce that the color space is only set in the conversion case, because some day we might want to add color conversion to the other paths as well.

Djuffin: Good point. I rewrote it so if `format` is explicitly set it's always a conversion case.

Djuffin:
@padenot Could you please take a look? I'd like to merge this PR some time soon.

index.src.html
```diff
@@ -4179,10 +4205,36 @@
             [=combined buffer layout/allocationSize=].
     9. Return |combinedLayout|.
 
+: <dfn for=VideoFrame>Convert to RGB frame</dfn> (with |format|, |rect| and |colorSpace|)
+ :: 1. Let |canvas| be a new {{OffscreenCanvas}} constructed with
+        |rect|.|width| and |rect|.|height|.
```
padenot, line 4210: This isn't something that we can do in spec, using other specs directly like that. All the objects you use can have been overridden in the global this algorithm is running in.

We should use internal stops for canvas instead.

index.src.html
```diff
@@ -4269,6 +4323,18 @@
     to specify an offset and stride for each plane in the destination
     {{BufferSource}}. If unspecified, the planes will be tightly packed. It is
     invalid to specify planes that overlap.
+: <dfn dict-member for=VideoFrameCopyToOptions>format</dfn>
+:: A {{VideoPixelFormat}} for the pixel data in the destination
+    {{BufferSource}}. Potential values are: {{RGBA}}, {{RGBX}}, {{BGRA}},
+    {{BGRX}}. If unspecified or `null`, the the destination
+    {{BufferSource}} will match {{VideoFrame/format}} .
+: <dfn dict-member for=VideoFrameCopyToOptions>colorSpace</dfn>
+:: A {{PredefinedColorSpace}} that <em class="rfc2119">SHALL</em> be used as
```
padenot, line 4332: Why not must?

padenot (changes requested):

index.src.html
Djuffin: I'm probably missing something here. Could you elaborate a bit?

Isn't it what we do in the Rendering section?

referring to [CanvasDrawImage](https://html.spec.whatwg.org/multipage/canvas.html#canvasdrawimage) [drawImage()](https://html.spec.whatwg.org/multipage/canvas.html#dom-context-2d-drawimage)

padenot: No, that's unrelated, we reference it in your link. What happens when `drawImage` is called with a `VideoFrame` is well-defined there. Here you're _using_ another Web Platform object from the innards of our spec.

e.g. what happens if in the document we're doing this conversion, I've maliciously done:

```js
window.OffscreenCanvas.prototype = function() { alert("what now"); };
```

before anything from our spec run?

Djuffin: This algorithm does not refer to any particular js code. It explicitly says what object needs to be created by the user agent. That's why I don't think that global overrides should concern us here. With a similar approach half of this spec can be put into question with code like 

```
window.VideoFrame.prototype = function() { alert("what now"); };
```

index.src.html
Djuffin: done

index.src.html
Djuffin: I added a note clarifying that UAs don't have to do it via literally creating an OffscreenCanvas as long as the result is the same. I hope this eliminates possible confusion. 

If you still think that the wording leaves room for misinterpretation, I'm very much open to suggestions how to rewrite it.

padenot: This is not about "interpretation", it's simply not something we do in specs, and this needs to be rewritten with an abstract operation exported from the canvas code (if that's what you want to use), and calling into this abstract operation.

See https://github.com/whatwg/html/pull/5510#pullrequestreview-411744788, for a reference of this happening, and https://github.com/w3c/IntersectionObserver/issues/427 for the fix.

Djuffin: Thanks for the links!

I've noticed that [drawImage()](https://html.spec.whatwg.org/multipage/canvas.html#drawing-images) (step 6) does not specify any details about the rendering (sampling, color conversion etc). It's only saying that "The user agent may use any filtering algorithm". 
So there isn't really an algorithm to extract and call. 
I will have to rewrite it in a more abstract manner, similar to the text for `drawImage()`

padenot: We can add an export to something hand-wavy from canvas now, and merge what we want to do here, and then fix the canvas spec as well, your call.

chrisn (approved):
Looks good, but let's resolve the discussion on overridden globals (I don't mind if that's here or in a follow up PR).

index.src.html
Djuffin: I [asked](https://github.com/whatwg/html/issues/10210) HTML spec owners about the best way to reuse canvas pixel format conversion algorithm.

index.src.html
```diff
@@ -3764,10 +3767,22 @@
...
+          1. Let |newOptions| be the result of running the <a>Clone Configuration</a>
+               algorithm with |options|.
+          2. Assign `undefined` to |newOptions|.{{VideoFrameCopyToOptions/format}}.
+          3. Let |rgbFrame| be the result of running the [=VideoFrame/Convert to RGB frame=]
+               algorithm with [=this=], |options|.{{VideoFrameCopyToOptions/format}},
+               and |options|.{{VideoFrameCopyToOptions/colorSpace}}.
+          4. Return the result of calling {{VideoFrame/copyTo()}} on |rgbFrame| with
+               |destination| and |newOptions|.
+    7. Otherwise, if |options|.{{VideoFrameCopyToOptions/format}} is not `undefined` or `null`
+        throw a {{NotSupportedError}} {{DOMException}}.
```
Kaiido, line 3781: This also affects the prior steps 1 & 2, but I doubt you can throw directly here. A method that returns a Promise is supposed to return a rejected Promise instead of throwing (like is done in steps 4 and 5).

Djuffin: Good point. Thanks

index.src.html
Djuffin: I've looked at the rest of the WebCodecs spec. It's pretty light on the details when it comes to format and color spaces conversions. (for example [Rendering](https://www.w3.org/TR/webcodecs/#videoframe-rendering) and [AudioData.copyTo](https://www.w3.org/TR/webcodecs/#dom-audiodata-copyto) )

So I rewrote "Convert to RGB frame" in the similar manner. 
Please take another look.

Djuffin: @padenot gentle ping

index.src.html
```diff
@@ -3745,29 +3748,41 @@
 
 : <dfn method for=VideoFrame>copyTo(|destination|, |options|)</dfn>
 :: Asynchronously copies the planes of this frame into |destination| according
-    to |options|. The format of the data is the same as this {{VideoFrame}}'s
-    {{VideoFrame/format}}.
+    to |options|. The format of the data is specified by {{VideoFrameCopyToOptions/format}},
+    if present, otherwise {{VideoFrame/format}}.
```
padenot, line 3752: When looking at how this is rendered in HTML, I find it very confusing, can you rephrase it or make it more explicit or something?

index.src.html
```diff
@@ -3745,29 +3748,41 @@
...
+          1. Let |newOptions| be the result of running the <a>Clone Configuration</a>
+               algorithm with |options|.
+          2. Assign `undefined` to |newOptions|.{{VideoFrameCopyToOptions/format}}.
+          3. Let |rgbFrame| be the result of running the [=VideoFrame/Convert to RGB frame=]
+               algorithm with [=this=], |options|.{{VideoFrameCopyToOptions/format}},
+               and |options|.{{VideoFrameCopyToOptions/colorSpace}}.
+          4. Return the result of calling {{VideoFrame/copyTo()}} on |rgbFrame| with
+               |destination| and |newOptions|.
+    7. Otherwise, if |options|.{{VideoFrameCopyToOptions/format}} is not `undefined` or `null`
+        return a promise rejected with a {{NotSupportedError}} {{DOMException}}.
```
padenot, line 3781: We can use ["exists"](https://infra.spec.whatwg.org/#map-exists) (often used for `dictionaries`) here but we should also check for other invalid values. But isn't that duplicate work with "Parse VideoFrameCopyToOptions" invoked above? You've modified it appropriately.

index.src.html
```diff
@@ -4048,10 +4063,17 @@
     6. Let |optLayout| be `undefined`.
     7. If |options|.{{VideoFrameCopyToOptions/layout}} [=map/exists=], assign
         its value to |optLayout|.
-    8. Let |combinedLayout| be the result of running the [=VideoFrame/Compute
-        Layout and Allocation Size=] algorithm with |parsedRect|,
-        {{VideoFrame/[[format]]}}, and |optLayout|.
-    9. Return |combinedLayout|.
+    8. Let |format| be `undefined`.
+    9. If |options|.{{VideoFrameCopyToOptions/format}} is `undefined`,
+        assign {{VideoFrame/[[format]]}} to |format|.
```
padenot, line 4068: Exists again here.

index.src.html
```diff
@@ -4183,10 +4205,34 @@
...
+            {{VideoFrame/[[visible top]]}}, {{VideoFrame/[[visible width]]}},
+            and {{VideoFrame/[[visible height]]}} respectively.
+	      6. Assign |frame|'s {{VideoFrame/[[duration]]}} and |frame|'s
+            {{VideoFrame/[[timestamp]]}} to {{VideoFrame/[[duration]]}} and
+            {{VideoFrame/[[timestamp]]}} respectively.
+        7. Assign |colorSpace| to {{VideoFrame/[[color space]]}}.
+        8. Let |resource| be a new [=media resource=] containing the result of
+            conversion of [=media resource=] referenced by |frame|'s
+            {{VideoFrame/[[resource reference]]}} into a color space and pixel
+            format specified by |colorSpace| and |format| respectively.
```
padenot, line 4228: As good as it can be for now I guess, but thanks for checking everything on HTML and all.

index.src.html
```diff
@@ -4273,6 +4321,18 @@
...
+:: A {{VideoPixelFormat}} for the pixel data in the destination
+    {{BufferSource}}. Potential values are: {{RGBA}}, {{RGBX}}, {{BGRA}},
+    {{BGRX}}. If unspecified or `null`, the the destination
+    {{BufferSource}} will match {{VideoFrame/format}} .
+: <dfn dict-member for=VideoFrameCopyToOptions>colorSpace</dfn>
+:: A {{PredefinedColorSpace}} that <em class="rfc2119">MUST</em> be used as
+    a target color space for the pixel data in the destination
+    {{BufferSource}}, but only if {{VideoFrameCopyToOptions/format}} is one of
+    {{RGBA}}, {{RGBX}}, {{BGRA}}, {{BGRX}}, otherwise it is ignored.
+    If unspecified or `null`, {{srgb}} is used.
```
padenot, line 4334: Exists

index.src.html
```diff
@@ -4273,6 +4321,18 @@
     to specify an offset and stride for each plane in the destination
     {{BufferSource}}. If unspecified, the planes will be tightly packed. It is
     invalid to specify planes that overlap.
+: <dfn dict-member for=VideoFrameCopyToOptions>format</dfn>
+:: A {{VideoPixelFormat}} for the pixel data in the destination
+    {{BufferSource}}. Potential values are: {{RGBA}}, {{RGBX}}, {{BGRA}},
+    {{BGRX}}. If unspecified or `null`, the the destination
```
padenot, line 4327: Exists

index.src.html
```diff
@@ -4273,6 +4321,18 @@
...
+    {{BufferSource}} will match {{VideoFrame/format}} .
```
padenot, line 4328: "will be in the same format as the source", I think matching is fairly vague.

index.src.html
Djuffin: yeah, you're right we don't need this step

index.src.html
```diff
@@ -4324,14 +4322,14 @@
 : <dfn dict-member for=VideoFrameCopyToOptions>format</dfn>
 :: A {{VideoPixelFormat}} for the pixel data in the destination
     {{BufferSource}}. Potential values are: {{RGBA}}, {{RGBX}}, {{BGRA}},
-    {{BGRX}}. If unspecified or `null`, the the destination
-    {{BufferSource}} will match {{VideoFrame/format}} .
+    {{BGRX}}. If it does not [=map/exist=], the the destination
```
Kaiido, line 4325: nit: There is a double "the" here that you may want to remove?

index.src.html
Djuffin: Sure, thanks for pointing out that my approach with getContext/drawImage unconventional

Djuffin:
@padenot PTAL

padenot (approved):
All good, thanks!

Djuffin:
@aboba PTAL and merge if you don't have new comments.

PR #654, opened 2023-03-16
Extend EncodedVideoChunkMetadata for Spatial SVC

aboba:
Fixes https://github.com/w3c/webcodecs/issues/619

Rebase of PR https://github.com/w3c/webcodecs/pull/636

Work-in-progress do not merge

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/pull/654.html" title="Last updated on Apr 25, 2023, 9:12 PM UTC (75e0540)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/654/5c30e12...75e0540.html" title="Last updated on Apr 25, 2023, 9:12 PM UTC (75e0540)">Diff</a>

PR #756, opened 2024-01-04
Extend EncodedVideoChunkMetadata for Spatial Scalability

### LINKED ISSUE #619: Inconsistent SVC metadata between WebCodecs and WebRTC Encoded Transform API
Currently the [`EncodedChunkMetadata` dictionary](https://w3c.github.io/webcodecs/#encoded-video-chunk-metadata)  includes `SvcOutputMetadata`, which only supports temporal scalability ([temporalLayerId](https://w3c.github.io/webcodecs/#dom-svcoutputmetadata-temporallayerid)).  

However, within the Encoded-Transform API, [RTCEncodedFrameMetadata](https://www.w3.org/TR/webrtc-encoded-transform/#RTCEncodedVideoFrameMetadata) includes its own (incompatible) SVC metadata. 

Here is a comparison of the two approaches:
https://docs.google.com/presentation/d/1lFAUSvApbBYfBNJH_xcRW0YjD0aF5T1ZqjyyDJMesJw/edit#slide=id.g1a4ac56601a_3_0

Can we define the complete set of SVC metadata in WebCodecs and then have it inherited by Encoded Transform? 

Related: https://github.com/w3c/webrtc-encoded-transform/issues/170

aboba:
Fixes #619

Rebase and update of PR #654

Related: https://github.com/w3c/webrtc-encoded-transform/issues/220

***

### :boom: Error: 400 Bad Request :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Jan 9, 2024, 10:27 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [CSS Spec Preprocessor](https://api.csswg.org/bikeshed/) - CSS Spec Preprocessor is the web service used to build Bikeshed specs.

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2Fw3c%2Fwebcodecs%2F9efde67f1de0fdf0194f96bc8cb8f1eeb9197d80%2Findex.src.html&md-warning=not%20ready)

```
Error running preprocessor, returned code: 2.
FATAL ERROR: Couldn't find target frameId 'dict-member':
&lt;span data-dict-member-info="" for="EncodedVideoChunkMetadat/frameId">&lt;/span>
 ✘  Did not generate, due to errors exceeding the allowed error level.
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/webcodecs%23756.)._
</details>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -1713,13 +1713,13 @@ dictionary EncodedVideoChunkMetadata {
   VideoDecoderConfig decoderConfig;
   SvcOutputMetadata svc;
   BufferSource alphaSideData;
+  unsigned long long frameId;
+  sequence <unsigned long long> dependencies;
 };
 
 dictionary SvcOutputMetadata {
   unsigned long temporalLayerId;
   unsigned long spatialLayerId;
-  unsigned long long frameId;
-  sequence <unsigned long long> dependencies;
 };
 </xmp>
 
--- a/index.src.html
+++ b/index.src.html
@@ -1648,28 +1648,29 @@ Algorithms {#videoencoder-algorithms}
             2. Assign |outputConfig| to
                 {{VideoEncoder/[[active output config]]}}.
         7. If |encoderConfig|.{{VideoEncoderConfig/scalabilityMode}}
-            describes multiple [=temporal layers=]:
-            1. Let |svc| be a new {{SvcOutputMetadata}} instance.
-            2. Let |temporal_layer_id| be the zero-based index describing the
-                temporal layer for |output|.
-            3. Assign |temporal_layer_id| to
-                |svc|.{{SvcOutputMetadata/temporalLayerId}}.
-            4. Assign |svc| to
-                |chunkMetadata|.{{EncodedVideoChunkMetadata/svc}}.
+            describes multiple [=temporal layers=] or [=spatial layers=]:
+                  1. Let |svc| be a new {{SvcOutputMetadata}} instance.
         8. If |encoderConfig|.{{VideoEncoderConfig/scalabilityMode}}
+            describes multiple [=temporal layers=]:
+            1. Let |temporal_layer_id| be the zero-based index describing the
+                temporal layer for |output|.
+            2. Assign |temporal_layer_id| to
+                |svc|.{{SvcOutputMetadata/temporalLayerId}}.
+        9. If |encoderConfig|.{{VideoEncoderConfig/scalabilityMode}}
             describes multiple [=spatial layers=]:
-            1. Let |svc| be a new {{SvcOutputMetadata}} instance.
-            2. Let |spatial_layer_id| be the zero-based index describing the
+            1. Let |spatial_layer_id| be the zero-based index describing the
                 spatial layer for |output|.
-            3. Assign |spatial_layer_id| to
+            2. Assign |spatial_layer_id| to
                 |svc|.{{SvcOutputMetadata/spatialLayerId}}.
-            4. Assign |svc| to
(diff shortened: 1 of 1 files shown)


aboba:
@kalradivyanshu PTAL.

aboba:
For a frame to be decodable, all its dependencies need to have been received and decoded without an error callback. From the conference server perspective, this means not only tracking what frames were sent to each participant, but also the transport status (whether the frame was completely received) and whether it was successfully decoded. Currently the underlying encoder API limits avenues available for repair to keyframe generation, retransmission and forward error correction.  Alternate Long Term Reference (LTR) frames or layer refresh (LRR) are not yet supported.

index.src.html
```diff
@@ -1652,11 +1656,20 @@
                 |svc|.{{SvcOutputMetadata/temporalLayerId}}.
             4. Assign |svc| to
                 |chunkMetadata|.{{EncodedVideoChunkMetadata/svc}}.
-        8. If |encoderConfig|.{{VideoEncoderConfig/alpha}} is set to `"keep"`:
+        8. If |encoderConfig|.{{VideoEncoderConfig/scalabilityMode}}
+            describes multiple [=spatial layers=]:
+            1. Let |svc| be a new {{SvcOutputMetadata}} instance.
```
Djuffin, line 1661: This will override everything produced by the step 8 for SVC modes that have both temporal and spatial layers. for example: L2T2

index.src.html
```diff
@@ -1704,6 +1717,9 @@
 
 dictionary SvcOutputMetadata {
   unsigned long temporalLayerId;
+  unsigned long spatialLayerId;
+  unsigned long long frameId;
```
Djuffin, line 1721: `frameId` and `dependencies` are never set

aboba:
@tonyherre PTAL

index.src.html
tonyherre, line 1721: Could we have frameId and dependencies directly in EncodedVideoChunkMetadata? They're general fields, which make sense and are generally useful even in configs without multiple temporal/spatial layers.

aboba: Yes, we can do that.

index.src.html
aboba: I think I have fixed this. PTAL.

index.src.html
aboba: @Djuffin @tonyherre `dependencies` would be set to the sequence of `frameId` values that the `encodedChunk` depends on. But the bigger question is the behavior of `frameId`, which is discussed here: https://github.com/w3c/webrtc-encoded-transform/issues/220

aboba:
@kalradivyanshu  It has been noted that spatial scalability is not widely used today because it is not hardware accelerated and therefore create power and thermal issues on mobile devices. As a result, applications are using spatial simulcast instead. Also, the current WebCodecs API does not support layer refresh, which means that if a spatial frame is lost, a base layer keyframe is required, rather than just creating a new spatial frame referencing a received base-layer frame (e.g. moving to a new Long-Term Reference). 

@Djuffin has argued that these problems need to be fixed before spatial scalability could become popular in WebCodecs, and therefore that it would make sense to focus on a new encoder API that can address the problems rather than just shipping a (potentially unusable) feature.

PR #758, opened 2024-01-09
Add contentHint section to codec-registries

tguilbert-google:
This PR addresses #735

Notably, it:
- Updates the codec registry entry requirements to include a section on `contentHint guidelines`.
- Adds `AudioEncoderConfig.contentHint` and updates `VideoEncoderConfig.contentHint` to match.
- Adds a `contentHint` section to the Opus registry.

***
<a href="https://pr-preview.s3.amazonaws.com/tguilbert-google/webcodecs/pull/758.html" title="Last updated on Jan 9, 2024, 1:35 AM UTC (95c678e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/758/21e4a15...tguilbert-google:95c678e.html" title="Last updated on Jan 9, 2024, 1:35 AM UTC (95c678e)">Diff</a>

PR #759, opened 2024-01-10
Clarify VideoEncoderConfig.contentHint

### LINKED ISSUE #735: `contentHint`: notes aren't normative and shouldn't describe normative behaviour
I just found a tab with my review of https://github.com/w3c/webcodecs/pull/729 unsubmitted, so I'm opening this new issue with my comments:

See [`contentHint`](https://w3c.github.io/webcodecs/#dom-videoencoderconfig-contenthint):

> NOTE: Any codec-specific encoding options take precedence over [contentHint](https://w3c.github.io/webcodecs/#dom-videoencoderconfig-contenthint).

This seems to be a requirement, but is in an informative note, and that's not correct. We want to describe in what case and which codec-specific option overrides this `contentHint`. Also, we might want this for audio, e.g. we could make it map to those Opus flags: https://github.com/xiph/opus/blob/master/include/opus_defines.h#L191-L199

tguilbert-google:
Fixes #735, addresses comments from #758.

This PR removes the note from `VideoEncoderConfig.contentHint` and forces User Agent to respect other explicitly set encoder options.

It also adds an equivalent `AudioEncoderConfig.contentHint` section.

***
<a href="https://pr-preview.s3.amazonaws.com/tguilbert-google/webcodecs/pull/759.html" title="Last updated on Feb 5, 2024, 10:31 PM UTC (0795078)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/759/21e4a15...tguilbert-google:0795078.html" title="Last updated on Feb 5, 2024, 10:31 PM UTC (0795078)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -52,6 +52,7 @@ spec: webrtc-svc; urlPrefix: https://www.w3.org/TR/webrtc-svc/
 
 spec: mst-content-hint; urlPrefix: https://www.w3.org/TR/mst-content-hint/
     type: dfn; text: video content hints; url:#video-content-hints
+    type: dfn; text: audio content hints; url:#audio-content-hints
 
 spec: ECMASCRIPT; urlPrefix: https://tc39.es/ecma262/
     type: dfn; text: the current Realm; url: #current-realm
@@ -2051,6 +2052,7 @@ dictionary AudioEncoderConfig {
   [EnforceRange] unsigned long numberOfChannels;
   [EnforceRange] unsigned long long bitrate;
   BitrateMode bitrateMode = "variable";
+  DOMString contentHint;
 };
 </xmp>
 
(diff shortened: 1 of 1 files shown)


aboba:
The current [Opus configuration options](https://www.w3.org/TR/webcodecs-opus-codec-registration/#opus-encoder-config) are missing [configuration parameters ](https://draft.ortc.org/#opus-codec-settings*)such as `[signal`](https://draft.ortc.org/#dfn-signal) (with settings "auto", "music" and "voice") and `[application`](https://draft.ortc.org/#dfn-application) (with settings "voip", "audio" and "lowdelay"). IMHO, it might make more sense to add these to the [Opus registry,](https://www.w3.org/TR/webcodecs-opus-codec-registration) or at least to specify how 
these settings map to the Audio Content Hint values of "", "speech", "speech-recognition", and "music".  For example, you could indicate that "" corresponds to the `signal` value of "auto", "speech" corresponds to "voice" and "music" corresponds to "music", with "speech-recognition" having no effect. The "speech" content hint might map to the `application` value of "voip",  the "music" hint could map to "audio" and "speech-recognition" could map to "lowdelay".

tguilbert-google:
That's a good point. Those configuration parameters are defined in a spec, and can be referred to from the codec registry, unlike the implementation specific headers (e.g. `OPUS_APPLICATION_VOIP`). I should have attempted to use these in #758.

Under the current proposed definition of `AudioEncoderConfig.contentHint`, adding these could be done in parallel, without conflict: this PR's definition of `contentHint` specifies to respect any explicitly set flags, and then make a best effort to configure additional encoder flags; the `application`/`signal` flags in the Opus registry would take precedence over `contentHint`.

Would we want to keep `AudioEncoderConfig.contentHint` for symmetry with the video half of things, or is it useful for other audio codecs with fewer configuration parameters? Or would `contentHint` be a "quick and easy" top level alias for settings that are prescriptively mapped, and no longer just a "best effort"?

An explicit mapping for Opus seems easy with the proposed flags, but that might not be as easy for other audio/video codecs... Would defining an explicit mapping for Opus force all other codecs into defining mappings? I think this is something @Djuffin wasn't too keen on doing (but that was maybe only for implementation specific mappings `OPUS_APPLICATION_VOIP`).

aboba:
For [`alaw`](https://www.w3.org/TR/webcodecs-alaw-codec-registration/), [`ulaw`]([url](https://www.w3.org/TR/webcodecs-ulaw-codec-registration/)) and[ `pcm`](https://www.w3.org/TR/webcodecs-pcm-codec-registration/), it's not obvious to me what can be done to implement audio content hints like "speech", "music" or "speech-recognition". 

Does anything come to mind for[ `flac`](https://www.w3.org/TR/webcodecs-flac-codec-registration/) or [`aac`](https://www.w3.org/TR/webcodecs-aac-codec-registration/)?

Djuffin (approved):

padenot:
> Does anything come to mind for[ `flac`](https://www.w3.org/TR/webcodecs-flac-codec-registration/) or [`aac`](https://www.w3.org/TR/webcodecs-aac-codec-registration/)?

Not that I know of.

padenot (approved):

tguilbert-google:
Following discussions with @Djuffin, which summarized @aboba's point of view on this PR, I updated the PR to only modify `VideoEncoderConfig.contentHint`.

In essence, a `contentHint` can be applied to multiple different codecs for video, but only Opus seems to benefit from a `contentHint` on the audio side.

Instead of adding `contentHint` to audio, I will send a follow-up PR which adds the `application`/`signal` flags directly to `OpusEncoderConfig`.

PR #764, opened 2024-01-31
{Audio,Video}DecoderConfig with detached description are invalid

### LINKED ISSUE #762: Clarify the `Clone configuration` algorithm
When looking a `Clone Configuration`, and after having found a bug in Gecko, [a DOM developer said](https://bugzilla.mozilla.org/show_bug.cgi?id=1874293#c6):

> This looks very underspecified in the spec. For one, I don't think any spec defines what assign a copy of config[m] to clone[m]
> means. And either that can throw exceptions, and it should be specified what needs to happen, or it doesn't and we need to 
> figure out what to do in our implementation.

It can throw an exception, a minimal test case (adapted from our bug) would be:

```js
var buf = new ArrayBuffer(4);
buf.transferToFixedLength()
VideoDecoder.isConfigSupported({"codec": "፝a", "description":buf})
```

Chromium says: 

```
Uncaught (in promise) TypeError: Failed to execute 'isConfigSupported' on 'VideoDecoder': description is detached.
```

which makes sense to me, but we need to spec it.

padenot:
This is what Chromium (at least) does. This means isConfigSupported and configure will now throw if description has been detached.

This also fixes a bug in the ImageDecoder version of this.

This fixes #762.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/pull/764.html" title="Last updated on Jan 31, 2024, 10:26 AM UTC (250f7d8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/764/9e0f22e...250f7d8.html" title="Last updated on Jan 31, 2024, 10:26 AM UTC (250f7d8)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -1916,7 +1916,9 @@ To check if an {{AudioDecoderConfig}} is a <dfn export>valid AudioDecoderConfig<
 1. If {{AudioDecoderConfig/codec}} is empty after
     [=strip leading and trailing ASCII whitespace|stripping leading and trailing ASCII whitespace=],
     return `false`.
-2. Return `true`.
+2. If the result of running  IsDetachedBuffer (described in
+    [[!ECMASCRIPT]]) on {{AudioDecoderConfig/description}} is `true`, return `false`.
+3. Return `true`.
 
 <dl>
   <dt><dfn dict-member for=AudioDecoderConfig>codec</dfn></dt>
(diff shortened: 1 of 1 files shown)


sandersdan (approved):

Djuffin (approved):

index.src.html
```diff
@@ -1916,7 +1916,9 @@
 1. If {{AudioDecoderConfig/codec}} is empty after
     [=strip leading and trailing ASCII whitespace|stripping leading and trailing ASCII whitespace=],
     return `false`.
-2. Return `true`.
+2. If the result of running  IsDetachedBuffer (described in
+    [[!ECMASCRIPT]]) on {{AudioDecoderConfig/description}} is `true`, return `false`.
```
petervanderbeken, line 1920: Given that `description` is a WebIDL value this should probably use https://webidl.spec.whatwg.org/#buffersource-detached instead.

padenot: Thanks Peter, fixed in https://github.com/w3c/webcodecs/commit/c97f2989c9bd3e9bbca9b5a0f96a3ef029e7d7a0.

PR #766, opened 2024-02-15
Add HBD YUV Planar pixel formats

sandersdan:
This change adds the planar YUV formats described in issue #384, most of which are 10 or 12-bit. (P010/P012/P016/RGBAF16 are not included.)

***
<a href="https://pr-preview.s3.amazonaws.com/sandersdan/webcodecs/pull/766.html" title="Last updated on Feb 20, 2024, 8:53 PM UTC (b7c9e46)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/766/e979025...sandersdan:b7c9e46.html" title="Last updated on Feb 20, 2024, 8:53 PM UTC (b7c9e46)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -4315,21 +4315,37 @@ number and order of the planes. Each format is described in its own sub-section.
 enum VideoPixelFormat {
   // 4:2:0 Y, U, V
   "I420",
+  "I420P10",
+  "I420P12",
   // 4:2:0 Y, U, V, A
   "I420A",
+  "I420AP10",
+  "I420AP12",
   // 4:2:2 Y, U, V
   "I422",
+  "I422P10",
+  "I422P12",
+  // 4:2:2 Y, U, V, A
+  "I422A",
+  "I422AP10",
+  "I422AP12",
   // 4:4:4 Y, U, V
   "I444",
+  "I444P10",
+  "I444P12",
+  // 4:4:4 Y, U, V, A
+  "I444A",
+  "I444AP10",
+  "I444AP12",
   // 4:2:0 Y, UV
   "NV12",
-  // 32bpp RGBA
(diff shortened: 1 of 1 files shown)


index.src.html
```diff
@@ -4433,27 +4553,236 @@
...
+  <dt><dfn enum-value for=VideoPixelFormat>I422P10</dfn></dt>
   <dd>
+    This format is composed of three distinct planes, one plane of Luma and two
+    planes of Chroma, denoted Y, U and V, and present in this order.
+
+    The U an V planes are [=sub-sampled=] horizontally by a [=factor=] of 2
+    compared to the Y plane, and not [=sub-sampled=] vertically.
+
+    Each sample in this format is 10 bits, encoded as a 16-bit integer in
+    little-endian byte order.
```
padenot, line 4569: I wonder if we should say something about those 6 bits (or 4, depending on the depth) that aren't used.

Are they to be ignored, are they always set to 0? In terms of developer ergonomics, if I read back the values to an `ArrayBuffer` to perform some computations, should I always mask the unused bits or can I simply read the value with a `DataView` as uint16_t.  In terms of implementation, is that a fair assumption we can make, or do some decoders leave the bits alone, and so this could be a source of fingerprinting?

We should also specify (probably globally, since it applies to all formats) that the sample values are unsigned integer, I tried to look for that but didn't find it for the video side of things.

padenot (approved):
Generally looks perfect, but I have some questions, that we might be able to address in a follow-up.

Djuffin (approved):

index.src.html
aboba: Looks good to me too.

aboba (approved):

PR #777, opened 2024-03-05
Add OpusEncoderConfig.[signal/application]

tguilbert-google:
Follow up work for #759.

`AudioEncoderConfig.contenHint` was not added, since only the Opus codec would benefit from it. Instead, we favored adding the individual codec configuration parameters directly to the Opus registry.

This PR adds `OpusEncoderConfig.signal` and `OpusEncoderConfig.application`.

### DIFF
--- a/opus_codec_registration.src.html
+++ b/opus_codec_registration.src.html
@@ -226,7 +226,7 @@ enum OpusSignal {
 </pre>
 
 The {{OpusSignal}} describes the type of audio signal being encoded. See
-[secion 9.3.1.1](https://draft.ortc.org/#opus-codec-options*) of [[OBJECT-RTC]].
+[section 9.3.1.1](https://draft.ortc.org/#opus-codec-options*) of [[OBJECT-RTC]].
 
 <dl>
   <dt><dfn enum-value for=OpusSignal>auto</dfn></dt>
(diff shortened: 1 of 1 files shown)


opus_codec_registration.src.html
```diff
@@ -197,6 +213,67 @@
...
+enum OpusSignal {
+  "auto",
+  "music",
+  "voice",
+};
+</xmp>
+</pre>
+
+The {{OpusSignal}} describes the type of audio signal being encoded. See
+[secion 9.3.1.1](https://draft.ortc.org/#opus-codec-options*) of [[OBJECT-RTC]].
```
tguilbert-google, line 229: Again, I'm not sure if referencing that section is allowed and adds much. The `signal` and `application` parameters are not defined with a lot of details there.

opus_codec_registration.src.html
```diff
@@ -115,6 +121,8 @@
 <xmp>
 dictionary OpusEncoderConfig {
   OpusBitstreamFormat format = "opus";
+  OpusSignal signal = "auto";
+  OpusApplication application = "audio";
```
tguilbert-google, line 125: There is no default in libopus for `OPUS_SET_APPLICATION` (unlike the default of `"auto"` for `OPUS_SET_SIGNAL`).  I think `"audio"` is the most generic/least surprising as a default.

opus_codec_registration.src.html
aboba: I would keep the reference non-normative. Or you can just copy the text.

padenot (approved):

Issue #607, opened 2022-11-15
Announcement: Human face metadata entry to WebCodecs VideoFrame Metadata Registry


ttoivone:
As per requirement of the [VideoFrame Metadata Registry](https://w3c.github.io/webcodecs/video_frame_metadata_registry.html), this issue has been filed to allow discussion about the face metadata on being added into [VideoFrameMetadata](https://www.w3.org/TR/webcodecs/#dictdef-videoframemetadata) dictionary. The proposal is in the [Media capture extensions PR #78](https://github.com/w3c/mediacapture-extensions/pull/78) and the [face detection explainer](https://github.com/w3c/mediacapture-extensions/blob/main/face-detection-explainer.md) shall be updated.

To prevent discussions in multiple places, I do suggest that the discussion about the metadata is actually done in Media Capture Extensions [Media capture extensions PR #78](https://github.com/w3c/mediacapture-extensions/pull/78) where the explainer and spec will be updated, instead of this issue.

aboba:
FYI, the metadata proposal has been modified to make it more generic so as to make it more useful for segmentation.

This in turn raises the question as to how a WebCodecs Encoder could take advantage of the segmentation information (e.g. per-macroblock QP).

chrisn:
One concern I have is that this adds face detection specific attributes under a top level `segments` entry, which is a rather generic name. It's still specific because of the [`SegmentType` enum](https://w3c.github.io/mediacapture-extensions/#dom-segmenttype).

So I'd prefer to either rename `segments` to something indicating face detection as the source of the metadata, or relax the requirements to make `SegmentType` optional.

Aside from that, I'd be happy to review a PR that adds to the registry.

youennf:
Discussed at TPAC:
- segments naming or SegmentType extension point should be discussed in WebRTC WG.
- above issue is not a blocker for the registry PR.

Djuffin:
I've just noticed that this is already a part of [Media Capture and Streams Extensions editor draft](https://w3c.github.io/mediacapture-extensions/#human-face-segmentation)

https://github.com/w3c/mediacapture-extensions/pull/78

aboba:
[...]

PR #787, opened 2024-05-15
Add segments metadata entry that refers to https://w3c.github.io/mediacapture-extensions/#human-face-segmentation

youennf:
Fixes https://github.com/w3c/webcodecs/issues/607


youennf:
Can we merge this PR?

chrisn:
We'll also need to publish the registry to /TR. I'll post a CfC..

chrisn:
The CfC concluded (see [mailing list](https://lists.w3.org/Archives/Public/public-media-wg/2024Jun/0009.html)), so we can merge this and then publish the registry.

PR #802, opened 2024-06-10
Mark ImageDecoderInit.data as AllowSharedBufferSource.

### LINKED ISSUE #799: `ImageBufferSource` should be `[AllowShared]`
This is what [Chromium implements](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/webcodecs/image_decoder_init.idl;l=7), but isn't in the spec. We're implementing like Chromium in Firefox, but we need to make the spec say that.

@dalecurtis, anything to add based on implementation experience? I see that it can either transfer or copy, so we should be good for TOCTOU issues and related common problems.

dalecurtis:
This is what Chromium implements, but we forgot to update the spec when other data fields were changed to allow shared buffers.

Fixes #799

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/pull/802.html" title="Last updated on Jun 10, 2024, 5:00 PM UTC (90ab77c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/802/9eaaad4...90ab77c.html" title="Last updated on Jun 10, 2024, 5:00 PM UTC (90ab77c)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -5747,7 +5747,7 @@ ImageDecoderInit Interface {#imagedecoderinit-interface}
 --------------------------------------------------------
 <pre class='idl'>
 <xmp>
-typedef (BufferSource or ReadableStream) ImageBufferSource;
+typedef (AllowSharedBufferSource or ReadableStream) ImageBufferSource;
 dictionary ImageDecoderInit {
   required DOMString type;
   required ImageBufferSource data;


dalecurtis:
ISTR the reason we expand the typedef in Chrome is just due to an IDL compiler limitation, so I left this as the unexpanded version. Let me know if you'd prefer we use the expanded version.

Djuffin (approved):

Issue #599, opened 2022-10-26
Relationship between VideoFrameMetadata and VideoFrameCallbackMetadata


youennf:
https://wicg.github.io/video-rvfc/#dictdef-videoframecallbackmetadata defines some metadata that might be useful to be in VideoFrameMetadata.
It might also be useful to expose future VideoFrameMetadata fields in VideoFrameCallbackMetadata.
The following could be done:
1. VideoFrameCallbackMetadata derives from VideoFrameMetadata
2. Move some metadata from VideoFrameCallbackMetadata to VideoFrameMetadata: captureTime, receiveTime and rtpTimestamp.

@tguilbert-google, @dalecurtis, thoughts?

dalecurtis:
It's a good question. Thanks for asking @youennf.

I think they're a bit unwieldy for direct derivation in either direction. E.g., Presentation time metadata and the fact that `VideoFrameCallbackMetadata` has a `mediaTime` field, but with `VideoFrame` we expect that to be the `timestamp` field. 

I prefer just cross-pollinating the fields we think are important between the two without worrying too much about perfect sync. I.e., I agree the ones you list in 2 seem reasonable to add to `VideoFrameMetadata`.

youennf:
The main question is whether there will be any field in VideoFrameMetadata we would not want to expose in VideoFrameCallbackMetadata.
AIUI, we will always want to expose VideoFrameMetadata fields in VideoFrameCallbackMetadata.
It does not seem great to have to update VideoFrameCallbackMetadata spec and implementation every time we add a VideoFrameMetadata field in the registry.

Hence why deriving VideoFrameCallbackMetadata from VideoFrameMetadata is nice (one issue is name clashing though).
Composition is good too, I guess we could add a `VideoFrameMetadata other` to VideoFrameCallbackMetadata for instance. But this would duplicate some existing fields we might want to upstream to VideoFrameMetadata, as @tguilbert-google pointed out.

I'll file an issue about adding captureTime, receiveTime and rtpTimestamp to VideoFrameMetadata.

dalecurtis:
I think that's actually a bit neat in reverse. So I'm in favor of `VideoFrameCallbackMetadata` inheriting from `VideoFrameMetadata` assuming we can get all those fields added to the registry. Our requirements for registry entry are much stricter than what VideoFrameCallback defined though, so that may be a bit tough.

youennf:
[...]

Issue #601, opened 2022-10-27
Expose in VideoFrameMetadata some fields from VideoFrameCallbackMetadata


youennf:
It seems useful to expose some fields in VideoFrameMetadata like:
- https://wicg.github.io/video-rvfc/#dom-videoframecallbackmetadata-capturetime
- https://wicg.github.io/video-rvfc/#dom-videoframecallbackmetadata-receivetime
- https://wicg.github.io/video-rvfc/#dom-videoframecallbackmetadata-rtptimestamp

This could be added as an entry in VideoFrameMetadata registry.

dalecurtis:
These sound fine to me, but @aboba points out that https://wicg.github.io/video-rvfc/ may not meet the registry requirements yet.

I know at TPAC there was some talk of moving rVFC into the HTML spec. Do we still want to do that?

@chrisn @tguilbert-google

youennf:
I think the idea would be to write a registry entry that would define these values without referencing rvfc. Instead, it would reference webrtc-pc and/or media capture-main.

tguilbert-google:
Yes, I was still planning on moving rVFC to the HTML spec. I haven't found the time to do so yet. Embedding the spec into the HTML spec is more work than just referencing it.

Previous (closed) PR tracking this work:
https://github.com/whatwg/html/pull/5332#issuecomment-1251535284

chrisn:
I agree, adding this into HTML makes sense.

aboba:
The RVFC metadata items seem quite useful to me, but they also raise some questions for the behavior of other APIs: 

1. In the VideoFrame(s) produced by MediaStreamTrackProcessor, should we expect `capturetime` metadata to be present? 
2. Is there an expectation that the WebCodecs encoder will carry Videoframe metadata items through to encodedChunk metadata? 

What I'd really like to be able to do is to trace performance through each stage of the receive and send pipeline.

youennf:
> 1. In the VideoFrame(s) produced by `MediaStreamTrackProcessor`, should we expect `capturetime` metadata to be present?

It depends on MediaStreamTrack's source. If source is getUserMedia, yes.

> 2\. Is there an expectation that the WebCodecs encoder will carry `Videoframe` metadata items through to encodedChunk metadata?  (e.g. `capturetime`)

EncodedVideoChunk has no metadata so I would think that it would not be carried automatically.
[...]

PR #813, opened 2024-07-05
Add captureTime, receiveTime and rtpTimestamp to VideoFrameMetadata

guidou:
These fields are useful for WebRTC-based applications.
See issue #601

### DIFF
--- a/video_frame_metadata_registry.src.html
+++ b/video_frame_metadata_registry.src.html
@@ -61,6 +61,18 @@ VideoFrameMetadata members {#videoframemetadata-members}
     <td>segments</td>
     <td>[Human face segmentation](https://w3c.github.io/mediacapture-extensions/#human-face-segmentation)</td>
   </tr>
+  <tr>
+    <td>captureTime</td>
+    <td>[Capture time](https://wicg.github.io/video-rvfc/#dom-videoframecallbackmetadata-capturetime)</td>
+  </tr>
+  <tr>
+    <td>receiveTime</td>
+    <td>[Receive time](https://wicg.github.io/video-rvfc/#dom-videoframecallbackmetadata-receivetime)</td>
+  </tr>
+  <tr>
+    <td>rtpTimestamp</td>
+    <td>[RTP timestamp](https://wicg.github.io/video-rvfc/#dom-videoframecallbackmetadata-rtptimestamp)</td>
+  </tr>
 </table>
 
 
--- a/video_frame_metadata_registry.src.html
+++ b/video_frame_metadata_registry.src.html
@@ -63,15 +63,15 @@ VideoFrameMetadata members {#videoframemetadata-members}
   </tr>
   <tr>
     <td>captureTime</td>
-    <td>[Capture time](https://wicg.github.io/video-rvfc/#dom-videoframecallbackmetadata-capturetime)</td>
+    <td>[Capture time](https://w3c.github.io/mediacapture-extensions/#dfn-capture-timestamp)</td>
   </tr>
   <tr>
     <td>receiveTime</td>
-    <td>[Receive time](https://wicg.github.io/video-rvfc/#dom-videoframecallbackmetadata-receivetime)</td>
+    <td>[Receive time](https://w3c.github.io/mediacapture-extensions/#dfn-receive-timestamp)</td>
   </tr>
   <tr>
     <td>rtpTimestamp</td>
-    <td>[RTP timestamp](https://wicg.github.io/video-rvfc/#dom-videoframecallbackmetadata-rtptimestamp)</td>
+    <td>[RTP timestamp](https://w3c.github.io/mediacapture-extensions/#dfn-rtp-timestamp)</td>
   </tr>
 </table>
 
--- a/video_frame_metadata_registry.src.html
+++ b/video_frame_metadata_registry.src.html
@@ -63,15 +63,15 @@ VideoFrameMetadata members {#videoframemetadata-members}
   </tr>
   <tr>
     <td>captureTime</td>
-    <td>[Capture time](https://w3c.github.io/mediacapture-extensions/#dfn-capture-timestamp)</td>
+    <td>[Capture time](https://w3c.github.io/mediacapture-extensions/#dom-videoframemetadata-capturetime)</td>
   </tr>
   <tr>
     <td>receiveTime</td>
-    <td>[Receive time](https://w3c.github.io/mediacapture-extensions/#dfn-receive-timestamp)</td>
+    <td>[Receive time](https://w3c.github.io/mediacapture-extensions/#dom-videoframemetadata-receivetime)</td>
   </tr>
   <tr>
     <td>rtpTimestamp</td>
-    <td>[RTP timestamp](https://w3c.github.io/mediacapture-extensions/#dfn-rtp-timestamp)</td>
+    <td>[RTP timestamp](https://w3c.github.io/mediacapture-extensions/#dom-videoframemetadata-rtptimestamp)</td>
   </tr>
 </table>
 


guidou:
cc @Djuffin, @padenot , @youennf, @aboba

aboba:
Does this PR imply any behavior changes in WebCodecs API?  

For example, on encoding is there an expectation that `VideoFrame.captureTime` is copied to `EncodedVideoChunk.captureTime`.  Or on decoding is `EncodedVideoChunk.receiveTime` or `EncodedVideoChunk.rtpMetadata` to be copied to `VideoFrame.receiveTime` or `VideoFrame.rtpMetadata`? 

If there are no changes in behavior (e.g. if the attributes don't affect the encode or decode process or some other aspect of WebCodecs) then the attributes could be defined in another specification where behavior is affected (e.g. mediacapture-transform?), and added to the [VideoFrame Metadata Registry](https://github.com/w3c/webcodecs/blob/main/video_frame_metadata_registry.src.html).

guidou:
This PR as currently written does not imply any behavior in the WebCodecs API, although I would expect the things you mentioned (e.g., forwarding them to/from EncodedVideoChunk) as potentially useful.

The idea for this PR is to provide information to applications so that they can do similar things to what they can do with `requestVideoFrameCallback` (e.g., better A/V sync and delay measurements). This doesn't require any other behavior changes in WebCodecs (at least for applications using mediacapture-transform  + WebRTC).

guidou:
I think we can specify forwarding to EncodedVideoChunk in  a separate PR since this one has value on its own without  specifying further changes to WebCodecs.

Djuffin:
I used to be skeptical about these timestamps since they are not passed through the encoding-decoding cycle, but since we already have entries in VideoFrame Metadata Registry that don't do that, I think it's okay now.

And RTC software like Teams, Mean and Facetime can really use it for A/V sync and latency estimation, even if they have to pass this information via separate channels.

Djuffin (approved):

aboba:
There is still the question of where the metadata orginates. For example, are we saying that:

1.  The `MediaStreamTrackProcessor` method will provide `VideoFrame.captureTimestamp` if the MST is obtained from a local capture? 
2. The `MediaStreamTrackProcessor` method will provide `VideoFrame.receiveTimestamp` and `VideoFrame.rtpMetadata` if the MST is obtained remotely via WebRTC-PC?

Djuffin:
I thought for all metadata entries the answer to these questions is [MAY](https://datatracker.ietf.org/doc/html/rfc2119#section-5).

aboba:
@Djuffin MAY might be ok for these metadata fields.  However, is `VideoFrame.timestamp` and `EncodedVideoChunk.timestamp` optional for WebCodecs implementations?

youennf:
> I thought for all metadata entries the answer to these questions is [MAY](https://datatracker.ietf.org/doc/html/rfc2119#section-5).

I agree from a WebCodecs POV.
But it is not sufficient from an interop point of view.
Probably each spec defining a MST video source should describe which metadata it generates, just like each spec defines which constraints are supported by a given source.
Putting the definition at the source ensures the same metadata is exposed via MSTP or via VideoFrame constructor (from a video element).

That would mean mediacapture-main and webrtc-pc here.
As of mediacapture-transform VideoTrackGenerator, nothing seems needed though we could add a note stating that metadata are preserved.

guidou:
FWIW, the `requestVideoFrameCallback` spec where these fields are originally defined say that `captureTime` applies to local cameras and remote frames (WebRTC), `receiveTime` to WebRTC frames, and `rtpTimestamp`  to WebRTC frames. But I agree with @youennf that having each MST source spec indicate the metadata it generates is the best way to organize that. 

In any case, we need to have entries for these fields in the VideoFrameMetadata registry.

Djuffin:
> @Djuffin MAY might be ok for these metadata fields.  However, is `VideoFrame.timestamp` and `EncodedVideoChunk.timestamp` optional for WebCodecs implementations? 

They're mandatory. 

Is there some kind of deep connection here that I miss?

aboba (changes requested):
Changes to `VideoFrame` registry are ok, but should probably not reference RVFC. receiveTime and rtpMetadata could be defined in WebRTC-PC or WebRTC-Extensions and captureTime could be defined in Media Capture & Streams or Media Capture Extensions.

chrisn:
[Minutes](https://www.w3.org/2024/07/09-mediawg-minutes.html#t07) from 9 July 2024 Media WG meeting. @aboba summarised the conclusion in https://github.com/w3c/webcodecs/pull/813#pullrequestreview-2169568794.

Djuffin:
Summary of WG discussion:
[HTMLVideoElement.requestVideoFrameCallback](https://wicg.github.io/video-rvfc) is not the best spec to reference here, because it doesn't describe how and when these timestamps are set. 
Corresponding changes need to be made in [MediaStreamTrackProcessor](https://www.w3.org/TR/mediacapture-transform/) and [Media Capture and Streams](https://www.w3.org/TR/mediacapture-streams/) specs. Something along the lines: "MediaStreamTrackProcessor sets capture timestamps for VideoFrames coming from camera..."

Later this PR should reference these specs.

vinniespina (approved):

aboba:
"And RTC software like Teams, Mean and Facetime can really use it for A/V sync and latency estimation, even if they have to pass this information via separate channels."

[BA] To do A/V sync, `captureTime` and `receiveTime` need to be provided for both audio and video.  

Also, if they are to be usable for non-RTP transports, they need to be defined in a way that is independent of RTP/RTCP. For example, on the local peer, `captureTime` represents the capture time of the first byte according to the local wallclock. On a remote peer, `captureTime` is set by the receiver.  For example, the local peer's `captureTime` can be serialized on the wire and then set on the receiver (e.g. not adjusted to the receiver wallclock). `receiveTime` is set on the receiver, based on the receiver's wallclock. (`receiveTime` - `captureTime`) can then be used to estimate the sender/receiver offset as well as Jitter.

video_frame_metadata_registry.src.html
```diff
@@ -61,6 +61,18 @@
     <td>segments</td>
     <td>[Human face segmentation](https://w3c.github.io/mediacapture-extensions/#human-face-segmentation)</td>
   </tr>
+  <tr>
+    <td>captureTime</td>
+    <td>[Capture time](https://wicg.github.io/video-rvfc/#dom-videoframecallbackmetadata-capturetime)</td>
```
aboba, line 66: The RVFC text is too RTP centric. For a WebCodecs application, captureTime can be serialized on the wire and set by a WebCodecs application for frames received from a remote peer. Also, NTP timestamp format does not imply a global clock so change “synchronization” to “alignment”.

video_frame_metadata_registry.src.html
```diff
@@ -61,6 +61,18 @@
...
+  </tr>
+  <tr>
+    <td>receiveTime</td>
+    <td>[Receive time](https://wicg.github.io/video-rvfc/#dom-videoframecallbackmetadata-receivetime)</td>
```
aboba, line 70: Comment: can we allow this to be present in a WebCodecs application as well (set by the receiver)?

aboba, line 70: Comment: can we allow this to be present in a WebCodecs application as well (set by the receiver)?

video_frame_metadata_registry.src.html
```diff
@@ -61,6 +61,18 @@
...
+  </tr>
+  <tr>
+    <td>rtpTimestamp</td>
+    <td>[RTP timestamp](https://wicg.github.io/video-rvfc/#dom-videoframecallbackmetadata-rtptimestamp)</td>
```
aboba, line 74: The RVFC text is ok here but I'd still probaby copy it over rather than referencing it.

video_frame_metadata_registry.src.html
handellm: The plan is to describe this as "set for remote webrtc tracks" and use DOMHighResTimestamp relative to local Performance.timeOrigin. I agree it should be settable by a WebCodecs application.

video_frame_metadata_registry.src.html
handellm: I planned to define what a "rtp timestamp" is over at webrtc-extensions and simply define here that it's present for "remote webrtc tracks" together with a reference.

guidou:
This PR has been updated to reference mediacapture-extensions where these concepts are now properly defined (similar to human face segmentation).

video_frame_metadata_registry.src.html
guidou: The PR is now updated to reference mediacapture-extensions, where these concepts are now defined.

aboba (approved):

aboba:
@guidou Does this PR also resolve #599 ?

PR #825, opened 2024-08-19
Bind auto-publish jobs to actual spec updates

tidoust:
Each time an update was made to any of the specs in the repository, all 16 specs got re-published because the auto-publish workflow did not know what was changed. This took time for no good reason.

The [`paths` filter](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore) allows to run a workflow conditionally. Unfortunately, it only applies at the workflow level (and cannot be set at the job level within a workflow in particular). This means the only direct way to avoid running a job is to create a workflow per spec.

This adds an auto-publish workflow template and a companion script to generate auto-publish workflows for each and every spec in the repository, that only run when needed, meaning only when the spec source did change. The script only needs to run once in a while when new specs get added or when the workflows need to change.

The new auto-publish workflows created by the script replace the previous `auto-publish.yml` workflow.

Note: The `auto-publish.yml` workflow still had a `BUILD_FAIL_ON: nothing` property, which was meant to be a temporary measure and should no longer be needed (see https://github.com/w3c/webcodecs/commit/adfb63c591152cb2ccfd89f74319a73198576638). This update switches back to `BUILD_FAIL_ON: warning`.

This is the same approach as that done for w3c/encrypted-media: https://github.com/w3c/encrypted-media/commit/cefbd73c13c8b69ad8e01b7dd75aea3788a7dc8b


Djuffin (approved):

tidoust:
Merging. The setting seems to work for w3c/encrypted-media in particular. I'm around to fix issues if that ends up not working as expected.

Issue #669, opened 2023-04-27
Signaling codec resource exhaustion (formerly Custom error types)


Djuffin:
Currently WebCodecs rely on DOMExceptions with different names, but DOMException names are [deprecated](https://webidl.spec.whatwg.org/#idl-DOMException-error-names) . 
For proper and granular error reporting WebCodecs might need a special error type, like WebGpu [pipeline error](https://www.w3.org/TR/webgpu/#gpupipelineerror).

We might also need a special error type for rejecting promises and reporting errors via VideoEncoder and VideoDecoder callbacks.

chrisn:
Looking at [the list of DOMException names](https://webidl.spec.whatwg.org/#idl-DOMException-error-names) it's only `IndexSizeError`, `InvalidAccessError,` `TypeMismatchError`, and `URLMismatchError` that are marked as deprecated. The WebCodecs spec doesn't use any of those so I think we're OK in that regard.

aboba:
Can we use `EncodingError` to indicate an issue with the data provided to the Encoder/Decoder, and something else (e.g. `OperationError`) to indicate a resource issue (e.g. hardware encoder resource no longer available),

xingri:
@Djuffin, @aboba  If it is possible, can I get a reference on the current working in progress? 
Would like to investigate any possible changes on application side.

chrisn:
Discussed in Media WG meeting, 30 May 2023, [minutes](https://www.w3.org/2023/05/30-mediawg-minutes.html#t03).

padenot:
None.

Djuffin:
I'm looking at how to suffice signal that we ran out of HW encoders or decoders.
Most logical behavior seems simply reporting [QuotaExceededError](https://webidl.spec.whatwg.org/#quotaexceedederror)
and then I realized that it's basically [Resource Reclamation](https://www.w3.org/TR/webcodecs/#resource-reclamation), that we already have in the spec.

So we don't even need to change much. But we'll need to adjust the definition of the active encoder/decoder i.e. declare newly created encoder/decoder that hasn't outputted anything yet - inactive. Then UA will be able to *reclame* it immediately after it realized that there is no system resources for it.

aboba:
@Djuffin Reading Section 11: Resource Reclamation section, I had some a question:
[...]

PR #836, opened 2024-09-12
Allow newly created codecs to be reclaimed if the system is out of resources

Djuffin:
This PR addresses https://github.com/w3c/webcodecs/issues/669

***
<a href="https://pr-preview.s3.amazonaws.com/Djuffin/webcodecs/pull/836.html" title="Last updated on Oct 2, 2024, 2:27 AM UTC (b5e1d38)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webcodecs/836/6fd01eb...Djuffin:b5e1d38.html" title="Last updated on Oct 2, 2024, 2:27 AM UTC (b5e1d38)">Diff</a>

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -3696,7 +3696,7 @@ dictionary VideoFrameMetadata {
 :: The presentation timestamp, given in microseconds. For decode,
     timestamp is copied from the {{EncodedVideoChunk}} corresponding
     to this {{VideoFrame}}. For encode, timestamp is copied to the
-    {{EncodedVideoChunk}}s corresponding to this {{VideoFrame}}. 
+    {{EncodedVideoChunk}}s corresponding to this {{VideoFrame}}.
 
     The {{VideoFrame/timestamp}} getter steps are to return
     {{VideoFrame/[[timestamp]]}}.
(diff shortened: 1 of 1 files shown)


index.src.html
```diff
@@ -6035,7 +6035,8 @@
 received a call to `encode()`, `decode()`, `configure()`, `flush()` or `reset()`
 in the past `10 seconds`, or has called its `output()` callback in the past `10
 seconds`. Addionally, {{VideoEncoder}}s are considered [=active=] if they are
-making progress in encoding queued {{VideoFrame}}s.
+making progress in encoding queued {{VideoFrame}}s. Newly created codecs are not
+considered [=active=] until they call an output callback for the first time.
```
padenot, line 6039: Are we sure? I'd have expected input not output here.

aboba: Ah, yes. Just calling `encode()` or `decode()` is not sufficient if they error (which they can do if resources are not available).

Djuffin: The sentence that I added is intended to mark a newly created coded inactive before it actually got some work done. I do realize that it looks like the new condition contradicts the previous one. I expect that readers will realize that the last condition supersedes the previous ones, but if it looks misleading I'm open to suggestions.

padenot: Yeah, we can't have contradiction in normative text. Let's proceed step by step here:

- Say one of the method mentioned is called. Two things can happen: it can succeed or fail. If it fails, the codec isn't considered active, abort those steps.
- A message is queued. Two things can happen: it can execute or note. If it cannot execute, it means that other things are executing, and so the codec is already active, abort those steps.
- If it executes, then we can mark it as active, we'll e.g. configure a decoder, or submit a frame for encode/decode. It can be that there will be no output for some time: decoder and encoder latency can be very large, and encoding time can also be extremely long.
- When an output callback fires, we should also consider the codec as active, to let script enqueue other tasks, as it is often how the encoding/decoding is driven.

what is important is therefore if one of the methods enqueue a control message successfully OR having an output callback, and that should be our criteria for marking a codec as active.

Djuffin: Until the codec successfully processes at least one input it can't be considered fully initialized, because encoding or decoding might use lazy resource allocation.
It means that we can't really be sure that we have sufficient system resources until we have the first output ready. 

That's why I think that merely having a control message enqueues shouldn't be enough for the codec to be marked as active. After producing the first output processing control messages should be enough to keep the codec active.

padenot: "Accepts", not process, I'd say, but sure. There can be seconds between the two.

Djuffin:
@padenot @aboba 
I simplified definition of an active codec. It's a bit more vague but 
  - closer to how it actually works in real life
  - we already used this definition of activity for `VideoEncoder`
  - aligned with what we discussed at TPAC 2024

padenot (approved):
That should be permissive enough for implementing something useful in practice, thanks!

aboba (approved):

PR #847, opened 2024-11-10
update link to the chrome webcodecs article

ninofiliu:
The article moved and its old link returns a 404, but here's the new link


chrisn:
Thanks!

PR #858, opened 2024-12-10
Specify format of buffer we're copying into

jyavenard:
Per spec https://w3c.github.io/webcodecs/#dom-audiodatacopytooptions-format if the format is not specified, we are to use the same format as the source. Which could be interleaved as the format for decoding isn't specified.

Calling copyTo with a planeIndex of 0 with an interleaved buffer will attempt to copy all planes.

So we just set the destination format to be f32-planar.

This allows playback to start on browsers that aren't Blink based (Firefox, Safari)


padenot:
@ chcunningham has left Google for a while now, but this looks good, thanks.

PR #869, opened 2025-01-14
Handle case where AudioContext.outputLatency isn't implemented.

jyavenard:
Not all User Agents implement AudioContext.outputLatency (such as Safari). On those, totalOutputLatency would always be NaN, resulting in no video frames ever displayed.


padenot (approved):

padenot:
Thanks!

PR #880, opened 2025-02-12
GitHub actions: replace ubuntu-20.04 with ubuntu-latest

deniak:
[GitHub will no longer support the Ubuntu 20.04 runner image starting from April 1, 2025](https://github.com/actions/runner-images/issues/11101). That PR updates the actions to use ubuntu-latest instead


padenot:
Thanks! I don't expect that we depend on anything in particular here, so I'm just merging it.

Issue #626, opened 2023-01-13
Audio codec, priming samples and padding, without container or information from the encoder


padenot:
How are implementations supposed to handle encoder delay / priming samples / packet padding?

If PCM is encoded with a particular codec, and then decoded, does it roundtrip properly? I don't think it can, because the number of priming samples is not returned by the encoder, so there's no way to mux the file properly to signal this (e.g., using an edit list in an mp4, or any other codec/container pair specific way).

It's possible to make a guess for some codecs (e.g. AAC is frequently 2112 frames), but this isn't ideal (and sometimes it depends on the implementation of a codec).

This is important for accurate synchronization, but a requirement for musical applications (because it breaks sample-accurate looping).

dalecurtis:
Agreed! Chrome doesn't do anything today for this. @tguilbert-google

tguilbert-google:
You're right.

Chrome pads the end of Opus encodes with silence.

Would the idea to spec how much priming/padding should be used, or just to expose the information and let users trim accordingly?

tguilbert-google:
As briefly discussed with Dale today, this could probably be an addition to [EncodedAudioChunkMetadata](https://w3c.github.io/webcodecs/#dictdef-encodedaudiochunkmetadata)

padenot:
A design that would be fairly clean would be to have this information on the `EncodedAudioChunk` themselves.

**For the encode side**, this would output `EncodedAudioChunk.timestamp` that can be negative, and `EncodedAudioChunk.duration` that can be different from the number of audio frames actually obtained when decoding (internally).

Then **when decoding**, it's clear that if the pts is negative, the output frames are trimmed, up to a start time of zero, and if the duration is different from the number of frames, the frames are also trimmed, so that the number of frames matches the duration. This can be done by the Web Codec implementation, so there's no confusion.

This means that encoding / decoding round trips properly without having to do anything, and that there is no extradata to carry at any point, each packet carries the information. Additionally, with this design, it's now impossible to ignore the fact that some audio isn't to be rendered, generally yielding better output. More over, this means that comparing timestamps for audio and video makes sense, because they're on the same timeline, and not offset by the encoder delay. This also makes looping of audio assets seamless by default when looping the PCM output (while keeping in mind that decoders need to be reset and that _all_ packets, including the ones with negative timestamps, are to be decoded again for the output to be correct on the second loop iteration). 

[...]

PR #948, opened 2026-09-03
Specify audio decoder priming sample discard and timestamp adjustment

### LINKED ISSUE #944: How should timestamps be handled for audio discarded during decoding?
While trying to fix the negative timestamp test failures in Chromium:
* https://wpt.fyi/results/webcodecs/audioDecoder-codec-specific.https.any.html%3Fopus?label=master&label=experimental&aligned&q=Test%20decoding%20a%20with%20a%20negative%20timestamp

I've run into a question about how things like Opus pre-skip should be handled. The opus test file in this case has a pre-skip of 312 samples (6500us), which Chromium's decoding system happily discards. To ensure timestamps are accurate we adjust the -42 forward (to 6548us).

As I see it, we have a couple choices:
* AudioDecoder shouldn't discard anything.
* AudioDecoder shouldn't adjust timestamps for discarded audio (not sure how this works when full packets can end up discarded).
* Agree that timestamps should be adjusted for discarded audio.
* Do nothing and switch to a Opus test file with zero pre-skip

@padenot @youennf @tguilbert-google WDYT?

dalecurtis:
Audio decoders now automatically discard priming samples if present in the bitstream or configuration description. Decoded outputs have their initial timestamps adjusted forward by the duration of discarded priming samples.

- Add [[priming samples to discard]] slot to AudioDecoder.
- In configure(), initialize slot from description if specified.
- In decode(), update slot from chunk data if specified.
- In Output AudioData, discard priming samples up to slot value, decrement slot, discard completely dropped outputs, and adjust surviving output's timestamp forward by discarded duration rounded to microseconds.

Due to a bikeshed stacking issue, the codec registry is updated in a secondary PR. 

Fixes #944
See #626

### DIFF
--- a/index.src.html
+++ b/index.src.html
@@ -127,6 +127,11 @@ Definitions {#definitions}
     inputs are provided. The underlying codec implementation <em
     class="rfc2119">MUST</em> emit all outputs in response to a flush.
 
+: <dfn id="priming-samples">Priming Samples</dfn>
+:: Audio samples produced at the start of decoding that represent encoder delay
+    or filter warm-up and are not part of the original audio content. Also
+    commonly referred to as "encoder delay" or "pre-skip".
+
 : <dfn lt="system resources">Codec System Resources</dfn>
 :: Resources including CPU memory, GPU memory, and exclusive handles to specific
     decoding/encoding hardware that <em class="rfc2119">MAY</em> be allocated by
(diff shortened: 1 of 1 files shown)


padenot (commented):
This is likely fine, but we might need to settle #949 first.

index.src.html
```diff
@@ -555,22 +572,33 @@
...
+                2. Let |resource| be the [=media resource=] described by |output|.
+                3. Let |resourceReference| be a reference to |resource|.
+                4. Assign |resourceReference| to
+                    {{AudioData/[[resource reference]]}}.
+                5. Let |timestamp| be the {{EncodedAudioChunk/[[timestamp]]}} of the
+                    {{EncodedAudioChunk}} associated with |output|.
+                6. If |discardedFrames| is greater than 0:
+                    1. Increment |timestamp| by the result of dividing
+                        |discardedFrames| by the sample rate of |output|,
+                        multiplied by 1,000,000.
```
padenot, line 592: From memory, we need to round to the closest here, but I'm not 100% sure. I think in my test it ended up being more correct more often.

padenot (approved):
Looks good, thanks a lot. As said in the other PR, this will need a follow-up, that I'll likely do.

index.src.html
dalecurtis: Ah yeah, you're right. We do round internally as well. Done.

